PR #55, opened 2025-09-30
fix: detect GitHub CLI auth via CLI token

### LINKED ISSUE #54: Not working - GH Not Authenticated Error
Image

Image

lox:
GitHub CLI authentication status was being checked solely via a cached keytar token, so fresh installs (or wiped keychains) showed "GitHub Connection Failed" even though `gh auth status` reported a login. The fix makes `GitHubService.isAuthenticated` fall back to `gh auth status / gh auth token`, store the token, and validate via gh api user, eliminating the false failure toast.

Fixes #54.

### DIFF
--- a/src/main/services/GitHubService.ts
+++ b/src/main/services/GitHubService.ts
@@ -116,8 +116,15 @@ export class GitHubService {
    */
   async isAuthenticated(): Promise<boolean> {
     try {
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thanks for catching this and fixing!

PR #56, opened 2025-09-30
feat: add vitest and a test for GitHub CLI auth bug

lox:
Adds vitest and a test that proves the issue for #54, will pass after #55 😇


rabanspiegel:
Thank you, again! Added this, next to file vs in dedicated testing folder. 🙏

PR #57, opened 2025-09-30
UI updates and new diff viewer

jasonkneen:


arnestrickmann:
thanks for the interest in contributing! Unf. we won't accept this PR.

PR #67, opened 2025-10-01
fix(worktree): use execFile and stable IDs for shell safety

tomascupr:
## Summary
- Enforce safe git calls with `execFile` argv (no shell interpolation)
- Sanitize workspace names via `slugify()`
- Stable worktree IDs derived from absolute path: `wt-<sha1-12>`
- Detect default branch via `git remote show origin` (fallback `main`)
- Align create/list/remove to the same ID scheme

## Problem
1) Shell injection risk when composing `git` as a string command.
2) Inconsistent IDs between create/list caused removal flakiness.
3) Hardcoded `main` broke merges on repos with custom defaults.

## Solution
- Use `execFile('git', [...args])` everywhere in `WorktreeService`.
- Slugify user input for branch/dir names.
- Generate deterministic IDs from the absolute worktree path (`sha1` → 12 chars).
- Determine default branch from remote HEAD; fallback gracefully.

## Backward compatibility
- Existing DB rows keep their IDs. Removal paths include `worktreePath`/`branch`, so users can delete legacy entries safely. New worktrees use the stable ID.

## Test Plan
- Type‑check/build main: `npm run build:main` ✅
- Create/remove multiple worktrees with spaces/special chars in names ✅
- Verify PR flow on a non‑`main` default branch ✅

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,7 @@ jspm_packages/
 # OS
 .DS_Store
 Thumbs.db
+
+# AI assistant instructions
+CLAUDE.md
+AGENTS.md
--- a/src/main/services/WorktreeService.ts
+++ b/src/main/services/WorktreeService.ts
@@ -2,6 +2,7 @@ import { execFile } from "child_process";
 import { promisify } from "util";
 import path from "path";
 import fs from "fs";
+import crypto from "crypto";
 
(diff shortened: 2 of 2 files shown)


arnestrickmann:
Looks great - thank you!

PR #73, opened 2025-10-02
Add resizable right sidebar

kirkins:
It's hard to use the terminal in the right sidebar if you need to for some reason.

One idea is to make the right sidebar draggable so you can make it bigger when needed. Perhaps another solution would be a full-screen button. I went draggable sidebar seems simpler.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -89,6 +89,13 @@ const App: React.FC = () => {
   // We only actively detect Codex and Claude Code; Factory (Droid) docs are shown as an alternative.
   const showAgentRequirement = (isCodexInstalled === false) && (isClaudeInstalled === false);
 
+  // Right panel resize state
+  const [rightPanelWidth, setRightPanelWidth] = useState<number>(() => {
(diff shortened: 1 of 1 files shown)


kirkins:
Other idea for another solution.

https://x.com/kirkins/status/1973762968459579773

arnestrickmann:
hi @kirkins - thank you for your interest in contributing. We just released the option to collapse the left sidebar and will add it for the right one as well in the coming days. @rabanspiegel

arnestrickmann:
> Other idea for another solution.
> 
> https://x.com/kirkins/status/1973762968459579773

Fully agree to this. We think it makes a lot of sense to directly induce the CLIs for CC and Codex as well-

rabanspiegel:
Yes, thank you @kirkins. The right (and left) sidebar will be collapsible and draggable with the option to close the terminal viewer. Thank you thinking in lock step with us

rabanspiegel:
PR based on (now) outdated UI. draggability coming by eod

PR #77, opened 2025-10-02
perf(chat): debounce @-mention search and memoize side panels

tomascupr:
This PR addresses severe typing lag reported in large workspaces by reducing unnecessary computation and isolating re-renders.

Changes
- Debounce @-mention file index search in `ChatInput` (120ms). Prevents running a 5k-item search on every keystroke.
- Memoize right-side panels (`FileChangesPanel`, `WorkspaceTerminalPanel`) with `React.memo` to avoid re-rendering while typing in the chat.

Why
- In large repos with many indexed entries and active polling, repeated synchronous searches and panel re-renders can cause multi-second input lag.
- Debouncing and memoization keep the chat input smooth while preserving existing behavior.

Test Plan
1. Open a large repository (thousands of files indexed).
2. Type in the chat input without `@`: typing remains responsive.
3. Type `@` then some characters: suggestions appear after a brief pause (~120ms), arrow keys and Enter work to insert.
4. Observe that the right panels do not flicker or re-render on every keystroke.

Notes
- Scope is intentionally small and safe. No behavioral changes to file change polling or PR status.
- Follow-ups: optionally pause polling during fast typing and consolidate Git status calls to further reduce load in gigantic repos.


arnestrickmann:
thank you, @tomascupr - appreciate your contributions!

PR #96, opened 2025-10-07
feat: Open GH profile in browser when username is clicked

matsjfunke:


matsjfunke:
https://github.com/user-attachments/assets/e440185e-b689-4df7-9cae-6dd369bbaaee

PR #158, opened 2025-10-18
ci: update to node 22 and ensure version consistency

matsjfunke:
Had issues starting emdash since i was on a different node version, which lead me to add save guards and improve the docs regarding node version 

Updated node version to newest 22


arnestrickmann:
thank you!

PR #159, opened 2025-10-18
perf: parallelize cli detection

matsjfunke:
- changed UI the detect CLIs button felt unnecessary replaced with refresh icon & useEffect to load on page open
- used async spawn to run the detection in parallel (only minor speed improvement but felt nicer)
 
https://github.com/user-attachments/assets/d8cfa9b5-0b2c-45fc-937e-7b20008d83bf


arnestrickmann:
lgtm!

PR #225, opened 2025-10-28
feat: add per-workspace terminal tabs

notkainoa:
## Summary
- add a per-workspace terminal store that persists open terminals and the active tab
- render a tab strip with create/close controls inside `WorkspaceTerminalPanel`
- surface hover-only workspace delete buttons so the list stays compact

## Testing
- [x] pnpm test
- [x] pnpm lint
- [x] Manual regression of terminal tab open/close/restore


arnestrickmann:
curious about your general emdash workflow and how you use the terminal in the right side bar. would be great to find out more about it

PR #228, opened 2025-10-29
project deletion from sidebar

rabanspiegel:


arnestrickmann:
.emdash/.planlock.json and .emdash/planning.md shouldnt be included in the PR - will fix the issue of planning.md file being tracked in the staged changes when switching between plan mode and regular mode within a conversation. 

when these two files are removed from the PR I can merge it

PR #291, opened 2025-11-20
Fix sidebar home highlight state

notkainoa:
## Summary
- Added background when in Home tab
- Fixed issue with no background when in projects


arnestrickmann:
hi! thank you for your PR - could you include a screenshot? this would be amazing

PR #396, opened 2025-12-03
Feat/fix-branding-dev-mode

Musti7even:
Set emdash logo and name for dock icon, window title, and app name when running from source (instead of electron logo).
Added also proper logo of emdash fitting into the macOS enviroment.
This ensures consistent branding in both development and production builds.

- Set app name to 'emdash' via app.setName()
- Configure dock icon on macOS in development mode
- Set window title and icon explicitly in BrowserWindow
- add new macOS style emdash logo


arnestrickmann:
good stuff, thank you

PR #414, opened 2025-12-04
feat: make right sidebar open by default

Musti7even:
### Summary
Changes the right sidebar (containing the Changes and Terminal panels) to be open by default instead of collapsed.

### Motivation
Previously, users had to discover and manually open the right sidebar to access the Changes and Terminal panels. This created a discovery barrier for important features.

With this change, users have immediate access to:
- **Changes panel**: See file modifications right after their first prompt completes
- **Terminal panel**: Access the workspace terminal without extra steps

This is especially valuable after the first prompt finishes and changes are made—users can see the changes immediately without needing to find the sidebar toggle first.

### Changes
- Updated `RightSidebarProvider` default from `collapsed=true` to `collapsed=false`
- Updated initial state in `App.tsx` to match the new default
- Removed explicit `defaultCollapsed` prop to use the provider's default


arnestrickmann:
lgtm!

PR #416, opened 2025-12-04
make downloads first and declutter badges

samuelstroschein:
## Changes

- makes the downloads first, not last because higher social cred
- removes low signal badges like release or open issues because that's clear by looking at the github page itself

## Before 

CleanShot 2025-12-04 at 12 09 22@2x

## After

CleanShot 2025-12-04 at 12 09 29@2x


arnestrickmann:
nice, thank you samuel! @samuelstroschein

PR #417, opened 2025-12-04
feat: enable telemetry for developer builds

Musti7even:
### Problem
Currently, only DMG builds (production releases) send telemetry to PostHog because credentials are only injected during CI builds. Developer builds (cloned repo) don't have access to PostHog credentials, so telemetry silently fails even though users haven't opted out.

This creates an inconsistent experience where:
- DMG users: Telemetry works ✅
- Developer users: Telemetry doesn't work ❌ (even though they're normal users too)

### Solution
Since PostHog credentials are public and safe to commit, we now:
1. **Hardcode public credentials** in `src/main/appConfig.json` so developer builds work out of the box
2. **Add build step** to copy `appConfig.json` to `dist/main/` during compilation
3. **Update documentation** to clarify that credentials are public and both build types now support telemetry

### Changes
- ✅ Added PostHog credentials to `src/main/appConfig.json`
- ✅ Updated `build:main` and `dev:main` scripts to copy `appConfig.json` to `dist/main/`
- ✅ Updated telemetry documentation to reflect the new behavior

### How It Works
- **Developer builds**: Load credentials from `src/main/appConfig.json` → copied to `dist/main/appConfig.json` during build
- **DMG builds**: CI still overwrites `dist/main/appConfig.json` with secrets (same values, but allows flexibility)
- **Users can still opt out**: Via `TELEMETRY_ENABLED=false` env var or in-app settings

### Notes
- PostHog credentials are public and safe to commit (as noted in the plan)
- CI workflow continues to work as before (overwrites `dist/main/appConfig.json`)
- No breaking changes - users can still disable telemetry the same way

### DIFF
--- a/docs/content/docs/security/telemetry.mdx
+++ b/docs/content/docs/security/telemetry.mdx
@@ -137,9 +137,10 @@ All telemetry data passes through a **sanitizer** that:
 
 ### Local Development
 
-- Official builds include PostHog credentials via CI
-- Local development builds **do not send telemetry** unless credentials are explicitly added for testing
-- Development mode is clearly marked in events
+- PostHog credentials are **public and safe to commit** to the repository
+- Both official builds (DMG) and developer builds (cloned repo) send telemetry by default
+- Official builds use credentials injected via CI, while developer builds use credentials from `src/main/appConfig.json`
+- Development mode is clearly marked in events via `is_dev: true` and `install_source: "dev"`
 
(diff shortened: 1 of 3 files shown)


Musti7even:
PR is ready now. Tested and checked results in PostHog @arnestrickmann

arnestrickmann:
good stuff

PR #429, opened 2025-12-07
Improve task CTA prominence

Musti7even:
Summary:
**Make “Start New Task” the clear primary action with solid styling and consistent sizing.**

Improve Tasks header hierarchy and helper copy for clearer guidance.
Move selection controls to their own right-aligned row to reduce header clutter.

**Before**
image

**After**
image


Musti7even:
Position and UI of Select button is not ideal but its not too relevant atm. Its ok and the CTA and all buttons are clear imo

PR #441, opened 2025-12-09
Fix: Add new projects to top of sidebar instead of bottom

Musti7even:
New projects were added to the bottom of the sidebar. IMO they should be at the top for easier access


Musti7even:
closed because of wrong changes

Issue #449, opened 2025-12-11
AppImage and Debian packages fail to detect existing Augment Code CLI installation


MadHuslista:
## Bug Description

When launching Emdash via AppImage or Debian package on Linux (e.g., from application menu or desktop shortcut), the application fails to detect the Augment Code CLI (`auggie`) even though it is correctly installed and accessible from the terminal.

## Environment

| Component | Version |
|-----------|--------|
| **Operating System** | Linux Mint 22 (Kernel 6.8.0-88-generic) |
| **Node.js** | v22.14.0 |
| **npm** | 11.6.2 |
| **Augment Code CLI** | 0.11.1 |
| **Emdash** | v0.3.39 (AppImage) |

## Steps to Reproduce

1. Install Augment Code CLI globally: `npm install -g @augmentcode/auggie`
2. Verify CLI works in terminal: `auggie --version` → outputs `0.11.1`
3. Launch Emdash AppImage from file manager or application menu (not from terminal)
4. Navigate to Connections/Providers settings
5. Observe that Augment Code CLI is shown as "Not Installed" or unavailable

## Expected Behavior

Emdash should detect the Augment Code CLI installation regardless of how the application is launched.

## Actual Behavior

The CLI is only detected when Emdash is launched from a terminal session (e.g., `./emdash-x86_64.AppImage`), but not when launched via desktop entry, application menu, or file manager.

## Workaround

Launch Emdash from a terminal session:
```bash
./emdash-x86_64.AppImage
```

This works because the terminal session has the user's shell environment (including PATH) properly configured.

## Root Cause Analysis

After investigating the codebase, the issue appears to be in `src/main/main.ts`. The application has PATH handling for macOS (lines 24-43) and Windows (lines 45-67), but **no equivalent PATH handling for Linux**.

### Current Code Structure:

```typescript
// macOS PATH handling (lines 24-43)
if (process.platform === 'darwin') {
  // Uses fix-path, prepends Homebrew paths, queries login shell
  // ...
}

// Windows PATH handling (lines 45-67)
if (process.platform === 'win32') {
  // Adds npm global path from %APPDATA%/npm
  // ...
}

// Linux: NO PATH HANDLING!
```

### Why This Matters:

When Electron apps are launched via desktop entry (AppImage or .deb), they don't inherit the user's shell environment. The `PATH` environment variable doesn't include:
- `~/.nvm/versions/node/*/bin/` (nvm-managed Node.js)
[...]

PR #469, opened 2025-12-12
fix: hydrate linux PATH for CLI detection.

arnestrickmann:
Ensure Linux launches detect user-installed CLIs by hydrating PATH from common bins and login shell.


arnestrickmann:
closes https://github.com/generalaction/emdash/issues/449

PR #479, opened 2025-12-15
turn terminal scope selector into a dropdown

notkainoa:
Change Terminal scrope tab selector to be a dropdown.

image

## Improvements
- Faster: Reduced unnecessary re-renders so the selector responds quicker
- More accessible: Added proper descriptions for screen readers
- Cleaner: Improved how the component handles its internal state

### DIFF
--- a/PLANNING.md
+++ b/PLANNING.md
@@ -1,28 +1,28 @@
-# Plan Mode (Read‑only)
+# Plan Mode (Read-only)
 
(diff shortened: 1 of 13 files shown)


arnestrickmann:
thanks for contributing! i'd prefer the options to be rendered without the icons because i think the tree icon is not super intuitive. otherwise looks great

notkainoa:
Done. Removed the icons.

notkainoa:
resolved all merge conflicts @arnestrickmann

PR #483, opened 2025-12-15
feat: Add Monaco Editor-based diff viewer with aggregated view

Musti7even:


Musti7even:
Add AllChangesDiffModal for viewing all file changes in a single scrollable view
Refactor ChangesDiffModal to use Monaco Editor instead of syntax highlighter
Add 'Check Changes' button to FileChangesPanel
Add @monaco-editor/react dependency"

PR #493, opened 2025-12-17
fix: prevent AppImage env pollution breaking Python venvs

### LINKED ISSUE #485: Python scripts fail in Emdash worktree terminal when using virtual environments (PYTHONHOME conflict)
## Bug Description

When executing a Python script from within an Emdash worktree terminal using an activated virtual environment, Python fails to initialize with a fatal error. The same script works correctly when run from an external terminal with the same virtual environment activated.

The error indicates that Python cannot find platform-independent libraries due to a `PYTHONHOME` environment variable conflict between the Electron/AppImage environment and the virtual environment.

## Environment

| Component | Version |
|-----------|--------|
| **Operating System** | Linux Mint 22 (Kernel 6.8.0-88-generic) |
| **Node.js** | v22.14.0 |
| **npm** | 11.6.2 |
| **Emdash** | v0.3.41 (AppImage) |
| **Python** | 3.12.3 (system), 3.11 (in venv) |
| **Package Manager** | UV 0.6.6 |

## Steps to Reproduce

1. Open Emdash and add a project with a Python virtual environment
2. Create a task (worktree is created)
3. Open the worktree terminal in Emdash
4. Activate the virtual environment: `source venv/bin/activate`
5. Run a Python script: `python <script>.py`
6. **Result**: Python crashes with fatal initialization error

## Expected Behavior

- Python scripts should execute identically in Emdash worktree terminals as they do in external terminals
- Virtual environment activation should work correctly within Emdash terminals
- No environment-related errors should occur when the same setup works externally

## Actual Behavior

Python fails to initialize with the following error output:

Image

Video sample with the issue. 

https://github.com/user-attachments/assets/d6229a59-05aa-4481-90ce-b88d2c2d85d8

Musti7even:
# Fix: Prevent AppImage Environment Pollution Breaking Python Virtual Environments

## Problem

Fixes #485 - Python scripts fail in Emdash worktree terminals when using virtual environments on Linux AppImage builds.

### What Was Happening

When Emdash runs as an AppImage on Linux, the Electron process inherits AppImage runtime environment variables like:
- `PYTHONHOME` - Points to AppImage's bundled Python runtime (e.g., `/tmp/.mount_emdashXXXXX/usr/lib/python3.x`)
- `APPDIR`, `APPIMAGE`, `OWD` - AppImage packaging metadata
- Other Electron/packaging artifacts

**The bug:** In `ptyManager.ts`, when spawning terminal PTYs, we were doing:
```typescript
const useEnv = {
  TERM: 'xterm-256color',
  ...process.env,  // ← Inherited ALL parent env vars, including AppImage pollution
  ...(env || {}),
};
```

This caused Python virtual environments to fail with:
```
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'
```

**Why it failed:** When users activated a Python venv and ran scripts, Python would read the AppImage's `PYTHONHOME` instead of the virtual environment's Python, breaking module resolution.

**Why it worked externally:** External terminals don't inherit Electron/AppImage environment variables, so Python worked correctly there.

## Solution

Changed `ptyManager.ts` to pass **only essential environment variables** instead of blindly inheriting `process.env`:

```typescript
const useEnv: Record<string, string> = {
  TERM: 'xterm-256color',
  COLORTERM: 'truecolor',
  TERM_PROGRAM: 'emdash',
  HOME: process.env.HOME || os.homedir(),
  USER: process.env.USER || os.userInfo().username,
  SHELL: process.env.SHELL || defaultShell,
  ...(process.env.LANG && { LANG: process.env.LANG }),
  ...(process.env.DISPLAY && { DISPLAY: process.env.DISPLAY }),
  ...(process.env.SSH_AUTH_SOCK && { SSH_AUTH_SOCK: process.env.SSH_AUTH_SOCK }),
  ...(env || {}),  // User-provided overrides
};
```

### Why This Works

1. **Login shells rebuild the environment**: We spawn shells with `-il` (login + interactive) flags, which are specifically designed to rebuild a complete user environment from scratch by reading shell configuration files (`.profile`, `.bashrc`, `.zshrc`, etc.)

2. **Follows Unix conventions**: This is the same approach used by `sudo -i`, `ssh`, and other tools that need to create clean user environments

3. **Self-healing**: If AppImage or other packaging solutions introduce new problematic variables in the future, they're automatically excluded without code changes

4. **User control**: Users get exactly the environment their shell configuration defines, which is predictable and what they'd expect from a "real" terminal

### ⚠️ Testing Limitations

**This bug is specific to Linux AppImage builds** and cannot be reproduced on macOS or Windows native builds. The issue only manifests when:
1. Emdash is packaged as an AppImage on Linux
2. The AppImage runtime sets `PYTHONHOME` to its internal Python
3. This pollutes the terminal environment

### What We've Verified

✅ **Code analysis confirms the bug**: `...process.env` in `ptyManager.ts` leaks all parent environment variables including AppImage artifacts

✅ **Regression testing on macOS**: Python venvs continue to work correctly, no functionality broken

✅ **Solution follows Unix principles**: Same approach as `sudo -i`, `ssh` for clean environments

### Testing Needed

🔍 **Seeking verification from Linux AppImage users** (especially the original reporter @MadHuslista from #485):

```bash
# On Linux, running Emdash as AppImage:
# 1. Open Emdash terminal
env | grep PYTHONHOME  # Before fix: shows AppImage path; After fix: empty

# 2. Test Python venv
mkdir -p ~/test-python-venv && cd ~/test-python-venv
python3 -m venv venv
source venv/bin/activate
python -c "import sys; print(sys.executable)"  # Should work without errors

# 3. Verify no AppImage pollution
env | grep -E 'APPDIR|APPIMAGE|OWD'  # Should be empty
```

### Expected Results (on Linux AppImage)

**Before the fix:**
- ❌ Python venv activation succeeds but running scripts fails
- ❌ `PYTHONHOME` points to AppImage's bundled Python (e.g., `/tmp/.mount_emdashXXXXX/usr`)
- ❌ Fatal initialization error: `ModuleNotFoundError: No module named 'encodings'`

**After the fix:**
- ✅ Python venv activation and script execution work correctly
- ✅ `PYTHONHOME` is not set (or only set by venv activation, not AppImage)
- ✅ No AppImage pollution variables present
- ✅ Common tools (git, npm, etc.) continue to work normally

## Impact

- **Fixes**: Python virtual environments now work correctly in Emdash terminals on Linux AppImage
- **Improves**: Overall terminal environment hygiene across all platforms
- **No breaking changes**: Users will get cleaner, more predictable terminal environments
- **Platform-specific**: Primarily affects Linux AppImage builds, but improves macOS/Windows slightly too

## Design Rationale

We considered several approaches:

1. ❌ **Filter specific problematic variables** - Too fragile, requires maintenance as packaging evolves
2. ❌ **Only remove AppImage-specific vars** - Misses other potential Electron pollution
3. ✅ **Only pass essential vars, let shells rebuild** - Simple, robust, Unix-philosophy approach

The chosen solution is minimal, self-healing, and leverages existing shell initialization systems rather than trying to curate what's "bad."

## Related Issues

- Closes #485
- Similar to issues reported in other Electron-based apps (Cursor, VSCode) where packaging artifacts pollute child process environments

## Request for Testing

**@MadHuslista** (or other Linux AppImage users): Would you be able to test this fix by:
1. Building Emdash from this branch: `fix/appimage-python-venv-conflict`
2. Running it as an AppImage on Linux
3. Testing your Python virtual environment workflow that was failing before

Your feedback would be invaluable to confirm this fix resolves the issue!

### DIFF
--- a/src/main/services/ptyManager.ts
+++ b/src/main/services/ptyManager.ts
@@ -49,7 +49,7 @@ export function startPty(options: {
   const useCwd = cwd || process.cwd() || os.homedir();
 
   // Build a clean environment instead of inheriting process.env wholesale.
-  // 
+  //
   // WHY: When Emdash runs as an AppImage on Linux (or other packaged Electron apps),
   // the parent process.env contains packaging artifacts like PYTHONHOME, APPDIR,
   // APPIMAGE, etc. These variables can break user tools, especially Python virtual


MadHuslista:
@Musti7even  Really thanks guys! You're awesome 
I haven't had yet the opportunity to test, but I'll do it asap 
Thanks again!

MadHuslista:
@Musti7even   Tested from the branch and validated the app image :) 
Reproduced the same workflow used to showcase the issue, and now it's working flawlessly
Thanks! 

https://github.com/user-attachments/assets/02fa22b7-423b-4096-a9a5-3e55c5b8d0e4

PR #517, opened 2025-12-27
fix(github): gracefully handle repos without remotes and detect system gh auth

matteovhaxt:
- Detect system-wide gh CLI authentication (gh auth status)
  - Check for remote existence before git fetch --prune
  - Check for GitHub remote before gh CLI commands
  - Return empty arrays instead of throwing errors
  - Prevents error stack traces for non-GitHub/non-remote repos

  Fixes: "GitHub CLI not authenticated" false positives
  Fixes: "fatal: 'origin' does not appear to be a git repository" errors
  Fixes: "no git remotes found" errors on non-GitHub repos

### DIFF
--- a/src/main/ipc/gitIpc.ts
+++ b/src/main/ipc/gitIpc.ts
@@ -674,10 +674,18 @@ current branch '${currentBranch}' ahead of base '${baseRef}'.`,
       }
 
       try {
+        // Check if remote exists before attempting to fetch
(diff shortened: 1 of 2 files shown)


arnestrickmann (commented):
lgtm

arnestrickmann:
thanks! @matteovhaxt

PR #524, opened 2025-12-29
Fix/ handle repo without remote

matteovhaxt:
- Add hasRemote() helper to check remote existence before fetch/push
- Update WorktreeService to skip fetch when no remote exists
- Update PrGenerationService to check remote before fetching
- Update git:list-remote-branches to fall back to local branches
- Update computeBaseRef() to return local branch names (e.g., 'main') instead of remote refs (e.g., 'origin/main') for local-only repos
- Skip push operations in worktree creation when no remote is configured

Before, task creation failed in repos without remotes


arnestrickmann (commented):
thanks for the addition, @matteovhaxt

Issue #549, opened 2026-01-05
[bug]: Chat input not focused after switching tasks


rabanspiegel:
### Bug Description
After switching between tasks in the sidebar, the chat input is not automatically focused. Users need an extra click before they can type.

### Steps to Reproduce
1. Open a task and type in chat input
2. Click a different task in sidebar
3. Try to type immediately

### Actual vs Expected Behavior
Actual: Need to click chat input before typing
Expected: Chat input should be focused automatically after switching tasks

### emdash Version
latest

### Operating System
macOS

notkainoa:
fixed with https://github.com/generalaction/emdash/pull/644

jschwxrz:
The issue is not completely fixed. When switching between tasks of different projects, the chat input is focused for 1-2 seconds, but unfocuses after that time.

PR #644, opened 2026-01-15
 fix: terminal focus not working when switching tasks (#639)

### LINKED ISSUE #639: [bug]: after switching task the agent is not focused
### Bug Description

when I switch task that's already running a CC instance I can't immediately interact with CC via the keyboard. I need another click into the terminal before it works

### Steps to Reproduce

see above

see above

### emdash Version

Version 0.3.51 (0.3.51)

### Operating System

macOS

### Additional Context

_No response_

arnestrickmann:
## Summary
  Fixes the issue where users need to click into the terminal before being able to type after
  switching tasks.

  ## Changes
  - **ChatInterface**: Added retry logic with exponential backoff for terminal focus
  - **MultiAgentTask**: Added terminal auto-focus support when switching tasks or tabs
  - **TerminalPane**: Added `tabIndex` attribute and focus on terminal ready
  - **TerminalSessionManager**: Enhanced focus method to handle container, xterm, and textarea
  elements

  ## Fixes
  Closes #639

  ## Implementation Details
  - Focus attempts with retry logic (150ms initial delay, then exponential backoff)
  - Multiple focus targets to ensure reliability across different scenarios
  - RequestAnimationFrame to ensure focus survives React re-renders
  - Focus triggered at multiple lifecycle points (attach, ready, task switch)

  ## Testing
  - Switching between tasks with Cmd+Arrow keys
  - Switching between multi-agent provider tabs
  - Opening tasks for the first time
  - All scenarios now properly focus the terminal without requiring a click

### DIFF
--- a/src/renderer/components/ChatInterface.tsx
+++ b/src/renderer/components/ChatInterface.tsx
@@ -58,13 +58,23 @@ const ChatInterface: React.FC<Props> = ({
 
   // Auto-focus terminal when switching to this task
   useEffect(() => {
-    // Small delay to ensure terminal is mounted and attached
-    const timer = setTimeout(() => {
+    // Focus terminal with retry logic to handle async mounting
+    const focusWithRetry = (attempts = 0) => {
       const session = terminalSessionRegistry.getSession(terminalId);
       if (session) {
         session.focus();
+        // Double-check focus after a frame to ensure it sticks
+        requestAnimationFrame(() => {
+          session.focus();
+        });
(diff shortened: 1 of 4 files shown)


arnestrickmann:
fixes https://github.com/generalaction/emdash/issues/639 and https://github.com/generalaction/emdash/issues/549

jonashaag:
This looks like someone (an LLM?) does not properly understand how React works? Is this a reliable fix?

PR #664, opened 2026-01-19
feat(integrations): add GitHub disconnect button in settings

notkainoa:
Users can now disconnect their GitHub account from the integrations settings panel, matching the behavior of the Linear and Jira integrations.

Old:
image

New:
image


arnestrickmann:
thanks @notkainoa!

PR #665, opened 2026-01-19
feat: auto-refresh PR status on window focus and polling

notkainoa:
Add automatic PR status refresh to update sidebar buttons when PRs are created not using the create pr button. Implements window focus refresh for all subscribed tasks and 30-second polling for the active task.

✅ Fully tested


arnestrickmann:
thanks! @notkainoa

PR #675, opened 2026-01-21
fix(sidebar): left sidebar empty state doesn't stretch vertically when toggled

NoumanAMalik:
When the sidebar is empty, and you toggle the left sidebar the "No Project Yet" empty state would adjust the width as the sidebar is collapsing, making the animation janky.

The fix was to set a fixed width on the element, based on what the sidebar is, so during animations, the card just overflows, and doesn't stretch.

Before:
before

After:
after

### DIFF
--- a/src/renderer/components/LeftSidebar.tsx
+++ b/src/renderer/components/LeftSidebar.tsx
@@ -109,6 +109,25 @@ const LeftSidebar: React.FC<LeftSidebarProps> = ({
 }) => {
   const { open, isMobile, setOpen } = useSidebar();
   const [deletingProjectId, setDeletingProjectId] = React.useState<string | null>(null);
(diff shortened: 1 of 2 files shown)


rabanspiegel:
Thanks for looking into this! 🙏  I think this doesn't address the root cause

The issue isn't just the animation when toggling - it's that the sidebar can be dragged to
intermediate widths between 0 and the minSize (16%), which causes the truncation. The ResizablePanel
should enforce a 'snap-to-collapse' behavior: when dragging below minSize, it should snap fully to 0,not allow these intermediate widths.

When the sidebar has projects/tasks, this snap behavior seems to work correctly (it stays at minSize
until it collapses). But with the empty state, something allows dragging to these intermediate
widths. 
   
I'll close this for now and will fix such that it doesn't go below 16%  when there's no project open yet!

PR #683, opened 2026-01-21
feat(terminal): Ability to set custom terminal fonts

NoumanAMalik:
In the Settings modal, under Interface, at the bottom, added a new field to add a custom. 
image

Whatever is added in this input, gets appended at the beginning of the default font families that were already defined from before. In the TerminalSessionManager.ts, I also added an event listener to automatically refresh the terminal whenever the font is changed, so the user sees it update right away.

### DIFF
--- a/src/main/ipc/appIpc.ts
+++ b/src/main/ipc/appIpc.ts
@@ -18,21 +18,19 @@ const execCommand = (
   opts?: { maxBuffer?: number; timeout?: number }
 ): Promise<string> => {
   return new Promise((resolve, reject) => {
-    exec(command, { maxBuffer: opts?.maxBuffer ?? 8 * 1024 * 1024, timeout: opts?.timeout ?? 30000 }, (error, stdout) => {
-      if (error) return reject(error);
-      resolve(stdout ?? '');
-    });
+    exec(
+      command,
+      { maxBuffer: opts?.maxBuffer ?? 8 * 1024 * 1024, timeout: opts?.timeout ?? 30000 },
+      (error, stdout) => {
+        if (error) return reject(error);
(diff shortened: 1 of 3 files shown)


rabanspiegel:
thank you ! will review

rabanspiegel:
Great idea @NoumanAMalik. 

A couple thoughts, it'd be great to have some save indicator or knowing whether the font has been set.

Two other fixes needed before merg: 
- TerminalSessionManager.applyTheme will overwrite the user’s custom fontFamily whenever theme updates run (theme overrides can include a font). Please make the user setting take precedence (e.g., skip applying theme font if a custom font is set, or reapply the custom font after applyTheme).
- When the initial async getSettings() resolves, we set fontFamily but don’t call fitAddon.fit(), so the terminal can be mis‑sized until a resize. Please run fitAddon.fit() after applying the initial settings.

Let me know if you want to take another stab or whether I should finish it 🙏

NoumanAMalik:
bet, I'll get to it and make another commit EOD

NoumanAMalik:
When typing inside the text input field, there is a clear button as well as an keyboard shortcut badge showing the user to click enter to save.
When the save is succesful of the font family, a check mark will be shown to the user.

https://github.com/user-attachments/assets/c6a17541-2419-4584-a486-cecb28d2b82b

For the 2 bugs:
1. When TerminalSessionManager.applyTheme rules, after the font family is set, we reapply the custom font after applyTheme sets the font.
2. if the terminal is open, we run fitAddon.fit()

@rabanspiegel if you see any other issues, I'd be glad to address them.

rabanspiegel:
Thank you, will review by EOD!

PR #702, opened 2026-01-31
Fix agent switch hotkeys

cschubiner:
- wire Cmd+Shift+J/K agent switcher for multi-agent worktree view
- support cmd+shift bindings + next/prev agent shortcuts in settings
- swap J/K so Next = K and Previous = J


rabanspiegel:
Thanks for the impetus @cschubiner ! 🤝

PR #718, opened 2026-02-02
feat: persist last agent selection and move initial prompt to main form

cschubiner:
## Summary

- **Persist agent selection**: Remember the agent configuration (agents + instance counts) from the last task creation. The next "New Task" will default to the same agents previously selected instead of always defaulting to a single agent.
- **Move initial prompt**: Moved the initial prompt field outside the "Advanced options" accordion so it's always visible and easily accessible.
- **Larger textarea**: Increased the initial prompt textarea from 3 to 5 rows for better visibility.

## Test plan

- [ ] Create a task with multiple agents (e.g., Codex + Claude Code 2x)
- [ ] Close modal and click "New Task" again - should show same agent selection
- [ ] Verify initial prompt is visible without expanding "Advanced options"
- [ ] Verify initial prompt textarea is larger than before

### DIFF
--- a/src/main/settings.ts
+++ b/src/main/settings.ts
@@ -62,6 +62,7 @@ export interface AppSettings {
   tasks?: {
     autoGenerateName: boolean;
     autoApproveByDefault: boolean;
+    lastAgentRuns?: Array<{ agent: string; runs: number }>;
   };
   projects?: {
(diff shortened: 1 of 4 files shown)


cschubiner:
Closing - PR should target cschubiner/emdash instead

PR #730, opened 2026-02-04
Reduce git status polling frequency

cschubiner:
## Summary\n- slow down git status polling to 20s in file changes and workspace list\n\n## Testing\n- npm run lint (fails: eslint not installed)


rabanspiegel:
This PR has significant scope creep and needs to be split. Closing for now

PR #735, opened 2026-02-04
Use async numstat for untracked status

cschubiner:
## Summary\n- switch untracked line counts to best-effort git numstat to avoid sync fs reads\n- document unknown counts tradeoff for untracked files\n\n## Testing\n- npm run lint (fails: eslint not found)


rabanspiegel:
opened #748 as a clean extraction. thanks!

PR #743, opened 2026-02-04
Reduce git status polling when inactive

cschubiner:
- pause git status polling when window is hidden or unfocused\n- switch to idle-driven refresh with long interval\n- keep refresh event-driven on focus/visibility changes


rabanspiegel:
Thank you for all PRs, @cschubiner 🙏. Will review tomorrow AM.

PR #759, opened 2026-02-05
feat: stage all files button

alexprudhomme:
This is a feature that I would be interested in for my personal workflow. Do you guys usually stage each file individually ? Or maybe simply use the git commands in the terminal ?


arnestrickmann:
@alexprudhomme thanks for contributing!

This change makes a lot of sense. Especially when having touched a lot of files. 

Merged your PR but did one little change which is that the "Stage all" button is only rendered as soon as the user has already staged one of the changed files (as you can see in the below screenshot).

Before staging the first file: 
Screenshot 2026-02-05 at 18 41 10

After staging the first file manually: 
Screenshot 2026-02-05 at 18 41 43

PR #767, opened 2026-02-05
feat: add default Plan mode for Claude Code tasks

mhenrichsen:
## Summary
- When creating a new task with Claude Code, it now defaults to starting in Plan mode (`/plan` injected via stdin before the prompt)
- Users can override per-task via a checkbox in Advanced Options, or globally via a new toggle in Settings > Tasks
- Only visible when all selected agents support plan mode (currently Claude Code)

## Test plan
- [ ] Create a new task with Claude Code — verify `/plan` appears in terminal before the prompt
- [ ] Uncheck plan mode in Advanced Options — verify prompt goes directly as CLI arg (no `/plan`)
- [ ] Test with non-Claude agents — verify the plan mode checkbox is hidden
- [ ] Test global settings toggle in Settings page
- [ ] Test multi-agent task with Claude Code — verify `/plan` injection works
- [ ] Test with no initial prompt + plan mode — verify `/plan` is sent alone
- [ ] Resume existing task — verify `/plan` is NOT re-sent

### DIFF
--- a/src/renderer/components/ChatInterface.tsx
+++ b/src/renderer/components/ChatInterface.tsx
@@ -956,7 +956,7 @@ const ChatInterface: React.FC<Props> = ({
                   onStartSuccess={() => {
                     setCliStartFailed(false);
                     // Mark initial injection as sent so it won't re-run on restart
-                    if (initialInjection && !task.metadata?.initialInjectionSent) {
+                    if ((initialInjection || planModeActive) && !task.metadata?.initialInjectionSent) {
                       void window.electronAPI.saveTask({
                         ...task,
                         metadata: {


rabanspiegel:
Thank you! I'd like to see this as a setting where you can select whether plan mode should be enabled by default // if supported by provider. Ideally this feature generalizes across providers and is extensible, similar to auto-approve. And then when creating a task it should resolve to present the alternate option. (If you select default === plan mode, then advanced settings should allow to deselect and vice versa)

nthomsencph:
Great option to have, but I disagree on whether it should be default. It is a divergence from Claude Code standard and, insofar as possible, it should be up to the developer to enable features if needed, not disable them, for a cleaner interface to CC.

ZacharyHampton:
I would love this as a setting.

Issue #451, opened 2025-12-11
Emdash requires Git remote (origin) for local-only projects, blocking worktree/task creation


MadHuslista:
## Bug Description

Emdash incorrectly requires a Git remote (`origin`) to be configured even for local-only projects that don't use GitHub or any remote repository features. This blocks core functionality like worktree and task creation for purely local development workflows.

## Context

- Related to #449 (AppImage CLI detection failure) - discovered while testing terminal launch workaround
- Related to #450 (Auto-updater 503 errors in terminal)

## Environment

| Component | Version |
|-----------|--------|
| **Operating System** | Linux Mint 22 (Kernel 6.8.0-88-generic) |
| **Node.js** | v22.14.0 |
| **npm** | 11.6.2 |
| **Emdash** | v0.3.39 (AppImage) |
| **GitHub CLI** | v2.45.0 (authenticated) |

### Scenario 1: Adding a Git repository with remote
1. Open Emdash (launched from terminal as workaround for #449)
2. Attempt to add a new project that is a Git repository with a GitHub remote
3. **Result**: Fails with error message:
   ```
   GitHub Connection Failed
   Git repository detected but couldn't connect to GitHub: GitHub CLI not authenticated
   Tip: Update GitHub CLI via your package manager (e.g., apt/dnf) and restart emdash.
   ```

### Scenario 2: Adding a local-only Git repository (no remote)
1. Add a project that is a Git repository with **no remote configured**
2. Project is added but with warnings in terminal output
3. Attempt to create a task within the project
4. **Result**: Task/worktree creation fails with error:
   ```
   Error
   Failed to fetch origin/main: Command failed: git fetch origin main
   fatal: 'origin' does not appear to be a git repository
   fatal: Could not read from remote repository.
   
   Please make sure you have the correct access rights
   and the repository exists.
   ```

## Expected Behavior

- Emdash should support fully local Git repositories without any remote configured
- Worktree creation should work with local branches  (no need to fetch from origin)
- Task creation should not require a remote repository
- GitHub authentication should only be required when explicitly using GitHub-specific features (creating issues, syncing with GitHub Projects, etc.)

## Actual Behavior

- Adding a project with a GitHub remote fails even when `gh` CLI is installed and authenticated
- Adding a local-only project succeeds but subsequent operations fail
- Worktree creation calls `git fetch origin main` which fails for local-only repos
- The error message doesn't acknowledge that the project intentionally has no remote

## Root Cause Analysis

[...]

PR #831, opened 2026-02-12
feat: support worktree creation from local-only branches

j-token:
## Added Features

- Local-only branches (not yet pushed to remote) now appear in the branch selection dropdown even when a remote exists
- Local branches are visually distinguished with a `(local)` suffix
- Users can select a local branch as the base for worktree creation

## Why This Implementation

The existing `git:list-remote-branches` IPC handler only queried `refs/remotes/origin` when a remote existed, making local-only branches invisible in the UI. Since the backend (`WorktreeService`) already fully supports local branches via `parseBaseRef()` and `fetchBaseRefWithFallback()`, only the IPC handler and renderer mapping needed changes — a minimal modification.

- After listing remote branches, also query `refs/heads/` for local branches
- Deduplicate using a `remoteBranchNames` Set (branches already on remote are excluded)
- Changed `value` to use `b.ref` so a valid git ref is stored as `baseRef`

## Modified Files

| File | Change |
|------|--------|
| `src/main/ipc/gitIpc.ts` | Added local-only branch listing and merging within `hasRemote` block |
| `src/renderer/hooks/useProjectManagement.tsx` | Changed `value` to `b.ref`, added `(local)` suffix for local branches |
| `src/renderer/components/BaseBranchControls.tsx` | Updated placeholder text to `No branches found` |

## Test Plan

- [x] Verify local branches appear with `(local)` suffix in branch dropdown for projects with a remote
- [x] Verify worktree creation works correctly when a local branch is selected
- [x] Verify branches existing on both local and remote are not duplicated
- [x] Verify projects without a remote still work as before
- [x] `pnpm run type-check` passes
- [x] `pnpm run lint` has no new errors

## Related Issues

- Relates to #451 (Emdash requires Git remote for local-only projects)

### DIFF
--- a/src/main/ipc/gitIpc.ts
+++ b/src/main/ipc/gitIpc.ts
@@ -1015,6 +1015,33 @@ current branch '${currentBranch}' ahead of base '${baseRef}'.`,
                   label: `${remoteAlias || remote}/${branch}`,
                 };
               }) ?? [];
+
+          // Also include local-only branches (not on remote)
+          try {
+            const { stdout: localStdout } = await execAsync(
+              'git for-each-ref --format="%(refname:short)" refs/heads/',
+              { cwd: projectPath }
+            );
+
+            const remoteBranchNames = new Set(branches.map((b) => b.branch));
+
+            const localOnlyBranches =
+              localStdout
+                ?.split('\n')
+                .map((line) => line.trim())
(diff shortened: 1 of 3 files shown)


arnestrickmann:
Good catch, thank you for the implementation @j-token!

PR #832, opened 2026-02-12
Add IntelliJ apps

aqeelat:
This pr adds some intellij apps, but only shows them if they're installed. We can also decide later to only show the other dropdown options such as Zed, Ghostty, iTerm2 when they're installed.


arnestrickmann:
LGTM!

PR #845, opened 2026-02-13
feat: Adding native menu handling for settings

chrishoffman:
Adds a proper macOS application menu so "Settings..." (⌘,) appears in the Emdash menu bar, matching standard  
Mac app conventions. Also includes standard Edit, View, Window, and Help menu. Help menu directs to Emdash docs. On non-macOS platforms, Settings is placed in the File menu.

The ⌘, shortcut is now owned by the native menu (which takes priority over the renderer's keyboard handler).

image

### DIFF
--- a/src/main/app/menu.ts
+++ b/src/main/app/menu.ts
@@ -19,7 +19,10 @@ export function setupApplicationMenu(): void {
           {
             label: app.name,
             submenu: [
-              { role: 'about' as const },
+              {
+                label: `About ${app.name}`,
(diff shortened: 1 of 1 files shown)


rabanspiegel:
Great initiative -- thank you! We should also add a 'Check for updates'  in the menu, like such

CleanShot 2026-02-12 at 18 30 44@2x

chrishoffman:
@rabanspiegel https://github.com/generalaction/emdash/pull/847

PR #847, opened 2026-02-13
feat: Add "Check for Updates..." to menu

chrishoffman:
This builds on top of https://github.com/generalaction/emdash/pull/845.

I decided to PR this one separately because it adds a new update screen that is triggered from the menu item and may require some discussion on its own. I tried to link it to the check in settings but it didn't feel right. I'm not sure if this is the best thing to do and I am not entirely familiar with the patterns in other Electron apps. It's also possible that moving updates into its own About/Updates section would be the best course of action. Additionally, this is just a simulated screen as this seems to be impossible to test.

https://github.com/user-attachments/assets/eb93305e-7546-4ce1-9442-efeb6eebc94c

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -462,10 +462,7 @@ const AppContent: React.FC = () => {
                 onClose={handleCloseSettings}
                 initialTab={settingsInitialTab}
               />
-              <UpdateModal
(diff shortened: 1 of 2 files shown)


arnestrickmann:
amazing, great addition. Thank you! @chrishoffman

PR #849, opened 2026-02-13
Feature/integrations svg logos

leondri:
Added SVG logos for CLI Agents that where not visible in dark mode. Also added SVG Logos for integration cards (Jira, Linear, Github)

Before: 
Previous-UI

After: 
Updated-UI


arnestrickmann:
sick, good addition. thank you!

Issue #777, opened 2026-02-05
Changes view can show state from a different chat/task


rabanspiegel:
There appears to be state leakage in the Changes view where diffs from a different chat/task can appear in the current context.

Observed behavior:
- Changes from another chat/task are sometimes shown.
- The view may self-correct later, but not consistently.

Expected behavior:
- Changes should be strictly scoped to the active chat/task/worktree context at all times.

Acceptance criteria:
1. No cross-chat/task change-state contamination in the Changes view.
2. Context switches reliably rebind to the correct scoped data.
3. Any background refresh/reconciliation cannot overwrite active context with stale/foreign state.
4. Add diagnostics to trace context IDs during load/update for debugging.

prath47:
Hi @arnestrickmann 

I would like to work on this issue. Is it open to pick it up ? or somebody else is working on this already ?

arnestrickmann:
@prath47 hey, 
Sounds good!

PR #862, opened 2026-02-14
feat: Clear file changes state when the task path changes and display…

prath47:
Fix: #777 

This PR includes changes as per:
1. Added Clear stale state after change in taskpath
2. used isLoading variable from useFileChanges to show loader on changes tab

Check Performed
1. Lint
2. Format
3. Build
4. Manually verified the working of the changes done;

### DIFF
--- a/src/renderer/hooks/useFileChanges.ts
+++ b/src/renderer/hooks/useFileChanges.ts
@@ -49,6 +49,9 @@ export function useFileChanges(taskPath?: string, options: UseFileChangesOptions
 
   useEffect(() => {
     setFileChanges([]); // Clear stale state immediately
+    if (taskPath) {
+      setIsLoading(true);
+    }
     taskPathRef.current = taskPath;
     hasLoadedRef.current = false;
   }, [taskPath]);


src/renderer/components/FileChangesPanel.tsx
```diff
@@ -740,136 +740,142 @@ const FileChangesPanelComponent: React.FC<FileChangesPanelProps> = ({
         </div>
       ) : (
         <div className="min-h-0 flex-1 overflow-y-auto">
```
prath47, line 742: only the changes are here

prath47:
@arnestrickmann Please help here

arnestrickmann:
Hi @prath47, thanks for opening this PR and for catching this issue!

One thing: 
When testing locally I wasnt able to see the spinner. 

Can you add a screen recording showing the spinner on task switch? Also, please add setIsLoading(true) in the same effect as the setFileChanges([]) clear — without it there's a blank frame before the spinner kicks in:   
                                                                                           
  useEffect(() => {                                                                                             
    setFileChanges([]);                                                                                         
    setIsLoading(true);
    taskPathRef.current = taskPath;              
    hasLoadedRef.current = false;
  }, [taskPath]);

When this is fixed, happy to merge this

prath47:
@arnestrickmann I think

```
    if (taskPath) {
      setIsLoading(true);
    }
```

this would suit better than just adding setIsLoading(true);
it's safer to wrap it in a check for taskPath to ensure we don't get stuck in a loading state if taskPath is undefined (e.g., when no project is open).

arnestrickmann:
Sounds good

arnestrickmann:
thanks! @prath47

arnestrickmann:
tested, lgtm!

PR #875, opened 2026-02-15
feat: allow empty branch prefix and allo custom prefix for every task

Laurin-Notemann:
This PR adds the option to leave the prefix empty and allows a custom branch prefix for every task. 
 
This is more of an idea; feel free to expand on it or move the option for a custom prefix somewhere else.


arnestrickmann:
Thanks Laurin! will review and lyk asap!

Issue #877, opened 2026-02-15
[bug]: opencode doesn't work (regression)


aqeelat:
### Bug Description

When I try to use opencode, terminal pane is empty 

<img width="1445" height="942" alt="Image" src="https://github.com/user-attachments/assets/7857d1fa-a2db-4fb8-a014-361611bfd48d" />

### Steps to Reproduce

-

### Actual vs Expected Behavior

-

### emdash Version

0.4.11

### Operating System

macOS 15

dkapo88:
Same for codebuff

arnestrickmann:
<img width="3026" height="1830" alt="Image" src="https://github.com/user-attachments/assets/a6da54f7-2040-47e4-b780-d147ab197884" />

PR #894, opened 2026-02-16
fix(pty): add keystroke injection mode for Amp and OpenCode agents

### LINKED ISSUE #326: [feat]: add modal after startup
### Feature Summary

add a modal that opens after first starting the app including: 
- embedded demo video
- toggle asking for session recording enablement "help us make the emdash better"
- usage instructions

### Problem or Use Case

when the user opens the app now, some elements of the app might be unintuitive at first 
for example: 
- github cli installation 
- cli installation
- task creation 
- project import
- ...

### Proposed Solution

track if the user has already opened the app before 

if not, open the modal

### Alternatives Considered

_No response_

### Additional Context

_No response_

arnestrickmann:
## Summary

- Adds a `useKeystrokeInjection` flag to the provider definition to support agents that need their initial prompt typed into the TUI after startup rather than passed as a CLI argument
- Enables this flag for **Amp** and **OpenCode**, fixing prompt delivery for agents whose CLI doesn't support passing prompts as arguments in interactive mode
- Updates `ptyManager.ts` to skip CLI arg injection when `useKeystrokeInjection` is true
- Updates `ChatInterface.tsx` and `MultiAgentTask.tsx` to route initial prompts through keystroke injection for opted-in agents

## Context

Some agents (like Amp) define an `initialPromptFlag` (empty string = positional arg) but their CLI doesn't actually accept prompts that way in interactive/TUI mode. This caused prompts to be silently dropped. The new `useKeystrokeInjection` flag allows an agent to declare that it has CLI support but still needs the prompt typed in after the TUI loads.

Closes #326

### DIFF
--- a/src/main/services/ptyManager.ts
+++ b/src/main/services/ptyManager.ts
@@ -281,8 +281,12 @@ export function startDirectPty(options: {
       cliArgs.push(provider.autoApproveFlag);
     }
 
-    // Add initial prompt
-    if (provider.initialPromptFlag !== undefined && initialPrompt?.trim()) {
+    // Add initial prompt (skip if agent uses keystroke injection instead)
+    if (
+      provider.initialPromptFlag !== undefined &&
+      !provider.useKeystrokeInjection &&
+      initialPrompt?.trim()
+    ) {
       if (provider.initialPromptFlag) {
         cliArgs.push(provider.initialPromptFlag);
       }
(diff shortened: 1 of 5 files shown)


arnestrickmann:
this fixes the issue for amp code and opencode https://github.com/generalaction/emdash/issues/877

Issue #891, opened 2026-02-16
[bug]: impossibe to quit vim in built-in terminal


tcrasset:
### Bug Description

Use the terminal, run `vim <A FILE>` and notice that you can't type anything so you're stuck in `vim`.

### Steps to Reproduce

Go to existing task in workspace, type `vim` in terminal, and notice you're stuck.

### Actual vs Expected Behavior

allow input to vim

### emdash Version

0.4.11

### Operating System

Ubuntu 25.10

### Additional Context

```
➜ ELECTRON_ENABLE_STACK_DUMPING=true ELECTRON_ENABLE_LOGGING=true emdash
[dotenv@17.2.4] injecting env (0) from ../../../../../../opt/Emdash/resources/app.asar/dist/.env -- tip: 🔐 prevent building .env in docker: https://dotenvx.com/prebuild
[shellEnv] Detected SSH_AUTH_SOCK: /home/tom/snap/bitwarden/current/.bitwarden-ssh-agent.sock
[1163645:0216/174023.196333:WARNING:bluez_dbus_manager.cc(248)] Floss manager not present, cannot set Floss enable/disable.
Found migrations at: /opt/Emdash/resources/app.asar/drizzle
[1164209:0216/174023.253735:WARNING:sandbox_linux.cc(420)] InitializeSandbox() called with multiple threads in process gpu-process.
Checking for beta autoupdate feature for deb/rpm distributions
Found package-type: deb
[1164409:0216/174025.417097:WARNING:runtime_features.cc(730)] AttributionReportingCrossAppWeb cannot be enabled in this configuration. Use --enable-features=ConversionMeasurement,AttributionReportingCrossAppWeb in addition.
[1164209:0216/174032.894781:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 1 times!
[1163645:0216/174034.499320:INFO:CONSOLE(478)] "[ErrorTracking] n is not defined [object Object]", source: http://127.0.0.1:12112/assets/index-Cm2pHr_h.js (478)
[1163645:0216/174034.499354:INFO:CONSOLE(507)] "Uncaught ReferenceError: n is not defined", source: http://127.0.0.1:12112/assets/index-Cm2pHr_h.js (507)
[1164209:0216/174038.601924:ERROR:gl_surface_presentation_helper.cc(260)] GetVSyncParametersIfAvailable() failed for 2 times!
[...]

PR #895, opened 2026-02-16
feat: Stop filtering focus reporting sequences from terminal input.

prath47:
Fix: #891 

This might fix the issue 

This PR includes changes as:
The focus reporting filter has been removed from the terminal implementation. This should allow vim and other applications to receive input correctly, even when they request focus events.


arnestrickmann:
Thanks! @prath47

PR #908, opened 2026-02-17
fix(search): truncate overflowing text in search results panel

thisisharsh7:
When searching for content in files, long file paths would overflow outside the search panel container, making the UI look broken. This PR added proper text truncation with ellipsis to:
- File paths in search results
- Match preview text

Before:

https://github.com/user-attachments/assets/114f7b4e-3685-4dca-8439-d89f8eb42f8e

After:

https://github.com/user-attachments/assets/b940f7e2-c8d3-4213-a031-3cdedea55576


arnestrickmann:
thanks! great improvement.

Issue #899, opened 2026-02-16
[bug]: Changes done from changes tab does not reflect immediately in IDE file exploler


prath47:
### Bug Description

I did changes from changes tab but it did not reflected in ide file 
i had to reopen the file to see the effect

### Steps to Reproduce

1. open any file in ide
2. changes something
3. now you see changes in changes tab
4. now revert the file from ide or change the content using changes tab and save
5. you'll see the changes do not reflect 
6. reopen the same file now you see the changes

### Actual vs Expected Behavior

the files should be in the sync all the changes should be consistent across all editors

### emdash Version

latest

### Operating System

macOs 15.6

arnestrickmann:
@prath47 what is your discord handle? are you already on our server? this would make communication a lot easier.

prath47:
@arnestrickmann I am already there i'll ping you

arnestrickmann:
alright, saw it

prath47:
@arnestrickmann can you assign this issue to me ?

PR #916, opened 2026-02-17
feat: Implement a file change event system to automatically reload ed…

prath47:
Fix: #899 

@arnestrickmann 

This PR includes changes as:
1. Modified certain files to listen for file change events 
2. When an event is received for a file that is currently open in the editor, 
loadFile is called to reload the content from disk.

The file synchronization issue has been resolved by implementing an event-based notification system that triggers the editor to reload content whenever a file is modified externally within the application.

Checks Performed
1. Format
2. Build
3. Manual verification of the fix


arnestrickmann:
@prath47,

Thanks for the PR! 

I fixed the stale closure issue in useFileManager.ts by using a  ref to track openFiles, so the event listener stays stable and always sees the latest state.

PR #932, opened 2026-02-18
fix(update): fix update dialog state sync, stale flash, and menu routing

chrishoffman:
- Fix update dialog not recognizing already-downloaded/downloading state (would re-trigger    
  checkForUpdates, emitting not-available instead of showing restart button)                    
  - Fix flash of stale state (e.g. "Try Again") when reopening the dialog via useLayoutEffect 
  reset
  - Remove old SettingsModal dialog - fully replaced by the embedded settings page; menu and all
   call sites now route to the embedded view
  - Add Changelog link to update dialog header and Help menus
  - Fix yellow focus ring on update dialog
  - Rename "Emdash Documentation" → "Docs" in Help menu to match settings page terminology

### DIFF
--- a/src/renderer/components/UpdateModal.tsx
+++ b/src/renderer/components/UpdateModal.tsx
@@ -7,7 +7,14 @@ import {
   DialogDescription,
 } from '@/components/ui/dialog';
 import { Button } from '@/components/ui/button';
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thank you @chrishoffman, appreciate this! Looks good 🚀

Issue #927, opened 2026-02-17
UX: Adding a custom skill feels unintuitive - no way to input skill content during creation


arnestrickmann:
## Feedback / UX Issue

**Description**
The flow for adding a custom skill feels unintuitive and clunky. When creating a new skill, the user is only prompted for a name and description. After creation, opening the skill just navigates to a terminal — there's no way to actually define the skill's instructions as part of the onboarding flow.

## Current Behavior
1. User clicks to add a new skill
2. Prompted for **name** and **description** only
3. Skill is created
4. Opening the skill just takes the user to a terminal

## Expected Behavior
The creation flow should allow the user to input the skill file content as part of the onboarding, either by:
- Writing out the skill instructions inline (e.g., a text editor / textarea)
- Uploading an existing skill file
- Or some guided flow that makes it clear how to define what the skill actually does

## Why This Matters
Without a way to define the skill content during creation, users are left wondering how to actually configure the skill. The jump straight to a terminal is disorienting if you're not already familiar with how skills work under the hood.

nehaaprasad:
Hello! @arnestrickmann 
please check out this PR!

PR #968, opened 2026-02-19
feat(skills): add content input field to skill creation flow

nehaaprasad:
fix: #927

### description

- add Instructions textarea to New Skill modal
- save provided instructions into SKILL.md (frontmatter from name/description)
- stop auto-opening terminal after creation

### Test Plan
- create skill with instructions → SKILL.md contains them
- create skill without instructions → default template is used

### DIFF
--- a/src/renderer/components/skills/SkillsView.tsx
+++ b/src/renderer/components/skills/SkillsView.tsx
@@ -227,7 +227,7 @@ const SkillsView: React.FC = () => {
           }
         }}
       >
-        <DialogContent className="max-w-2xl max-h-[90vh] overflow-y-auto">
+        <DialogContent className="max-h-[90vh] max-w-2xl overflow-y-auto">
           <DialogHeader>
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thanks for this; good UX improvement for #927.                                                            
                                                                                                               
Two things to address before merge:                                                                          
                                                                                                               
1. Deduplicate frontmatter generation; The template in SkillsService.ts:325-330 duplicates generateSkillMd() in validation.ts. Please extend generateSkillMd to accept an optional body param instead of copy-pasting the format.

2. YAML escaping on description; Free-text description with double quotes will break the frontmatter. Needs escaping before interpolation into the YAML string.

@naaa760

nehaaprasad:
@arnestrickmann 
updated it!

arnestrickmann:
Hi @naaa760, thank you for your contributions! could you in the future run pnpm run format before opening a PR so that the checks succeed?

Issue #912, opened 2026-02-17
[bug]: failing to pull from repo still shows 'create PR' command


tcrasset:
### Bug Description

I'm using Bitwarden as my ssh agent. Upon startup, emdash does a pull of the repo.If I do not have it open, or if I deny the signing request, the pull cannot succeed.

However, if I have already created the task and the PR in an Emdash session where the I had access to github, it still shows:
- the 'create pr' button, even though the PR already exists
- the 'CI Checks' from Github (which I guess come from the `gh cli`)

 

### Steps to Reproduce

/

### Actual vs Expected Behavior

I expect to not see the button 'Create PR' if the PR exists.

I expect to see a toast saying that the repo couldn't be pulled (or something like that).

### emdash Version

0.4.12

### Operating System

Ubuntu 25.10

### Additional Context

<img width="716" height="377" alt="Image" src="https://github.com/user-attachments/assets/485e56e7-ccc5-4d3b-92be-a1b4e48c75ca" />

```
➜ ELECTRON_ENABLE_STACK_DUMPING=true ELECTRON_ENABLE_LOGGING=true emdash
[dotenv@17.2.4] injecting env (0) from ../../../../../../../opt/Emdash/resources/app.asar/dist/.env -- tip: 🗂️ backup and recover secrets: https://dotenvx.com/ops
[shellEnv] Detected SSH_AUTH_SOCK: /home/tom/snap/bitwarden/current/.bitwarden-ssh-agent.sock
[1708619:0217/144331.358506:WARNING:bluez_dbus_manager.cc(248)] Floss manager not present, cannot set Floss enable/disable.
Found migrations at: /opt/Emdash/resources/app.asar/drizzle
[1709137:0217/144331.475000:WARNING:sandbox_linux.cc(420)] InitializeSandbox() called with multiple threads in process gpu-process.
Checking for beta autoupdate feature for deb/rpm distributions
Found package-type: deb
[1709448:0217/144334.787130:WARNING:runtime_features.cc(730)] AttributionReportingCrossAppWeb cannot be enabled in this configuration. Use --enable-features=ConversionMeasurement,AttributionReportingCrossAppWeb in addition.
Failed to fetch remote before listing branches Error: Command failed: git fetch --prune origin
Load key "/home/tom/.ssh/github_authentication_copied_from_bitwarden.pub": error in libcrypto
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
[...]

PR #969, opened 2026-02-19
fix: hide Create PR when PR exists and toast when pull fails

nehaaprasad:
fix: #912

### description:

- When the remote was unreachable (e.g. SSH agent off), the app still showed "Create PR" for branches that already had a PR and didn’t tell the user that the pull failed.

### Changes:

- Create PR: Only update PR state when the GitHub/gh request succeeds; on failure keep last known PR (memory + localStorage) so we show "View PR" instead of "Create PR" when a PR exists.
- Pull failed: If git fetch fails while listing branches, show a toast: "Could not pull from repository" with the error.

### Result: No "Create PR" when a PR already exists and the repo is unreachable; users see a clear message when pull fails.

### DIFF
--- a/src/main/ipc/gitIpc.ts
+++ b/src/main/ipc/gitIpc.ts
@@ -1013,8 +1013,7 @@ current branch '${currentBranch}' ahead of base '${baseRef}'.`,
           } catch (fetchErr) {
             log.warn('Failed to fetch remote before listing branches', fetchErr);
             fetchFailed = true;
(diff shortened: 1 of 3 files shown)


arnestrickmann:
Thank you! @naaa760 
Could you run pnpm run format?

nehaaprasad:
@arnestrickmann 
could you please let me know why this PR has been closed? I have run pnpm also.

Issue #770, opened 2026-02-05
Updater prompt can show stale version before latest


rabanspiegel:
The in-app update prompt can surface an older/stale version instead of the latest available release.

Observed behavior:
- Update prompt appeared, but not for the newest version.
- After restarting the app, the prompt appeared again with the latest version.

Expected behavior:
- When update UI is shown, it should always reflect the current latest version available for that channel.

Acceptance criteria:
1. Update checks resolve to the latest available version before showing the prompt.
2. Prompt metadata/version is consistent across repeated checks and app restarts.
3. No stale update prompt is shown if a newer version is already available.

PR #970, opened 2026-02-19
feat: Add no-cache header to auto-updater requests to ensure the late…

prath47:
fix: #770 

@arnestrickmann 

This PR includes changes as:
I modified  setupAutoUpdater to include Cache-Control: no-cache headers in update requests. This forces the auto-updater to bypass any intermediate caches (CDNs, proxies, etc.) and fetch the latest release information directly from the source.


arnestrickmann:
Thanks!

Issue #973, opened 2026-02-19
Bug: Terminal does not auto-focus when switching back to Emdash


arnestrickmann:
## Bug Report

**Description**
When switching back to Emdash from another application (e.g., via Alt+Tab), the agent terminal does not automatically regain focus. An extra click inside the terminal is required before keyboard input (like arrow keys) is recognized.

## Steps to Reproduce
1. Have an agent running in a terminal (e.g., Claude Code prompting a question)
2. Switch to another application (browser, editor, etc.)
3. Switch back to Emdash
4. Try to use arrow keys or type in the terminal

## Expected Behavior
The terminal should automatically regain keyboard focus when Emdash becomes the active window, so the user can immediately interact with the agent (e.g., answer prompts with arrow keys).

## Actual Behavior
The terminal does not have focus after switching back. An extra click inside the terminal pane is required before any keyboard input is registered.

## Why This Matters
This adds friction to a very common workflow — switching between your editor/browser and Emdash to check on or respond to agents. Auto-focusing the active terminal on window focus would make this seamless.

nehaaprasad:
Hello! @arnestrickmann 
please check out the PR!

PR #978, opened 2026-02-19
fix(terminal): auto-focus terminal when switching back

nehaaprasad:
fix : #973

### What
- When you switch back to Emdash from another app (e.g. Alt+Tab), the agent terminal now automatically receives keyboard focus.

### Why
- Previously, the terminal did not regain focus when returning to the window, so an extra click inside the terminal was needed before typing or using arrow keys. This was awkward when moving between the browser/editor and Emdash to respond to agents.

### How
- Single-agent tasks: `ChatInterface` listens for `window` focus and focuses the current terminal session.
- Multi-agent tasks: `MultiAgentTask` listens for `window` focus and focuses the active tab’s terminal.
- Existing dialog check in `TerminalSessionManager.focus()` still prevents stealing focus from open modals.

### DIFF
--- a/src/renderer/components/ChatInterface.tsx
+++ b/src/renderer/components/ChatInterface.tsx
@@ -275,6 +275,25 @@ const ChatInterface: React.FC<Props> = ({
   }, [task.id]);
 
   useEffect(() => {
+    let mounted = true;
+    let timer: ReturnType<typeof setTimeout> | null = null;
+    const handleWindowFocus = () => {
+      timer = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Hi @naaa760, did you test your implementation?

nehaaprasad:
@arnestrickmann 
Yes. I tested it manually (Alt+Tab away and back, terminal gets focus) and ran the full test suite all tests pass.

arnestrickmann:
amazing @naaa760

PR #990, opened 2026-02-20
fix(updater): open update modal from "Update Available" toast

chrishoffman:
The "Update Available" toast previously pointed users to Settings. Since the updater modal now exists and auto-starts the download on open, the toast action now opens it directly instead.

### DIFF
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -227,7 +227,7 @@ const AppContent: React.FC = () => {
   });
 
   // Show toast on update availability
-  useUpdateNotifier({ checkOnMount: true, onOpenSettings: () => openSettingsPage('general') });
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Hi @chrishoffman, thank you for the PR. 

I think that many people still go from the toast to the settings to then download the update from there. You are usually clicking on "Check updates" in the menu bar, right? 

Best, 
Arne

chrishoffman:
I'm not sure exactly! I hadn't seen the toast popup until recently and it felt like an extra step to me to be notified that there is an update and be directed to the settings page to click download update. The modal update view is new with the "Check for updates..." menu item and I thought it might make sense to reuse that here. I'm fine either way if you think it makes more sense in the current form.

PR #992, opened 2026-02-20
feat(pty): add shellSetup command for per-project terminal initialization

chrishoffman:
Adds a `shellSetup` field to `.emdash.json` that runs a shell command in every terminal (agent and plain) before the session starts. Useful for activating virtual environments, switching Node versions, or any other per-project shell setup. 

The original inspiration is to prevent the spinning that that the agents do to figure out how to configure your environment. This could be done manually in the launch terminal but there was no way to configure it in the agent terminal.

Here is my setup for emdash:
Screenshot 2026-02-19 at 10 10 06 PM


arnestrickmann:
Nice addition and clean implementation @chrishoffman Thanks!

Issue #1006, opened 2026-02-20
[bug]: glitch - remote ssh, requires refresh  to get project path


yigitkonur:
### Bug Description

just a simple refresh, fixes the issue but a glitch may make some users think it is failing

https://github.com/user-attachments/assets/2e9bf96b-aac2-4f7d-a985-b7f6867c56b4

what can be great?

1. it automatically navigates to home path (not root)
2. sorts folder from most recently modified to old (for all levels, not only home path)
3. colorization of git enabled repo (checking them + colorization of git repo would be cool

interesting idea: instead of listing all folders, like vibekanban, showing all .git contains folders and auto-navigating to them may be a thing you may wanna consider later.

### Steps to Reproduce

1. connect remote ssh
2. go next step after selection of key
3. it fails, press refresh it is fixing

### Actual vs Expected Behavior

already mentioned

### emdash Version

latest

### Operating System

latest macos

arnestrickmann:
Thanks for flagging @yigitkonur, fix will come in the next release!

arnestrickmann:
This is a fix: https://github.com/generalaction/emdash/pull/1017

will be included in the next release! @yigitkonur

PR #1017, opened 2026-02-21
fix(ssh): use connId directly for initial directory browse after connect

arnestrickmann:
## Summary
- Fixed a race condition in `AddRemoteProjectModal` where `browseRemotePath` was called after `setConnectionId`, but the state hadn't updated yet, causing the browse to use a stale/null `connectionId`
- Replaced the indirect `browseRemotePath()` call with inline directory listing that uses the freshly-returned `connId` directly, matching the pattern already used in `selectExistingConnection`

## Details
After a successful SSH connection test and connect in the auth step, the component called `browseRemotePath()` which depends on the `connectionId` state variable. However, since React state updates are asynchronous, `connectionId` was still null/stale at that point, causing the file listing to fail silently. The fix inlines the browse logic using the `connId` value returned directly from `sshConnect()`.

### DIFF
--- a/src/renderer/components/ssh/AddRemoteProjectModal.tsx
+++ b/src/renderer/components/ssh/AddRemoteProjectModal.tsx
@@ -533,8 +533,31 @@ export const AddRemoteProjectModal: React.FC<AddRemoteProjectModalProps> = ({
         });
         setConnectionId(connId);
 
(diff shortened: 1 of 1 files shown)


arnestrickmann:
fixes https://github.com/generalaction/emdash/issues/1006

PR #1024, opened 2026-02-21
Fix revert button to actually discard changes on staged files

### LINKED ISSUE #1022: [bug]: Revert button only unstages files instead of reverting changes
### Bug Description

The "Revert file changes" button (undo icon) doesn't actually revert file changes when the file is staged. It just unstages the file, which is redundant since there's already a separate unstage button (minus icon).

### Steps to Reproduce

1. Open a task and modify a file in the code editor
2. Stage the file using the plus icon
3. Click the revert button (circular arrow icon) next to the file
4. The file gets unstaged but the changes are still there

**Actual:** Clicking revert on a staged file only unstages it, keeping all the changes in the working directory.

**Expected:** Revert should discard all uncommitted changes and restore the file to the last committed version, regardless of whether it's staged or not.

### emdash Version

latest

### Operating System

macOS

### Additional Context

The tooltip correctly says "Discard all uncommitted changes to this file and restore it to the last committed version", but the implementation was only unstaging staged files instead of actually reverting them.

thisisharsh7:
The "Revert file changes" button wasn't working correctly for staged files. It would only unstage them instead of actually discarding the changes, making it functionally identical to the unstage button.

fixes #1022

### DIFF
--- a/src/main/services/GitService.ts
+++ b/src/main/services/GitService.ts
@@ -166,23 +166,6 @@ export async function revertFile(
   taskPath: string,
   filePath: string
 ): Promise<{ action: 'unstaged' | 'reverted' }> {
-  // Check if file is staged
-  try {
-    const { stdout: stagedStatus } = await execFileAsync(
-      'git',
-      ['diff', '--cached', '--name-only', '--', filePath],
-      {
-        cwd: taskPath,
-      }
-    );
-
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thank you for  addressing this issue! @thisisharsh7

PR #1027, opened 2026-02-21
refactor(sidebar): restyle left sidebar layout and alignment

leondri:
Screenshot 2026-02-21 at 16 56 33

Restyled sidebar using the same backend functionality:
* Indented tasks for better visibility
* Added Agents as secondary description
* Added date (e.g. "1d") description

### DIFF
--- a/src/renderer/components/TaskItem.tsx
+++ b/src/renderer/components/TaskItem.tsx
@@ -156,7 +156,9 @@ export const TaskItem: React.FC<TaskItemProps> = ({
     </PrPreviewTooltip>
   ) : (
     <div className="flex items-center gap-1.5">
-      {hasChanges && <ChangesBadge additions={totalAdditions} deletions={totalDeletions} className="text-xs" />}
+      {hasChanges && (
+        <ChangesBadge additions={totalAdditions} deletions={totalDeletions} className="text-xs" />
+      )}
       {compact && <span className="text-xs font-medium text-muted-foreground">{compact}</span>}
     </div>
   );
@@ -207,9 +209,7 @@ export const TaskItem: React.FC<TaskItemProps> = ({
         ) : (
           <>
             {isPinned && <Pin className="h-3 w-3 flex-shrink-0 text-muted-foreground" />}
-            <span className="block truncate text-sm font-medium text-foreground">
-              {task.name}
-            </span>
+            <span className="block truncate text-sm font-medium text-foreground">{task.name}</span>
           </>
         )}
         {showDirectBadge && task.useWorktree === false && (


arnestrickmann:
Thanks! @leondri, will review the PR and give feedback.

src/renderer/components/LeftSidebar.tsx
```diff
@@ -336,107 +289,98 @@ const LeftSidebar: React.FC<LeftSidebarProps> = ({
...
-                                  className="text-muted-foreground opacity-0 group-hover/project:opacity-100 group-data-[state=open]/collapsible:opacity-100"
+                              {typedProject.name}
+                            </button>
+                            {projectIsRemote && connectionId && (
+                              <RemoteProjectIndicator
+                                host={undefined}
+                                connectionState={'disconnected' as ConnectionState}
+                                size="md"
+                                disabled
+                              />
```
arnestrickmann: **Blocking:** `RemoteProjectIndicator` is hardcoded to always show `disconnected` with `host={undefined}`.

The `useRemoteProject` hook was removed but the indicator still renders — remote projects will always appear disconnected regardless of actual state. The `onReconnect` callback is also gone.

Since hooks can't be called inside a render callback, the fix is to extract this into a small child component that calls `useRemoteProject` internally:

```tsx
const SidebarRemoteIndicator: React.FC<{ project: Project }> = ({ project }) => {
  const remote = useRemoteProject(project);
  return (
    <RemoteProjectIndicator
      host={remote.host || undefined}
      connectionState={remote.connectionState as ConnectionState}
      size="md"
      onReconnect={remote.reconnect}
      disabled={remote.isLoading}
    />
  );
};
```

Then use `<SidebarRemoteIndicator project={typedProject} />` here.

src/renderer/components/LeftSidebar.tsx
```diff
@@ -164,11 +130,11 @@ const LeftSidebar: React.FC<LeftSidebarProps> = ({
   onReorderProjectsFull,
   onSidebarContextChange,
   onCreateTaskForProject,
-  onDeleteTask,
+  onDeleteTask: _onDeleteTask,
```
arnestrickmann, line 152: **Blocking:** `onDeleteTask` and `onDeleteProject` (lines 133, 137) are accepted as props but underscore-prefixed and never used. This means:
- Users can no longer delete tasks from the sidebar
- Users can no longer delete projects from the sidebar

These are feature regressions. Even if the primary sidebar action becomes archive-on-hover, delete should still be accessible — e.g. through the existing `TaskItem` context menu, or as a secondary hover action.

src/renderer/components/TaskItem.tsx
```diff
@@ -43,21 +62,22 @@ interface TaskItemProps {
 
 export const TaskItem: React.FC<TaskItemProps> = ({
   task,
-  onDelete,
+  onDelete: _onDelete,
```
arnestrickmann, line 66: **Blocking:** `onDelete` is renamed to `_onDelete` and never used (same for `showDelete` → `_showDelete` on line 70). The delete button was completely removed from `TaskItem`.

The context menu still has archive + pin + rename — `onDelete` should be wired back in there as a context menu option, even if the hover action is now archive instead of delete.

src/renderer/hooks/useTaskAgentNames.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export function useTaskAgentNames(taskId: string, fallbackAgentId?: string): AgentNameInfo {
+  const [info, setInfo] = useState<AgentNameInfo>(FALLBACK);
+
+  useEffect(() => {
+    let cancelled = false;
+
+    async function load() {
+      try {
+        const res = await window.electronAPI.getConversations(taskId);
```
arnestrickmann, line 27: **Performance:** This hook fires `getConversations(taskId)` per task on mount. With 20 tasks visible, that's 20 IPC round-trips.

Consider either:
- A batch API that fetches agent info for all tasks in one call
- A simple in-memory cache with short TTL
- Lifting the fetch to the parent list component and passing results down as props

src/renderer/hooks/useTaskAgentNames.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        const primaryProvider = getProvider(providerIds[0] as ProviderId);
+        const primaryName = primaryProvider?.name ?? providerIds[0];
+        const totalChats = conversations.filter((c: { provider?: string }) => c.provider).length;
+        const additionalCount = Math.max(0, totalChats - 1);
+        const displayLabel =
+          additionalCount > 0 ? `${primaryName} +${additionalCount}` : primaryName;
+
+        setInfo({ primaryName, additionalCount, displayLabel, providerIds });
+      } catch {
```
arnestrickmann, line 63: **Bug:** `additionalCount` is `totalChats - 1` (total conversations minus 1), not unique providers minus 1. If a task has 5 Claude conversations, this shows "Claude +4" — but there's really only 1 unique provider.

Should this be `providerIds.length - 1` instead of `totalChats - 1`? Or if showing chat count is intentional, the label should clarify (e.g. "Claude (5 chats)").

arnestrickmann:
Thanks for the changes, @leondri!
Sent you feedback 🙏

arnestrickmann:
Thanks, Leon!

PR #1030, opened 2026-02-21
fix(terminal): chat input not focused after switching tasks between projects

### LINKED ISSUE #549: [bug]: Chat input not focused after switching tasks

jschwxrz:
terminalpane useeffect dependency array included props (remote, theme, env, callbacks, etc.) that are only used at PTY spawn time. Parent rerenders created new object/function references, triggering unnecessary detach/attach cycles that moved the terminal DOM offscreen and lost focus.

Move all props except id to refs so the effect only reruns when switching to a different terminal. add a separate theme effect that calls settheme() without detach/attach.

fixes #549

### DIFF
--- a/src/renderer/components/TerminalPane.tsx
+++ b/src/renderer/components/TerminalPane.tsx
@@ -1,11 +1,4 @@
-import React, {
-  useEffect,
-  useRef,
-  useMemo,
-  useCallback,
-  forwardRef,
-  useImperativeHandle,
-} from 'react';
+import React, { useEffect, useRef, useMemo, forwardRef, useImperativeHandle } from 'react';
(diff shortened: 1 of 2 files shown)


arnestrickmann:
thanks for opening the pr jona! 
Will check and test asap.

arnestrickmann:
Thanks Jona! Good stuff. LGTM. Merging.

Issue #1033, opened 2026-02-21
Feature: Auto-infer workspace names from conversation context


arnestrickmann:
## Feature Request

**Description**
Users shouldn't have to manually name workspaces. Emdash should automatically infer a meaningful name based on the conversation or task context, similar to how ChatGPT/Claude auto-title conversations.

## Current Behavior
Users are required to provide a name when creating a workspace. This adds friction to the creation flow, especially when you just want to jump in and start working.

## Expected Behavior
- Workspace creation should not require a name upfront
- A name should be auto-generated based on the initial prompt or conversation context
- Users can optionally rename the workspace later if they want something more specific

## Suggested Approach
- Use the first message / task description to generate a short, descriptive name
- Could use a lightweight LLM call or simple heuristic (first N words, key phrases, etc.)
- Allow inline renaming for users who want to override the auto-generated name

jschwxrz:
looking into this

PR #1041, opened 2026-02-22
feat: auto-infer workspace names from conversation context

jschwxrz:
summary:

- task name is now optional
- auto populates when issue is linked or when initial prompt exists
- if task is created without issue or initial prompt and user does not add task-name, falls back to random name on submit
- fallback random name is updated with first task name generated based on "real" message user sends in the chat

how it works:
- uses keyword extraction to generate a conventional task name
- pre-task-creation: generate and use as task and worktree name
- post-task-creation: buffer accumulates user keystrokes. when message is sent to agent and the agents PTY output shows "busy", a task name is generated based on the user message and task is renamed
- manual task rename still possible through right clicking the task in the left sidebar

UI changes:
- task name shows optional
- create button enabled even when task name empty

Screenshot 2026-02-22 at 21 03 34

#1033

### DIFF
--- a/src/renderer/terminal/TerminalInputBuffer.ts
+++ b/src/renderer/terminal/TerminalInputBuffer.ts
@@ -35,7 +35,8 @@ export class TerminalInputBuffer {
         if (
           (code >= 0x40 && code <= 0x5a) || // A-Z
           (code >= 0x61 && code <= 0x7a) || // a-z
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thank you for this PR, 

We shipped another PR implementing this feature. 

So closing this one.

PR #1048, opened 2026-02-23
fix: show stage all button when no files are staged

### LINKED ISSUE #1047: UX: Stage all button is missing if no files are staged
The 'Stage All' button only appears if at least one file is staged, but does not appear when non are staged.

It should appear unconditionally.

Image

jschwxrz:
Shows "Stage All" Button also when no file is staged yet

fixes #1047


arnestrickmann:
@jschwxrz Thanks for opening this PR!

Issue #1039, opened 2026-02-22
[bug]: Github auth shows logged in as User even when i have name and username


prath47:
### Bug Description

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/fd26260c-684f-44d1-b161-93409a069544" />

### Steps to Reproduce

first time login using github

### Actual vs Expected Behavior

it should show proper name or user name of logged in user

### emdash Version

latest

### Operating System

macOs 15,6

PR #1051, opened 2026-02-23
Fix 1039

prath47:
fix: #1039 

This PR includes changes as :
1. Fix the user name in modal after successful sign in


arnestrickmann:
Good catch, thank you @prath47

PR #1069, opened 2026-02-24
fix(ci): add electron-rebuild step to Linux release build

### LINKED ISSUE #1068: [bug]: critical: Github Copilot Chat UI does not work anymore since 0.4.16
### Bug Description

Image

### Steps to Reproduce

- Open newest 0.4.16 version
- Use Github Copilot as the agent
- Be greeted with this screen

Setting the EMDASH_DISABLE_PTY=1 made it not work either.

I expected to be able to use emdash with Github Copilot CLI in the newest version.

Workaround:

Use 0.4.15, it works fine.

### emdash Version

0.4.16

### Operating System

Ubuntu 25.10

### Additional Context

_No response_

jschwxrz:
summary:
- fixes PTY unavailable error on Linux since 0.4.16 (NODE_MODULE_VERSION 127 vs. 123

cause:
- native modules (node-pty, sqlite3, keytar) are compiled for CI runners node 22 (ABI 127)
- those modules need to be recompiled for electrons node version (ABI 123)
- in 0.4.15, postinstall script handled this rebuild, but was changed to be skipped in CI in 0.4.16 (mac build needs to be per architecture, postinstall could not handle that)
- each platform is now expected to run "electron-rebuild" explicitly in the workflow
- mac and windows already had that step, linux was still missing it

fix:
- added electron-rebuild step for linux

fixes #1068

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -238,6 +238,12 @@ jobs:
           sed -i "s#__PH_KEY__#${PH_KEY}#g" dist/main/appConfig.json
           echo "Wrote dist/main/appConfig.json"
 
+      - name: Rebuild native modules for Electron (x64)
+        run: |
+          set -euo pipefail
+          ELECTRON_VERSION=$(node -p "require('electron/package.json').version")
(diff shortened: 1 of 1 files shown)


arnestrickmann:
@jschwxrz thanks!

PR #1076, opened 2026-02-24
docs(readme): escape install commands using <code> tags and HTML entities for pipes

Sebastian90746:
The install commands in the README table were not being rendered correctly because Markdown interpreted the pipe characters (`|`) inside the commands as column separators. As a result, parts of several commands were treated as separate table cells, causing missing code segments and incorrect text rendering.

This update escapes all pipe characters using `&#124;` and wraps each install command in `<code>` tags. The README now displays the full commands as intended without breaking the table layout.


arnestrickmann:
Thanks!, @Sebastian90746 good catch!

PR #1079, opened 2026-02-24
feat: show 'create pr' if commits are ahead of main

Valley-15:
Problem:
The 'Create PR' button was only visible when the working directory had uncommitted changes. If a user committed locally but hadn't yet opened a PR, the button would disappear.

Solution:
Updated FileChangesPanel.tsx to ensure branchAhead status is tracked even when hasChanges is true. This allows the "Create PR" button to persist as long as the local branch is ahead of main.

Verification:
Verified via code analysis that the branchAhead state is no longer reset by uncommitted file presence.


arnestrickmann:
Thanks!

Issue #1052, opened 2026-02-23
[bug]: UI shows to merge pull request although it is already closed


prath47:
### Bug Description

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/2d7df96c-d07f-4b79-97b7-144160678283" />

### Steps to Reproduce

create a pr
close it 
and check

### Actual vs Expected Behavior

It should merge for only open pull request !

### emdash Version

Latest

### Operating System

macOs 15.6

prath47:
@arnestrickmann please assign this to me

arnestrickmann:
Hey @prath47, i already fixed this. 

Maybe for you its an upstream of git auth not working?

prath47:
@arnestrickmann Steps i followed

1. created a task with github issue
2. created pr 
3. closed the pr 
4. did not close the issue

i still see the issue for newly created pr

<img width="1470" height="921" alt="Image" src="https://github.com/user-attachments/assets/c2102065-0546-4088-b37d-495ac335b00e" />

PR #1083, opened 2026-02-24
feat: prevent MergePrSection from rendering for closed pull requests

prath47:
fix: #1052 

This PR includes changes as: 
1. Added handling when PR is closed 

Check
1. Format
2. build 
3. Manual Verification


arnestrickmann:
Thank you @prath47, and good catch

Issue #1085, opened 2026-02-24
[feat]: Add a loading state in Github issue dropdown


prath47:
### Feature Summary

It would be nice if i can see the github issues are being fetched and track the process

### Problem or Use Case

It is easy to track if the issues are fetched or there are no issues exists

### Proposed Solution

in the loading state, we still render the github icon inside the disabled selet element, but exchange "Select a GitHub issue" with "Loading GitHub issues" + Spinner

<img width="475" height="91" alt="Image" src="https://github.com/user-attachments/assets/ec303147-1fac-4199-83d8-8b50f60eafc8" />

PR #1086, opened 2026-02-24
feat: add a loading state with a spinner to the GitHub issue selector

prath47:
feat: #1085 

This pr includes changes as:
1. added spinner and changed dropdown text when the issues are being fetched

Checks Performed
1. lint
2. build 
3. verification of the loading state


arnestrickmann:
Thanks for opening this PR! 
We should also bring the same experiecne to the other issue selectors, wdyt?

prath47:
Sure i'll add that

Issue #1074, opened 2026-02-24
[bug]: cannot enter commit message


tcrasset:
### Bug Description

<img width="757" height="747" alt="Image" src="https://github.com/user-attachments/assets/60ad52a0-9b65-4745-a2b0-7dd7f329a7a6" />

### Steps to Reproduce

1. Create task
2. Let agent work
3. Manually stage a change
4. The commit message field appears
5. Click on it
6. Notice you cannot enter any message

### Actual vs Expected Behavior

I expect to be able to type in my commit message.

### emdash Version

0.4.15

### Operating System

Ubuntu 25.10

tcrasset:
Related to https://github.com/generalaction/emdash/issues/1467

PR #1094, opened 2026-02-25
fix: add autoFocus to commit message input

kanweiwei:
## Summary
Added `autoFocus` attribute to the commit message input field.

This should fix issue #1074: cannot enter commit message - when the commit message input appears, it will now automatically receive focus, allowing users to type immediately without having to click on it first.

## Testing
- The change is minimal and only affects the commit message input field
- Users should now be able to type commit messages without extra clicks


arnestrickmann:
Thanks for opening this PR and the interest in contributing! @kanweiwei

Issue #1073, opened 2026-02-24
[feat]: show 'create pr' if commits are ahead of `main`


tcrasset:
### Feature Summary

I want the 'Create PR' button to be shown when no PR exists on the repo, but there are commits ahead of the base branch (e.g. `main`) locally.

### Problem or Use Case

Currently, the 'Create PR' button is only shown when there are staged/unstaged files, but is not shown when there are no unstaged files, even though I just committed, and am ready to create a PR.

### Proposed Solution

/

### Alternatives Considered

run `gh pr create` manually in my terminal, like a caveman

### Additional Context

See the missing Create PR button and my git log below

<img width="757" height="747" alt="Image" src="https://github.com/user-attachments/assets/c8e54737-55ca-4db9-a560-60b94d7cc026" />

rabanspiegel:
Thanks for flagging! Will fix this! So you don't have to create PRs like a caveman

PR #1098, opened 2026-02-25
fix: show Create PR button when branch is ahead of main

kanweiwei:
## Summary
Fix issue #1073: Show Create PR button when commits are ahead of main, even when there are no uncommitted file changes.

## Changes
- Removed the `hasChanges` check from branch status useEffect in FileChangesPanel.tsx
- Now branch status is always checked (except for remote SSH paths)
- Create PR button will show when local branch is ahead of main, regardless of whether there are uncommitted changes

## Before
Create PR button only showed when there were staged/unstaged files

## After  
Create PR button shows when branch is ahead of main, even with no uncommitted changes


arnestrickmann:
https://github.com/generalaction/emdash/pull/1079

Issue #1082, opened 2026-02-24
Feature: Confirm before closing app when agents are still running


arnestrickmann:
## Feature Request

**Description**
When the user closes Emdash (e.g., `Cmd+W` / `Cmd+Q` / clicking the close button) while one or more agents are still running, the app should prompt for confirmation before actually closing.

## Current Behavior
The app closes immediately without warning, even if agents are actively running tasks. This can result in interrupted work, lost agent progress, or half-finished worktree operations.

## Expected Behavior
1. User attempts to close the app while agent(s) are running
2. A confirmation dialog appears: e.g., *"You have X running agent(s). Are you sure you want to quit?"*
3. User can choose to close anyway or cancel and stay in the app

## Why This Matters
Accidentally closing the app mid-task can waste significant agent progress, especially on long-running operations. A simple confirmation dialog prevents this with zero friction for users who don't have running agents.

prath47:
Hey @arnestrickmann can you assign this feat to me ?

nehaaprasad:
@arnestrickmann 
please check out the PR, and all files were already formatted

winshaurya:
this feature is a must . please assign so i attempt to add this

arnestrickmann:
Feel free to pick it!

PR #1100, opened 2026-02-25
feat: confirm before closing app when agents are running

nehaaprasad:
close : #1082

### summary

- adds a confirmation dialog when the user tries to close Emdash while one or more agents are still running, so they don’t accidentally quit and lose work.

### changes

- show a confirmation dialog when closing the window (Cmd+W, close button) or quitting the app (Cmd+Q) while agents are running
- dialog text: "You have X running agent(s). Are you sure you want to quit?" with Cancel and Quit
- If no agents are running, the app closes without prompting
- Move shutdown cleanup from before-quit to will-quit so it only runs when the app actually quits, not when the user cancels
- Add getRunningAgentCount() in ptyIpc to count active PTYs
- Add closeConfirmation module to share confirmation state between window close and app quit

### testing

- All existing tests pass
- Updated ptyIpc test to use will-quit instead of before-quit

### DIFF
--- a/src/main/services/ptyIpc.ts
+++ b/src/main/services/ptyIpc.ts
@@ -45,7 +45,7 @@ const wcDestroyedListeners = new Set<number>();
 let isAppQuitting = false;
 
 export function getRunningAgentCount(): number {
-  return owners.size;
+  return providerPtyTimers.size;
 }
 
 type FinishCause = 'process_exit' | 'app_quit' | 'owner_destroyed' | 'manual_kill';


arnestrickmann:
Hi @naaa760,

getRunningAgentCount() uses owners.size, which counts all open PTY terminals — not actively running agents. After an agent finishes, the PTY stays alive (it execs into an interactive shell), so this would always prompt even when no agents are working.
Please use providerPtyTimers.size instead — it already tracks actual agent runs via maybeMarkProviderStart/maybeMarkProviderFinish.

nehaaprasad:
@arnestrickmann 
updated it

arnestrickmann:
Thank you. @naaa760,
What exactly did you update? 

@greptile

nehaaprasad:
Hello! @arnestrickmann 
sorry for being a little late for reply 
i added a confirmation dialog before closing when agents are running. 
Window close and app quit both use providerPtyTimers.size to detect active agents. 
Shutdown logic was moved to will-quit.

Issue #1081, opened 2026-02-24
Bug: UI freeze/stutter when typing slash command in new task


arnestrickmann:
## Bug Report

**Description**
The UI stutters and temporarily freezes when typing a slash command (`/`) immediately after starting a new task.

## Steps to Reproduce
1. Create/start a new task
2. Immediately begin typing in the input field
3. Type `/` to trigger a slash command
4. Observe the UI freeze/stutter

## Expected Behavior
Slash command autocomplete should appear instantly without any UI jank or freeze.

## Actual Behavior
The UI temporarily freezes after typing `/`, causing a noticeable stutter before the slash command menu appears (or input resumes).

## Possible Cause
The slash command menu may be triggering an expensive operation on first render (e.g., loading command definitions, filtering, or initializing a component tree that wasn't yet mounted). This could be worse on new tasks where certain resources haven't been warmed up yet.

## Suggested Fix
- Lazy-load or pre-cache the slash command list so it's ready before the user types `/`
- Debounce or optimize the filtering/rendering of the command palette
- Profile the input handler to identify what's blocking the main thread

DevDesai444:
Can I attend to this bug ?

arnestrickmann:
@DevDesai444 what would be your proposed fix?

DevDesai444:
I think this issue is because of main thread gets overloaded in our PTY terminal event path, not the slash command feature itself. When users types / multiple operation like onPtyData, activity classification regex scans, and sync storage writes are triggered all at once which is blocking the rendering and causing the freeze.
And maybe if we optimise it like, process same chunk only once in onPtyData and Put Cap on input for activity classification we can resolve the issue. 
We can also remove per activity localStorage.setItem writes and writing only when value actually changes.

@arnestrickmann What do you think ?

PR #1107, opened 2026-02-25
Fix Slash-Command UI Freeze on New Tasks (Bug#1081)

DevDesai444:
## Description

Fixes UI freezewhen typing `/` immediately after creating a new task ([#1081](https://github.com/generalaction/emdash/issues/1081)).

## Problem Identified

The first slash-command interaction on a fresh task could block the renderer thread due to a combination of startup-time hot paths:

1. **Large PTY output bursts were written synchronously** into xterm in a single pass.
2. **Activity classification fanned out too aggressively** (multiple per-provider/per-kind listeners), increasing work exactly when the terminal first became active.
3. **Activity classification processed full chunks**, which is expensive for large buffered output.
4. **Terminal activity wrote to localStorage repeatedly**, adding synchronous overhead during early typing.

This made the UI feel frozen/stuttery right when `/` was typed.

### 1) Reduce main-thread blocking from terminal output
- Added **frame-sliced terminal write draining** in `TerminalSessionManager` so big PTY bursts are processed incrementally instead of one heavy synchronous write.
- Preserved scroll behavior and first-frame handling.
- Added **slow input-path telemetry logging** (`terminalSession:slowInputHandler`) to surface future regressions.

### 2) Reduce activity-event overhead
- Main process now emits a **global PTY activity channel** (`pty:activity`) and **global PTY exit channel** (`pty:exit:global`).
- Renderer `activityStore` now **prefers the global feed** and only falls back to direct listeners when needed.

### 3) Bound classifier work per event
- Added shared chunk sampling (`sampleActivityChunk`) and classify only sampled tail data in hot paths.
- Applied this to activity classification paths used during startup/interactions.

### 4) Remove redundant synchronous storage writes
- Debounced/deduped `agent:locked` writes so terminal activity does not repeatedly hit localStorage for identical values.

## Why this addresses #1081 
The freeze was not caused by React command palette rendering itself; it was caused by renderer contention during first terminal interaction/output.  
This PR removes the biggest synchronous hotspots on that exact path, so slash command UI appears without the initial jank.

## Validation
- `pnpm run type-check` ✅
- `pnpm run lint` ✅ (existing repo warnings only)
- `pnpm run format` ✅
- `pnpm exec vitest run src/test/main/ptyIpc.test.ts` ✅
- Manual flow validated: create task → immediately type `/` → no visible freeze/stutter.

https://github.com/user-attachments/assets/31350cee-cc93-44e8-aa32-c8819f664ff4


arnestrickmann:
Hi @DevDesai444,
Can you please change the pr title to be more descriptive? 

Thank you for your interest in contributing and for opening this PR.

DevDesai444:
@arnestrickmann, is there any change required in this PR ?

arnestrickmann (commented):
reviewing

PR #1112, opened 2026-02-25
fix: improve tab bar layout with scrollable tabs, pinned context badges, and overflow

leondri:
Restructured the chat bar
* Context badges (Linear tickets, auto-approve) will always be visible, aligned on the right side
* chat tabs will expand, and if space isn't sufficient, fade out (horizontal scroll)
* Plus icon to add a new chat is always close to the latest chat

Screenshot 2026-02-25 at 12 40 27

Screenshot 2026-02-25 at 12 40 51


arnestrickmann:
Looks slick, thanks!

PR #1114, opened 2026-02-25
fix: opening emdash also opens cursor

### LINKED ISSUE #1109: [bug]: Starting emdash also starts Cursor
### Bug Description

When I launch emdash, it also starts Cursor. Really weird..

Is there an issue with running several vscode derivatives on my system at the same time?

### Steps to Reproduce

1. Start emdash from CLI

(doesn't matter if Cursor is already also running or not)

I expect only the emdash window to open.

Actual behaviour is both a Cursor window and an emdash window opens.

 

### emdash Version

v0.4.18

### Operating System

Debian 12

### Additional Context

I also can't get sqlite up because I am on too old version of GLIBC, need to upgrade my Debian -- so seems I can't use your prebuilt binaries until I have fixed that -- but I would suppose this is unrelated.

jschwxrz:
summary:
  - starting emdash on Linux also opens Cursor IDE
  - on startup, provider detection runs cursor-agent --version, and if not found falls back to
   cursor --version -- which is the IDE launcher on Linux
  - same pattern for Kiro (kiro-cli -> kiro)
  - also caused false-positive "connected" status when only the IDE was installed, not the CLI
   agent

  fix:
  - removed cursor and kiro IDE commands from commands in registry.ts, keeping only
  cursor-agent and kiro-cli
  - cli field (used for PTY spawning) was already correct, so agent functionality is
  unaffected

  limits:
  - could not verify on Linux but the code path is clear

  fixes #1109


arnestrickmann:
Thanks! Good one.

PR #1116, opened 2026-02-25
fix(agent): improve CLI detection to avoid false "not installed" alerts

### LINKED ISSUE #1111: [bug]: Emdash alerts "Claude Code isn’t installed" but still opens up Claude Code
### Bug Description

The detection logic of Claude Code installation is not working properly. Emdash simultaneously says Claude is not installed, but also starts Claude Code without any visible issue.

Image

### Steps to Reproduce

1. Open up a new Task using Claude Code

If Claude Code starts, there should not be alert about it not being found

### emdash Version

Version 0.4.18 (0.4.18)

### Operating System

macOS 26.2 (25C56)

### Additional Context

_No response_

jschwxrz:
summary:
- emdash shows claude code isn't installed banner but then starts claude code fine
- the detection check runs claude --version and if it exits with a non-zero code, emdash concludes the binary is missing
- detection can fail to find a cli binary that's only on the user's PATH via shell config files like .zshrc, while the actual agent startup (which uses a full shell) finds it fine 

fix:
- detection now checks if the binary was found by "which" or if the process actually ran before concluding it's missing
- when the binary cant be found directly, detection retries through the users login shell, matching how agents are actually started
- if the shell also cant find it (exit code 127), its still correctly reported as missing
- applies to all providers, not just claude

limits:
- could not verify fix because I could not reproduce, but fix is a general detection improvement and should fix the issue

fixes #1111

### DIFF
--- a/src/main/services/ConnectionsService.ts
+++ b/src/main/services/ConnectionsService.ts
@@ -169,7 +169,7 @@ class ConnectionsService {
       return 'connected';
     }
 
-    if (result.status !== null && !result.timedOut) {
+    if (result.status !== null && !result.timedOut && (result.stdout || result.stderr)) {
       return 'connected';
     }
 


arnestrickmann:
Thanks Jona!

PR #1149, opened 2026-02-27
fix: scoped PR loading state to task path to prevent spinner leak

Valley-15:
Problem:
The PR creation loading spinner was using a global boolean state within the useCreatePR hook. This caused the spinner to "leak" across different tasks; if a user started creating a PR for one task and then switched to another, the new task would erroneously show a loading spinner.

Solution:

- Modified useCreatePR.tsx to use a keyed state object (Record<string, boolean>) that tracks the loading status for each specific taskPath independently.
 
- Updated the hook to return a function, isCreating(path), which allows components to query the loading status of a specific task.
 
- Updated FileChangesPanel.tsx to consume this new keyed state, ensuring that the "Create PR" button only shows a spinner if that specific task is currently in the process of creating a PR.

Verification:

- Verified that the TypeScript compilation errors in FileChangesPanel.tsx were resolved by correctly calling the new isCreating function with the safeTaskPath.
 
- Verified via code logic that switching tasks now correctly resets the UI spinner because the loading state is now unique to each task's file path.

### DIFF
--- a/src/renderer/components/FileChangesPanel.tsx
+++ b/src/renderer/components/FileChangesPanel.tsx
@@ -171,7 +171,9 @@ const FileChangesPanelComponent: React.FC<FileChangesPanelProps> = ({
       return 'create';
     }
   });
-  const { isCreating: isCreatingPR, createPR } = useCreatePR();
+  // NEW CODE: Rename the hook result and call it with the current path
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Thanks for opening up this pr! @Valley-15 
Another pr that fixed this was already merged. 
Thanks again.

PR #1150, opened 2026-02-27
perf: resolve lag when expanding/collapsing sidebar sections

Valley-15:
Problem:
Expanding and collapsing project sections in the left sidebar was noticeably laggy due to the full unmounting and remounting of task list subtrees.

Solution:

- Optimized CollapsibleContent to use forceMount combined with CSS display: none (data-[state=closed]:hidden). This keeps the DOM nodes alive and makes toggling instantaneous.
 
- Wrapped ProjectItem and MenuItemButton in React.memo to prevent unnecessary re-renders when parent state changes.

Verification:
Verified via code analysis that the expensive mounting lifecycle is bypassed during toggles. TypeScript checks passed with 0 errors.

### DIFF
--- a/src/renderer/components/LeftSidebar.tsx
+++ b/src/renderer/components/LeftSidebar.tsx
@@ -87,7 +87,6 @@ const getConnectionId = (project: Project): string | null => {
   return (project as any).sshConnectionId || null;
 };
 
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thanks! @Valley-15

PR #1164, opened 2026-02-27
fix: filter shell noise from PATH to prevent ENAMETOOLONG crashes

Amwam:
## Problem

Every `execFileAsync('git', ...)` call in the packaged Electron app crashes with `spawn ENAMETOOLONG` (errno -63) for users whose shell startup emits output to stdout.

### Root cause

`main.ts` reads the user's login shell PATH via `zsh -ilc 'echo -n $PATH'`. The `-il` (interactive login) flags cause `.zshrc` to run fully, which can emit **nvm version switching messages, colored ASCII art banners, MOTD text**, and other noise to stdout. This noise gets captured as part of the PATH value:

```
Found '/Users/user/project/.nvmrc' with version <22.20.0>
Now using node v22.20.0 (npm v10.9.3)
         [38;2;251;39;93m****[0m  [38;2;255;204;0m****[0m
       ... hundreds of bytes of ASCII art ...
/actual/path:/another/path:...
```

Since the noise has no `:` separators, it becomes one enormous ~1500+ character "path entry." When Node.js resolves binaries like `git` via PATH, it constructs `<garbage>/git` which exceeds `PATH_MAX` (1024) → `ENAMETOOLONG`.

### User-visible symptoms

- **Worktree creation fails** with misleading error: *"Local branch 'master' does not exist"* — because `hasRemote()` catches all errors and returns `false`, then `rev-parse` also fails with `ENAMETOOLONG` which is caught and rethrown as a "branch doesn't exist" error
- **All CLI agent providers show "not installed"** — because detection `spawn` calls all fail with the same error

### 1. Filter shell noise from PATH entries (`src/main/main.ts`)

After reading the login shell PATH, filter entries to only those starting with `/`. This discards any shell startup noise (nvm messages, ASCII art, MOTD) that gets mixed into stdout. Applied to both macOS and Linux code paths.

```ts
const allEntries = (loginPath + ':' + process.env.PATH).split(':').filter(Boolean);
const validEntries = allEntries.filter((p: string) => p.startsWith('/'));
const merged = new Set(validEntries);
process.env.PATH = Array.from(merged).join(':');
```

### 2. Propagate system errors instead of swallowing them (`src/main/services/WorktreeService.ts`)

- **`hasRemote()`**: Previously caught ALL errors and returned `false`. Now propagates `ENAMETOOLONG`, `ENOENT`, and `EACCES` so users get actionable errors instead of misleading "no remote found" messages.
- **`fetchBaseRefWithFallback()` rev-parse catch**: Same treatment — system-level spawn failures now surface as *"Git failed to run (ENAMETOOLONG)"* instead of *"branch does not exist"*.

## Test plan

- [x] `pnpm run type-check` — passes
- [x] `pnpm run lint` — passes (no new warnings)
- [x] `pnpm exec vitest run` — all 265 tests pass
- [ ] Manual: launch packaged app with a `.zshrc` that emits nvm/ASCII noise → verify worktree creation succeeds and providers are detected

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -37,8 +37,9 @@ if (process.platform === 'darwin') {
     const loginPath = execSync(`${shell} -ilc 'echo -n $PATH'`, { encoding: 'utf8' });
     if (loginPath) {
       // Shell noise (nvm messages, ASCII art, motd) gets captured in stdout.
-      // Filter to only entries that look like real paths (start with /).
-      const allEntries = (loginPath + ':' + process.env.PATH).split(':').filter(Boolean);
+      // Split by both : and \n so noise fused with the first real path entry
+      // (e.g. "nvm output\n/usr/local/bin") is correctly separated.
+      const allEntries = (loginPath + ':' + process.env.PATH).split(/[:\n]/).filter(Boolean);
       const validEntries = allEntries.filter((p: string) => p.startsWith('/'));
       const merged = new Set(validEntries);
       process.env.PATH = Array.from(merged).join(':');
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thanks for the thorough diagnosis and fix! The root cause analysis is spot-on and the fix is clean. LGTM, approving. @Amwam

PR #1169, opened 2026-02-28
fix(ui): make (Having trouble?) link clickable

thisisharsh7:
Makes the “Having trouble?” text in the GitHub device flow modal clickable, linking directly to the project's issues page.


arnestrickmann:
Good catch, sir

Issue #1172, opened 2026-02-28
[bug]: README download link for debian package 404's


kevin51jiang:
### Bug Description

README download link is wrong

### Steps to Reproduce

In the readme at the top, there is a link to: https://github.com/generalaction/emdash/releases/latest/download/emdash-x64.deb

1. Click on the link
2. redirects you to https://github.com/generalaction/emdash/releases/download/v0.4.20/emdash-x64.deb, which is a 404 page

### Actual vs Expected Behavior

Expected: send you to https://github.com/generalaction/emdash/releases/download/v0.4.20/emdash-amd64.deb

Note the amd64 on the working link, vs the x64.deb on the invalid link

### emdash Version

0.4.20

### Operating System

its on the website

arnestrickmann:
Thanks for flagging!

A fix will be live very soon.

arnestrickmann:
https://github.com/generalaction/emdash/pull/1176

kevin51jiang:
You speedy

PR #1176, opened 2026-02-28
fix: correct Linux .deb download link filename

arnestrickmann:
## Summary

- Fix the Debian package download URL in both `README.md` and `docs/content/docs/installation.mdx`
- Change filename from `emdash-x64.deb` to `emdash-amd64.deb` to match the actual release asset name

Fixes GEN-447


arnestrickmann:
fixes https://github.com/generalaction/emdash/issues/1172

PR #1179, opened 2026-02-28
feat(ui): add tooltip to folder toggle icon in sidebar

thisisharsh7:
This PR added tooltip to the folder icon in the left sidebar that toggles task visibility

Screenshot 2026-02-28 at 11 06 12 AM


arnestrickmann:
Thank you for the PR @thisisharsh7! I do think that we do not nessecarily need a tooltip for this folder icon as I think it is self explanatory.

But really appreciate it!

Issue #1181, opened 2026-02-28
[bug]: getWindowsEssentialEnv() missing standard Windows system variables breaks .NET/NuGet and other tools


reneg:
### Bug Description

Emdash's minimal environment whitelist for spawned processes (getWindowsEssentialEnv() in ptyManager.ts) is
  missing several standard Windows system environment variables. This causes .NET SDK, NuGet, and potentially
  other Windows development tools to fail inside emdash terminals.

  ### Affected Variables

  The following standard Windows system variables are not included in getWindowsEssentialEnv():

  ```
┌─────────────────────────┬─────────────────────────────────────┬─────────────────────────────────────────┐
  │        Variable         │            Typical Value            │                 Used By                 │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ ProgramFiles            │ C:\Program Files                    │ .NET, NuGet, MSBuild, many tools        │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ ProgramFiles(x86)       │ C:\Program Files (x86)              │ 32-bit tool resolution                  │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ ProgramData             │ C:\ProgramData                      │ NuGet machine-wide settings, installers │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ CommonProgramFiles      │ C:\Program Files\Common Files       │ Various Windows APIs                    │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ CommonProgramFiles(x86) │ C:\Program Files (x86)\Common Files │ 32-bit common files                     │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ ProgramW6432            │ C:\Program Files                    │ WoW64 compatibility                     │
  ├─────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────────┤
  │ CommonProgramW6432      │ C:\Program Files\Common Files       │ WoW64 compatibility                     │
  └─────────────────────────┴─────────────────────────────────────┴─────────────────────────────────────────┘
```

### Steps to Reproduce

1. Open a project in emdash on Windows
2. Start a task terminal
3. Run dotnet build (or dotnet restore) on any .NET project

### Actual vs Expected Behavior

Expected Behavior

  dotnet build succeeds.

[...]

PR #1183, opened 2026-02-28
fix(pty): pass Windows Program Files env vars to PTY sessions

arnestrickmann:
## Summary
- Adds `ProgramFiles`, `ProgramFiles(x86)`, `ProgramData`, `CommonProgramFiles`, `CommonProgramFiles(x86)`, `ProgramW6432`, and `CommonProgramW6432` to the Windows essential environment variables passed through to PTY sessions

## Problem
On Windows, PTY sessions were missing Program Files path environment variables. This caused .NET, NuGet, MSBuild, and other tools that rely on these variables to fail or behave incorrectly when invoked from agent shells.

## Solution
Extended `getWindowsEssentialEnv()` in `ptyManager.ts` to include all standard Windows Program Files directory variables with sensible defaults, matching the approach already used for other essential Windows env vars like `SystemRoot` and `APPDATA`.

Fixes GEN-450


arnestrickmann:
fixes https://github.com/generalaction/emdash/issues/1181

Issue #1193, opened 2026-03-01
[feat]: auto copy on terminal selection


liamhess:
### Feature Summary

Auto-copy on selection in terminals.

### Problem or Use Case

In iTerm2 I have since basically ever activated auto copy on selection and it's a workflow I have really gotten used to.
I find myself every time I want to copy something from a terminal in emdash: selecting, trying to paste in another app just to find out that I forgot to cmd+c before.

### Proposed Solution

Add an opt in setting in the interface settings below the terminal font setting to toggle on or off saying something like `Auto-copy on select`.
Then the terminals should start doing that.

### Additional Context

I've created a PR that implements that. It is fully written in emdash with opencode and glm 5, but fully reviewed by me. However still consider that this is my first PR in this repository and I don't know a lot about the project setup in general.

arnestrickmann:
Hi @liamhess,

Thanks for opening this issue.

PR #1194, opened 2026-03-01
feat: add auto copy on select

liamhess:
Fixes https://github.com/generalaction/emdash/issues/1193 by adding auto copy on selection.


arnestrickmann:
Hi @liamhess,

Thanks for opening the PR - the addition makes sense. 

LGTM, merging.

PR #1199, opened 2026-03-01
fix(windows): run .cmd/.bat provider CLIs via cmd.exe to avoid silent startup failures

fALECX:
## Summary
On Windows, several provider CLIs installed via npm resolve to `.cmd` shims in the global npm bin directory. Spawning those shims directly can result in the process appearing to start but never producing useful output.

This change updates Windows process launching so `.cmd` and `.bat` commands are executed through `cmd.exe`, which is how these shims are intended to run.

## What changed
- Updated PTY launch path to wrap `.cmd` / `.bat` with:
  - `cmd.exe /d /s /c <fullCommandString>`
- Updated provider detection command execution to apply the same Windows shim handling.
- Added explicit warning logs for spawn errors and failed exits so failures are visible instead of silent.

## Why this fixes the issue
npm global binaries on Windows are command shims, not native executables. Running them through `cmd.exe` ensures providers like Codex, Copilot, and other npm-based CLIs launch reliably.

## Scope
- Minimal, targeted change (2 files)
- Windows-specific behavior only
- Provider-agnostic logic (not tied to a single agent)

## Testing
- Manual validation in this environment is limited by local engine mismatch (`Node v24`, `pnpm v9` vs repo requirements).
- Code path review confirms both runtime spawn and provider detection now handle `.cmd`/`.bat` consistently on Windows.

## Files
- `src/main/services/ptyManager.ts`
- `src/main/services/ConnectionsService.ts`

### DIFF
--- a/src/main/services/ConnectionsService.ts
+++ b/src/main/services/ConnectionsService.ts
@@ -42,6 +42,15 @@ const truncate = (input: string, max = 400): string =>
 
 const DEFAULT_TIMEOUT_MS = 3000;
 
+const quoteForCmdExe = (input: string): string => {
+  if (input.length === 0) return '""';
+  if (!/[\s"^&|<>()%!]/.test(input)) return input;
+  return `"${input
+    .replace(/%/g, '%%')
+    .replace(/!/g, '^!')
+    .replace(/(["^&|<>()])/g, '^$1')}"`;
+};
+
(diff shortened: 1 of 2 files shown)


fALECX:
Forgot to add: Problem was that CLI Agents are installed and recognized in emdhas, however upon opening an agent it would show an empty console. See problem above on as to why. Manual verification solved this for Claude and Codex specifically.

arnestrickmann:
Good catch, and thank you! @fALECX

PR #1203, opened 2026-03-01
feat(diff-viewer): replace modal diff viewer with full-screen inline diff viewer

jschwxrz:
summary:
- the old diff viewer was two modals with no commit history, no git operations, and no way to browse past changes replaced with a full-screen inline diff viewer with changes tab, history tab, and builtin commit/push/pull
- added 8 new git IPC endpoints and hardened existing ones with path validation and shell injection fixes

changes:
- diff viewer renders full-screen in the center panel instead of modals
- changes tab differentiates between tracked and untracked files
- history tab for browsing commits and per-commit file diffs
- commit area moved into the diff viewer with separate commit and push steps
- soft reset for unpushed commits and pull support
- new git backend: commit, push, pull, log, commit-files, commit-file-diff, soft-reset, latest-commit

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -349,6 +349,29 @@ jobs:
           console.log('Wrote dist/main/appConfig.json');
           NODE
 
+      - name: Check Azure Trusted Signing secrets
+        id: signing
+        shell: bash
+        env:
+          AZ_TENANT: ${{ secrets.AZURE_TENANT_ID }}
+          AZ_CLIENT: ${{ secrets.AZURE_CLIENT_ID }}
+          AZ_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
+        run: |
+          if [ -n "$AZ_TENANT" ] && [ -n "$AZ_CLIENT" ] && [ -n "$AZ_SECRET" ]; then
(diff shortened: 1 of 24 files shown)


jschwxrz:
@greptileai

arnestrickmann:
Good stuff, @jschwxrz!

We thought about (@rabanspiegel), going forward, that it could also make sense to move the component on the left to the right - but we can do this at a later point - wdyt? @jschwxrz 

CleanShot 2026-03-01 at 15 03 11@2x

There is one bug I discovered while testing. 

When I start the app, I am directed into the Home View. When I click on a task from there, the task doesnt open. 
When I first click on the parent project in the left sidebar and then click on one of the task names inside the project main view, the chat for this task correctly opens. 

This would need to be fixed before we merge. Otherwise, really nice.

arnestrickmann:
I notice, that the bug I outlined is a general issue. 

So I'm merging.

jschwxrz:
right is where committing happened before, so we could move it there. 

but I think keeping it left would make more sense, because the visual flow matches reading direction (left-right). If the sidebar is on the left, the user can progressively move their eyes from left -> right. If it's on the right, it is left -> right -> left -> right. 

@arnestrickmann @rabanspiegel  wdyt

PR #1224, opened 2026-03-02
Add TMPDIR to PTY spawn modes

### LINKED ISSUE #1213: [bug]: CLI tools that rely on TMPDIR hang silently in Emdash terminals
### Bug Description

CLIs that rely on `TMPDIR` for IPC discovery (e.g. Unix sockets, lock files) hang silently when run inside Emdash  terminals.

macOS sets `TMPDIR` to a per-user directory (`/var/folders/.../T/`). Emdash's minimal PTY environment omits it, so  tools fall back to /tmp or block indefinitely trying to locate sockets/discovery files that live under the real
 `TMPDIR`.

### Steps to Reproduce

Run any CLI that uses `TMPDIR`-based IPC (e.g. Obsidian CLI) in an Emdash terminal.

It hangs with no output.

Same command works in a regular terminal.

### emdash Version

v0.4.22

### Operating System

macOS 26.3

### Additional Context

Fix: Pass through `TMPDIR` in all three PTY spawn modes, same pattern as `LANG/SSH_AUTH_SOCK`.

sairus2k:
Closes #1213


arnestrickmann:
Thanks for flagging and for opening the PR!

arnestrickmann:
could you resolve the merge conflicts? 
would be amazing, thanks.

PR #1230, opened 2026-03-02
refactor: centralize app settings via typed RPC and shared provider (4)

Davidknp:
### Summary

- **Introduce a typed RPC layer for app settings** — replaces the old `settings:get` / `settings:update` IPC channels with a `createRPCController` handler (`appSettings.get` / `appSettings.update`) whose return types flow end-to-end from `settings.ts` through to the renderer with no `any` casts.
- **Create `AppSettingsProvider`** — a single React context backed by TanStack Query that owns fetching, caching, optimistic updates, and rollback for app settings. All 15+ settings components now call `useAppSettings()` instead of each managing their own `useState` + `useEffect` + `rpc` call.
- **Introduce `DeepPartial<AppSettings>` / `AppSettingsUpdate`** — typed deep-partial update shape so callers only pass the fields they're changing; the server-side `deepMerge` does the rest, eliminating the bug where components spread stale cached values on every save.
- **Remove 272 lines from `electron-api.d.ts`** — the hand-maintained `getSettings` / `updateSettings` declarations and their duplicated types (`ProviderCustomConfig`, etc.) are deleted; the ground truth lives in `settings.ts` now.
- **Net result: −1,132 lines, +353 lines** across 34 files.

### Why this was necessary

Every settings component had its own copy of the same pattern:

```ts
const [value, setValue] = useState(DEFAULT);
useEffect(() => { rpc.appSettings.get().then(setValue) }, []);
const save = async (next) => { setSaving(true); await rpc.appSettings.update(...); setSaving(false); };
```

This caused several concrete problems:

1. **Stale data bugs** — each component held its own in-memory snapshot; toggling one setting could silently overwrite another field changed elsewhere.
2. **No optimistic updates** — the UI lagged behind user input while waiting for the IPC round-trip.
3. **`settings:get` / `settings:update` errors** — several components called the old `window.electronAPI.getSettings()` API that no longer had a registered handler after a prior IPC refactor, causing visible console errors on load.
4. **Duplicated types** — `ProviderCustomConfig`, `AppSettings` sub-types, and related shapes were declared in both `settings.ts` and `electron-api.d.ts`, drifting out of sync.
5. **`any` return types** — the `@/shared/ipc/rpc` import in `settingsIpc.ts` used the `@/*` alias which resolves differently under the renderer's `tsconfig.json`, causing the entire `RpcRouter` type to collapse to `any` in the renderer.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -365,13 +365,6 @@ jobs:
             echo "::warning::Azure Trusted Signing secrets not configured. Windows build will be unsigned."
           fi
 
-      - name: Install Azure Trusted Signing dependencies
-        if: ${{ steps.signing.outputs.has_signing == 'true' }}
-        shell: pwsh
-        run: |
-          Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force
-          Install-Module -Name TrustedSigning -RequiredVersion 0.5.0 -Force -Repository PSGallery
-
       - name: Rebuild native modules (x64)
         shell: bash
         run: |
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "emdash",
-  "version": "0.4.22",
+  "version": "0.4.23",
   "description": "A cross-platform Electron app that orchestrates multiple coding agents in parallel",
   "main": "dist/main/main/entry.js",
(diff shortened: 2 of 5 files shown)


arnestrickmann:
sick refactor

PR #1245, opened 2026-03-03
fix(diff): make diff viewer more robust for large files

jschwxrz:
summary:
- diff viewer reconstructed file content from git diff --unified=2000 output
- breaks for files >4000 lines because the context window doesn't cover the entire file
- deleted-file detection was also broken: .every() on empty arrays returns vacuously true

fix:
- fetch full files at each revision via git show instead of reconstructing from diff output
- run git show in parallel with git diff so there's no latency cost
- fixed deleted file detection by tracking whether fsRead succeeded before blanking modified content

### DIFF
--- a/src/main/services/GitService.ts
+++ b/src/main/services/GitService.ts
@@ -2,10 +2,16 @@ import { execFile } from 'child_process';
 import { promisify } from 'util';
 import * as fs from 'fs';
 import * as path from 'path';
+import {
+  parseDiffLines,
+  stripTrailingNewline,
+  MAX_DIFF_CONTENT_BYTES,
+  MAX_DIFF_OUTPUT_BYTES,
+} from '../utils/diffParser';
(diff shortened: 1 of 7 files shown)


jschwxrz:
@greptileai

jschwxrz:
@greptileai

jschwxrz:
@greptileai

jschwxrz:
@greptileai

jschwxrz:
@greptileai

jschwxrz:
@arnestrickmann

I feel like greptile always finds something haha I do not really see the point of rerunning the reviews anymore. 

the two issues raised by greptile are not really valid:

1: a files entire content would need to be exactly __EMDASH_TOO_LARGE__ to trigger a false positive, not realistic
2: stat failure on remote hosts: every linux system has stat. if a remote host doesn't have it the ssh session would have bigger problems

arnestrickmann:
Thanks, @jschwxrz !

PR #1248, opened 2026-03-03
fix: allow terminal switching while lifecycle run script is active

### LINKED ISSUE #1246: [bug]: Cannot switch terminal if in "Run Lifecycle" and started the run script
### Bug Description

When starting the run script, the "Run Lifecycle" terminal opens, and if i open the dropdown to select another terminal, it flickers and stays in the "Run" terminal.
I need to stop the run script to be able to switch to another terminal.

### Steps to Reproduce

1. create a task with a run script
2. click the run button
3. open the dropdown and try to switch to another terminal within the same task
4. it wont work
5. stop the run script
6. try to switch again
7. that works

actual: not able to switch away from the "run" terminal
expected: be able to switch between different terminals, maybe if the run lifecycle is "running" that could be visible

### emdash Version

0.4.23

### Operating System

macos 26.3

### Additional Context

_No response_

jschwxrz:
summary:
- terminal dropdown won't switch away from "Run" while a lifecycle run script is active
- auto-switch effect keeps firing and snapping the selection back

fix:
- only auto-switch to the run terminal once, when the run first starts, instead of continuously while it's running

fixes #1246

### DIFF
--- a/src/renderer/components/TaskTerminalPanel.tsx
+++ b/src/renderer/components/TaskTerminalPanel.tsx
@@ -235,12 +235,15 @@ const TaskTerminalPanelComponent: React.FC<Props> = ({
     };
   }, [task?.id, refreshLifecycleState]);
 
-  // Auto-switch dropdown to Run when the run phase starts.
+  // Auto-switch dropdown to Run when the run phase first starts.
+  const prevRunStatusRef = useRef(runStatus);
   useEffect(() => {
-    if (runStatus === 'running' && selection.selectedLifecycle !== 'run') {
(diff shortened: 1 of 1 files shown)


jschwxrz:
@greptileai your review fabricates changes:

"Refactors selection state management into a useTerminalSelection hook, improving code organization." - already existed before

"Adds support for remote connections and improves terminal pane rendering." - not part of this PR at all

jschwxrz:
@greptileai will you update your review comment? Or will it stay like this?

arnestrickmann:
Great catch and fix!

PR #1249, opened 2026-03-03
fix: prevent worktree branches from auto-tracking base ref

dev-seahouse:
When creating worktrees, Git's branch.autoSetupMerge setting causes new branches to incorrectly track the base ref (e.g., origin/develop). This breaks the local-only workflow where users expect:

1. Branch created from base ref with no tracking set
2. Work locally on the branch
3. When pushing, tracking is set to origin/<branch-name>

Add --no-track flag to git worktree add commands in both WorktreeService and WorktreePoolService to prevent auto-tracking. Tracking is set explicitly via push --set-upstream after creation.


arnestrickmann:
Thanks for catching this edge case

PR #1272, opened 2026-03-04
feat(ui): always show changes button in file changes panel top bar

jschwxrz:
summary:
- the Changes button (opens diff viewer) only rendered when there were file changes
- since the diff viewer now also provides access to history, it should always be accessible

changes:
- unified the two separate top-bar layouts in FileChangesPanel.tsx into one consistent layout
- changes button renders unconditionally, no longer gated behind hasChanges
- empty state shows emdashes in green/red (— · —) instead of a static "Changes" text label
- all PR logic (create, draft, view, merge) unchanged


arnestrickmann:
Great catch, and makes a lot of sense. Thanks!

PR #1274, opened 2026-03-04
fix(ssh): resolve macOS SSH_AUTH_SOCK detection for 1Password agent

### LINKED ISSUE #1235: [bug]: macOS — ssh2js ignores SSH_AUTH_SOCK (1Password agent)
### Bug Description

Hi,

For the love of god and everything holy I couldnt get the 1password agent on my mac work with emdash. Funny thing is on the Windows machine I was using it works perfectly (a few tweaking where the socket paths are) but on mac no bueno.

Please find below steps to reproduce and actual behaviour. I have added my opinion about possibly cause, but cant tell.

### Steps to Reproduce

Steps to reproduce

Configure 1Password SSH agent on macOS
Set SSH_AUTH_SOCK to 1Password's socket via:

~/.zshrc export
launchctl setenv
LaunchAgent plist at ~/Library/LaunchAgents/
Full reboot

Verify from terminal: ssh-add -L lists keys, ssh user@host connects fine
Add SSH remote project in Emdash using "Agent" auth
Attempt connection

Expected
Emdash's ssh2js reads SSH_AUTH_SOCK, connects to 1Password's agent socket, retrieves keys, and authenticates.

Actual
Emdash log shows:
Agent: No more keys left to try
Client: agent auth failed
Outbound: Sending DISCONNECT (11)
ssh2js appears to connect to the default macOS agent (which has no keys) instead of 1Password's agent, regardless of the SSH_AUTH_SOCK value in the environment.
Verification that SSH_AUTH_SOCK is set correctly
$ launchctl getenv SSH_AUTH_SOCK
/Users/gergelycsata/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock

$ ls -la "$HOME/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock"
srw-------  1 gergelycsata  staff  0 Feb 28 15:12 /Users/.../agent.sock

$ ssh-add -L  # (in terminal with SSH_AUTH_SOCK set)
ssh-ed25519 AAAAC3... 
ssh-ed25519 AAAAC3... 
ssh-ed25519 AAAAC3... 
(6 keys total)
Everything works from the terminal. Only Emdash's ssh2js fails to use the correct agent.
Suspected cause

### emdash Version

v0.4.23

### Operating System

Mac OS Tahoe 26.3

### Additional Context

_No response_

jschwxrz:
summary:
- 1Password exposes its SSH agent via IdentityAgent under Host * in ~/.ssh/config, but the parser discarded all wildcard host blocks
- on macOS, GUI-launched apps inherit the default Apple agent socket (no keys) instead of the user's 1Password override

fix:
- parse Host * blocks and use their IdentityAgent as fallback when no specific host matches
- on macOS, check launchctl getenv SSH_AUTH_SOCK before trusting the inherited process.env

limits:
- could not e2e verify with 1Password on a real remote server

fixes #1235

### DIFF
--- a/src/main/ipc/sshIpc.ts
+++ b/src/main/ipc/sshIpc.ts
@@ -719,7 +719,9 @@ export function registerSshIpc() {
         }
 
         const hosts = await parseSshConfigFile();
-        const host = hosts.find((h) => h.host.toLowerCase() === hostAlias.toLowerCase());
+        const host = hosts
+          .filter((h) => !h.host.includes('*') && !h.host.includes('?'))
+          .find((h) => h.host.toLowerCase() === hostAlias.toLowerCase());
 
         if (!host) {
           return { success: false, error: `Host alias not found: ${hostAlias}` };
--- a/src/main/utils/__tests__/shellEnv.test.ts
+++ b/src/main/utils/__tests__/shellEnv.test.ts
@@ -77,7 +77,7 @@ describe('shellEnv', () => {
       expect(result).toBe('/existing/socket');
     });
 
-    it('should detect SSH_AUTH_SOCK from shell when not set', () => {
+    it('should detect SSH_AUTH_SOCK when not in process.env', () => {
       delete process.env.SSH_AUTH_SOCK;
       mockedExecSync.mockReturnValue('/shell/detected/socket');
 


jschwxrz:
@greptileai

PR #1278, opened 2026-03-04
feat(lifecycle): preserve lifecycle terminal logs across task switches

jschwxrz:
summary:
- switching to another task and back clears lifecycle (run/setup/teardown) terminal output, showing "no lifecycle output yet"
- log lines were stored only in React component state, which resets on task switch
- regular PTY terminals survive switches via disk-based snapshots, but lifecycle output had no persistence mechanism

changes:
- buffer log lines in TaskLifecycleService (main process, capped at 300 lines per phase)
- added lifecycle:getLogs IPC handler so the renderer can fetch buffered logs on task switch
- extracted shared formatLifecycleLogLine() helper and MAX_LIFECYCLE_LOG_LINES constant into src/shared/lifecycle.ts to deduplicate formatting logic between main and renderer

### DIFF
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -247,6 +247,7 @@ contextBridge.exposeInMainWorld('electronAPI', {
   lifecycleTeardown: (args: { taskId: string; taskPath: string; projectPath: string }) =>
     ipcRenderer.invoke('lifecycle:teardown', args),
   lifecycleGetState: (args: { taskId: string }) => ipcRenderer.invoke('lifecycle:getState', args),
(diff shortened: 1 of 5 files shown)


jschwxrz:
the race condition here isn't reachable. ipcMain.handle responses and webContents.send go through the same ordered Chromium IPC pipe. either the event was emitted before getLogs read the buffer (so it's already in the snapshot), or it was emitted after (so the invoke response arrives at the renderer first, and the listener appends on top). the live event can never overtake the invoke response

PR #1284, opened 2026-03-04
feat: instant PR detection when created via   
  CLI

aryanpatel-ctrl:
## Summary                                                                                                                                               
  - Detect GitHub PR URLs in terminal output instantly when created via CLI                                                                                             
  - Trigger immediate PR status refresh instead of waiting for 30-second poll cycle                                                                                     
  - PR detection now works identically whether created via UI button or CLI                                                                                             
                                                                                                                                                                        
  ## Changes                                                                                                                                                            
  - ptyIpc.ts: Added PR URL pattern matching on PTY output                                                                                                              
  - ptyManager.ts: Added getPtyCwd() to retrieve task path                                                                                                              
  - preload.ts: Exposed onPtyPrUrlDetected listener                                                                                                                     
  - useAutoPrRefresh.ts: Added event-driven refresh trigger                                                                                                             
  - Type definitions and test coverage added                                                                                                                            
                                                                                                                                                                        
  ## Test plan                                                                                                                                                          
  - [ ] Run gh pr create in agent terminal                                                                                                                              
  - [ ] Verify changes panel updates immediately

### DIFF
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -135,6 +135,15 @@ contextBridge.exposeInMainWorld('electronAPI', {
     ipcRenderer.on(channel, wrapped);
     return () => ipcRenderer.removeListener(channel, wrapped);
   },
+  onPtyPrUrlDetected: (listener: (data: { id: string; url: string; cwd?: string }) => void) => {
+    const channel = 'pty:pr-url-detected';
+    const wrapped = (
+      _: Electron.IpcRendererEvent,
+      data: { id: string; url: string; cwd?: string }
+    ) => listener(data);
+    ipcRenderer.on(channel, wrapped);
+    return () => ipcRenderer.removeListener(channel, wrapped);
+  },
   onAgentEvent: (listener: (event: AgentEvent, meta: { appFocused: boolean }) => void) => {
     const channel = 'agent:event';
     const wrapped = (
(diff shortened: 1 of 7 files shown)


aryanpatel-ctrl:
@greptileai review

arnestrickmann:
Hey, thanks for working on this! The instant detection is a great idea - waiting up to 30s after gh pr create is a real pain point.

A couple of things to address:

1. Merge conflicts in ptyIpc.ts and ptyManager.ts need resolving first.
2. Simplification opportunity — the tail buffer and cooldown map add quite a bit of complexity and introduce some of
the bugs Greptile flagged (re-firing after cooldown expires, single-entry map overwriting). The good news is most of
that isn't needed since refreshPrStatus already deduplicates concurrent requests. The detection can be as simple as:

```
const PR_URL_RE = /https?:\/\/github\.com\/[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+\/pull\/\d+/;

  function maybeEmitPrUrlDetected(id: string, chunk: string): void {
    const match = chunk.match(PR_URL_RE);
    if (match) {
      safeSendToOwner(id, 'pty:pr-url-detected', { id, url: match[0], cwd: getPtyCwd(id) });
    }
  }
```

URL splitting across chunks is practically impossible with the 16ms flush interval, and the 30s poll is the backstop anyway.

3. Small renderer fix — in useAutoPrRefresh, the event?.cwd || activeTaskPath fallback can refresh the wrong task for SSH PTYs where cwd is always undefined. Swapping to if (!event?.cwd) return; avoids that.

These changes would bring the diff down to ~30-40 lines while keeping the full feature. Happy to help if you have any questions!

-

Closing this for now — please feel free to reopen with the suggested changes, happy to review again!

PR #1300, opened 2026-03-05
fix(sidebar): auto collapse right sidebar on settings and skills views

### LINKED ISSUE #1297: [bug]: right sidebar auto-collapse behaves inconsistently depending on previous view
### Bug Description

When the **auto-collapse right sidebar** setting (`interface.autoRightSidebarBehavior`) is enabled, the right sidebar does not consistently collapse on non-task views. The behavior depends on which view the user was previously on.

### Steps to Reproduce

1. Enable "Auto-collapse right sidebar" in Settings → Interface
2. Navigate to a task (right sidebar opens with file changes + terminal)
3. Click on Skills or Settings
4. Observe: the right sidebar remains open

Compare with:
1. Be on the Home view (right sidebar is collapsed)
2. Click on Skills or Settings
3. Observe: the right sidebar stays collapsed (correct)

**Actual:** The right sidebar stays open when navigating from a task to Skills/Settings, but stays collapsed when navigating from Home to Skills/Settings.

**Expected:** The right sidebar should consistently collapse on all non-task views (Home, Skills, Settings, project home) regardless of which view the user came from.

### emdash Version

0.4.25

### Operating System

macOS 26.3

### Additional Context

The auto-collapse logic in `usePanelLayout` only checks for `showHomeView` and "repo home page" (project selected + no active task). It doesn't account for Skills or Settings as non-task views, so the sidebar remains in whatever state it was in when leaving a task.

jschwxrz:
summary:
- right sidebar auto-collapse only checked for home view and repo home page
- navigating from a task to Settings or Skills left the sidebar open because those views weren't considered non-task views

fix:
- added showSettingsPage and showSkillsView to usePanelLayout options
- shouldCollapse now includes an isNonTaskView check so the sidebar collapses on all non-task views consistently

fixes #1297

### DIFF
--- a/src/renderer/hooks/usePanelLayout.ts
+++ b/src/renderer/hooks/usePanelLayout.ts
@@ -14,6 +14,8 @@ export interface UsePanelLayoutOptions {
   showDiffViewer: boolean;
   isInitialLoadComplete: boolean;
   showHomeView: boolean;
+  showSettingsPage: boolean;
+  showSkillsView: boolean;
   selectedProject: { id: string } | null;
   activeTask: { id: string } | null;
 }
@@ -24,6 +26,8 @@ export function usePanelLayout(opts: UsePanelLayoutOptions) {
     showDiffViewer,
     isInitialLoadComplete,
     showHomeView,
+    showSettingsPage,
+    showSkillsView,
     selectedProject,
     activeTask,
   } = opts;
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Good catch, and thanks for the fix!

PR #1308, opened 2026-03-05
feat: add Forgejo integration

Codycody31:
Adds Forgejo as a supported forge integration, following the same pattern as the existing GitLab integration.

Confirmed the following:
- Connect to a self-hosted Forgejo instance
- Verify that issues load and search works in task creation
- Disconnect and reconnect

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "emdash",
-  "version": "0.4.26",
+  "version": "0.4.27",
   "description": "A cross-platform Electron app that orchestrates multiple coding agents in parallel",
   "main": "dist/main/main/entry.js",
   "packageManager": "pnpm@10.28.2",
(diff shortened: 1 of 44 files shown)


arnestrickmann:
Thanks for putting this together @Codycody31! The structure looks solid — you clearly modeled it on the GitLab integration which is the right approach.

A few things we'd like addressed before we can merge:

- seedIssueContext in taskCreationService.ts doesn't handle forgejoIssue — tasks linked to a Forgejo issue won't show
  the issue context card in the conversation UI. (The hasIssueContext guard needs updating too.) Note: this same gap
  exists for gitlabIssue, so bonus points if you fix both while you're in there.

- The ForgejoIssueSummary interface is duplicated between ForgejoService.ts and types/forgejo.ts — ideally this would
  live in src/shared/ so main and renderer share one definition.

- There are conflicts in 6-8 files from the Plain integration (#1339) that landed recently — those would also need
  resolving.

Thanks again for your interest in contributing!

Codycody31:
Should be good now?

arnestrickmann:
@Codycody31 reviewing!

PR #1330, opened 2026-03-06
fix(lifecycle): surface errors from lifecycle actions and auto-run setup

HajekTim:
## Summary

Lifecycle play button actions (setup/run/teardown) silently swallowed errors, leaving users with no feedback when something failed. Most notably, clicking "Run" after an app restart with a setup script configured would silently fail because in-memory lifecycle state resets to `idle`.

This PR fixes three problems:
- **Silent failures**: errors from lifecycle actions were never shown to the user
- **Greyed-out button after failure**: the Run button was permanently disabled after setup failed, with no way to retry except restarting the app
- **No debug info**: error messages like "Exited with code 1" gave no actionable information

### Changes

| File | What |
|------|------|
| `src/renderer/components/TaskTerminalPanel.tsx` | Show toast with "View logs" action on lifecycle failures; show retry icon (RotateCw) when setup failed; keep Run button enabled after failure; surface IPC exceptions |
| `src/main/services/TaskLifecycleService.ts` | Auto-run setup when status is `idle` or `failed` (retry); include last 5 lines of script output in error results for debugging |

### Detailed changes

**Error surfacing (TaskTerminalPanel.tsx):**
- Check return value of all lifecycle IPC calls (`lifecycleSetup`, `lifecycleRunStart`, `lifecycleTeardown`)
- Show destructive toast on failure with a **"View logs"** action button that navigates to the relevant lifecycle tab (e.g. Setup) where full script output is visible
- When Run fails due to setup, the toast navigates to the Setup tab specifically
- Surface IPC-level exceptions (main process crash, serialization errors) as toasts, not just `console.error`
- Show toast on any `!result.success`, with fallback text when error string is empty

**Retry support (both files):**
- Remove `setupStatus !== 'failed'` from `canStartRun` guard so the Run button stays enabled after setup failure
- Show `RotateCw` icon instead of `Play` when setup has failed, making it clear clicking will retry
- Tooltip reads "Retry setup and start run script" in this state
- `startRunInternal` auto-runs setup when status is `idle` (after restart) or `failed` (retry)

**Actionable error messages (TaskLifecycleService.ts):**
- `buildErrorDetail()` appends the last 5 lines of script output to error messages
- Toast description shows the first line; clicking "View logs" shows the full output in the lifecycle tab

### Root cause

Lifecycle state (`setup.status`, `run.status`) is stored in-memory only (`private states = new Map<>()`). After an app restart, all states reset to `idle`. If a project has a setup script configured in `.emdash.json`, clicking "Run" would hit the guard at `startRunInternal` which checks `setupStatus !== 'succeeded'` and silently returned `{ ok: false, error: 'Setup has not completed yet' }`. The renderer never checked this return value.

## Test plan

- [ ] Configure `.emdash.json` with a setup script: `"setup": "echo 'setting up' && sleep 1"`
- [ ] Start a task, click the **Run** play button — verify setup auto-runs before run starts
- [ ] Restart emdash, open the same task, click **Run** — verify it auto-runs setup instead of silently failing
- [ ] Set a failing setup: `"setup": "echo 'bad config' && exit 1"` — click **Run**:
  - [ ] Verify destructive toast appears with "Setup failed" title
  - [ ] Verify toast description includes script output (e.g. "bad config")
  - [ ] Verify "View logs" button navigates to the Setup lifecycle tab
  - [ ] Verify the play button shows a **retry icon** (↻) instead of ▶
  - [ ] Verify the button is **not greyed out** — clicking retries setup
  - [ ] Verify tooltip reads "Retry setup and start run script"
- [ ] Click **Setup** directly with a failing script — verify error toast with "View logs"
- [ ] Click **Teardown** with `"teardown": "exit 1"` — verify error toast with "View logs"
- [ ] Verify successful lifecycle actions show no toast (no regression)

### DIFF
--- a/src/main/services/TaskLifecycleService.ts
+++ b/src/main/services/TaskLifecycleService.ts
@@ -161,6 +161,18 @@ class TaskLifecycleService extends EventEmitter {
     }
   }
 
+  private buildErrorDetail(taskId: string, phase: LifecyclePhase, baseError: string): string {
+    const buf = this.logBuffers.get(taskId);
+    const lines = buf?.[phase] ?? [];
+    // Grab last few non-empty output lines for context
+    const tail = lines
+      .map((l) => l.replace(/^\[.*?\]\s*/, '').trim())
+      .filter(Boolean)
+      .slice(-5);
+    if (tail.length === 0) return baseError;
+    return `${baseError}\n${tail.join('\n')}`;
+  }
+
   private emitLifecycleEvent(
     taskId: string,
     phase: LifecyclePhase,
@@ -232,8 +244,9 @@ class TaskLifecycleService extends EventEmitter {
             untrackFinite();
             const message = error?.message || String(error);
             this.emitLifecycleEvent(taskId, phase, 'error', { error: message });
+            const detail = this.buildErrorDetail(taskId, phase, message);
             finish(
-              { ok: false, error: message },
+              { ok: false, error: detail },
               {
                 ...state[phase],
                 status: 'failed',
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Good catch and thanks for the PR! 

Looks good to me, merging.

PR #1347, opened 2026-03-07
feat: add Novita AI as optional LLM provider

Alex-wuhu:
feat: add Novita AI as optional LLM provider

This PR adds Novita AI as an optional LLM provider to Emdash. It uses the `opencode-ai` CLI agent as a base and configures it to use Novita's OpenAI-compatible endpoint.

### Features
- New `novita` provider registered in `src/shared/providers/registry.ts`.
- `NOVITA_API_KEY` pass-through in `src/main/services/ptyManager.ts`.
- Automatic configuration of `OPENAI_BASE_URL` for the `novita` provider.
- UI assets and configuration updated.

### Testing
- Verified provider definition in `registry.ts`.
- Verified environment variable injection in `ptyManager.ts`.

### DIFF
--- a/src/shared/providers/registry.ts
+++ b/src/shared/providers/registry.ts
@@ -21,7 +21,6 @@ export const PROVIDER_IDS = [
   'mistral',
   'pi',
   'autohand',
-  'novita',
 ] as const;
 
 export type ProviderId = (typeof PROVIDER_IDS)[number];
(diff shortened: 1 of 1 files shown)


arnestrickmann:
@Alex-wuhu 
Thanks for opening the PR! 

Is Novita AI your default provider these days?

arnestrickmann:
Thanks for the PR. I don’t think this is ready to merge yet.

  On the current PR head, novita was removed from the provider registry in src/
  shared/providers/registry.ts, but it is still referenced in src/renderer/lib/
  agentConfig.ts and src/renderer/providers/assets.ts. That leaves the branch
  inconsistent and it currently fails pnpm run type-check locally.

  Also, even in the earlier version, the Novita setup in src/main/services/
  ptyManager.ts still looked incomplete:

  - novita reuses the OpenCode CLI, but it doesn’t get the same OpenCode hook/
    runtime setup.
  - OPENAI_BASE_URL is switched to Novita, but OPENAI_API_KEY is only overridden
    when NOVITA_API_KEY is present, which could lead to the wrong key being
    used.

  Happy to re-review once those pieces are fixed.

PR #1350, opened 2026-03-07
fix(security): harden SSH command validation and patch vulnerable minimatch resolutions

bhaktofmahakal:
## Summary

This PR fixes two security issues:

1. Hardens the renderer-exposed SSH command IPC path by rejecting shell control characters before command execution.
2. Patches vulnerable `minimatch` resolutions across both direct and transitive dependencies.

### SSH command validation
- Extracted SSH renderer command validation into a dedicated utility.
- Preserved the existing allowlist behavior for safe commands.
- Added rejection for shell control characters such as:
  - `;`
  - `&`
  - `|`
  - `<`
  - `>`
  - `` ` ``
  - `$`
  - newlines / carriage returns

### minimatch dependency remediation
- Bumped direct `minimatch` dependency to `10.2.3`.
- Added narrow `pnpm.overrides` to force patched transitive versions:
  - `minimatch@3` → `3.1.4`
  - `minimatch@5` → `5.1.8`
  - `minimatch@9` → `9.0.7`
  - `minimatch@10` → `10.2.3`

## Why

The SSH IPC handler ultimately reaches a shell execution path, so prefix-only validation was insufficient against metacharacter injection attempts.

The dependency graph also still resolved vulnerable `minimatch` versions through transitive dependencies, so a direct bump alone was not enough.

## Testing

Ran:
- `corepack pnpm run format:check`
- `corepack pnpm run lint`
- `corepack pnpm run type-check`
- `corepack pnpm exec vitest run`
- `corepack pnpm run build:main`
- `corepack pnpm run build:renderer`

Result:
- 50 test files passed
- 526 tests passed
- 1 test skipped

## Notes

This keeps the changes minimal and focused:
- SSH fix is localized to the public renderer-exposed command validation path
- dependency remediation uses patch-level overrides to avoid unnecessary dependency churn

### DIFF
--- a/src/main/utils/sshCommandValidation.ts
+++ b/src/main/utils/sshCommandValidation.ts
@@ -17,19 +17,22 @@ const ALLOWED_COMMAND_PREFIXES = [
 const SHELL_CONTROL_CHARACTER_PATTERN = /[\r\n;&|<>`$]/;
 
 export function getSshExecuteCommandValidationError(command: string): string | null {
+  if (SHELL_CONTROL_CHARACTER_PATTERN.test(command)) {
+    return 'Command contains invalid shell control characters';
+  }
+
   const trimmed = command.trimStart();
 
   if (!trimmed) {
     return 'Command not allowed';
   }
 
-  if (SHELL_CONTROL_CHARACTER_PATTERN.test(trimmed)) {
-    return 'Command contains invalid shell control characters';
-  }
-
-  const isAllowed = ALLOWED_COMMAND_PREFIXES.some(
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Thanks for the contribution! I pushed a small fixup commit that addresses the review feedback: moved the control-char check before trimStart() so leading \r/\n are properly rejected, and tightened the prefix matching so e.g. pwdx doesn't pass the pwd allowlist entry. Added test cases for both. LGTM otherwise — will merge once CI is green.

PR #1431, opened 2026-03-12
fix(ssh): resolve saved connection by ID when testing with ID-only config

dhunganapramod9:
## Description

When the renderer calls `sshTestConnection` with only a connection ID and empty host/username (e.g. from `SshConnectionTestButton`), the main process previously used those empty values and the test could not succeed. This change detects that "ID-only" call shape, loads the saved connection from the DB, hydrates credentials from keytar, and runs the test against the resolved config.

## Rationale

`SshConnectionTestButton` only has access to `connectionId` and was intentionally calling the IPC with ID + empty fields and a TODO for the main process to handle it. The main process did not handle that case; this fix adds the same resolve-by-ID pattern already used by the `CONNECT` handler.

## Changes

- **Main process (`src/main/ipc/sshIpc.ts`)**: In the `TEST_CONNECTION` handler, detect ID-only calls (config has `id` but `host` or `username` empty). For those, load the saved connection from the DB, hydrate password/passphrase from keytar, and run the test using the resolved config.
- **Renderer (`src/renderer/components/ssh/SshConnectionTestButton.tsx`)**: Comment updated to state that the main process accepts ID-only and resolves from DB/keytar.

## Testing

- No new tests; behavior aligns with the existing `CONNECT` handler (resolve by ID from DB + keytar).
- Manual: use "Test" on a saved SSH connection from the UI that uses `SshConnectionTestButton`.

## Pre-PR Checklist

- [ ] Dev server runs: `pnpm run d` (or `pnpm run dev`) starts cleanly.
- [ ] Code is formatted: `pnpm run format`.
- [ ] Lint passes: `pnpm run lint`.
- [ ] Types check: `pnpm run type-check`.
- [ ] Tests pass: `pnpm exec vitest run`.
- [ ] No stray build artifacts or secrets committed.
- [ ] Documented any schema or config changes impacting users. (N/A — no schema/config changes.)

### DIFF
--- a/src/main/ipc/sshIpc.ts
+++ b/src/main/ipc/sshIpc.ts
@@ -161,6 +161,49 @@ export function registerSshIpc() {
       config: SshConfig & { password?: string; passphrase?: string }
     ): Promise<ConnectionTestResult> => {
       try {
+        // When renderer sends only connection id (e.g. SshConnectionTestButton), resolve from DB and keytar
+        const isIdOnly =
+          config.id &&
+          (typeof config.host !== 'string' || !config.host.trim()) &&
+          (typeof config.username !== 'string' || !config.username.trim());
+
+        let resolvedConfig: SshConfig & { password?: string; passphrase?: string } = config;
+        if (isIdOnly && config.id) {
+          const { db } = await getDrizzleClient();
+          const rows = await db
+            .select({
+              id: sshConnectionsTable.id,
+              name: sshConnectionsTable.name,
+              host: sshConnectionsTable.host,
+              port: sshConnectionsTable.port,
+              username: sshConnectionsTable.username,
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Hi @dhunganapramod9 

Would love to see this fix on our v1's codebase we are launching today or tomorrow. 

A quick heads-up: later this week we’ll be switching this repo from the current (“old”) Emdash codebase to the new v1 (beta launched last week; ~25% of our users have already moved over). Some PRs against the old code may become obsolete or need a quick rebase/adaptation after the switch.

If you get a chance to try v1, We'd love your thoughts — reply here or share feedback on Discord:
https://discord.com/invite/f2fv7YxuR2

PR #1436, opened 2026-03-12
fix(pty): prevent session ID collision for non-worktree tasks

Simonstorms:
## Summary

Fixes "Session ID ... is already in use" error when creating a new task without a worktree.

New non-worktree tasks were entering the multi-chat session discovery branch, which scans `~/.claude/projects/` for existing `.jsonl` session files. If any of those files belonged to an active Claude process, the discovered UUID was passed via `--session-id` and Claude rejected it as already in use.

The fix gates that discovery branch on `isResume` so new tasks always get a fresh session UUID instead of trying to adopt one from disk.

## Screenshot

Screenshot 2026-03-12 at 11 01 23

## Type of change
- [x] Bug fix

## Checklist
- [x] `pnpm run format`
- [x] `pnpm run lint`
- [x] `pnpm run type-check`
- [x] `pnpm exec vitest run` (638 tests passing)

### DIFF
--- a/src/main/services/ptyManager.ts
+++ b/src/main/services/ptyManager.ts
@@ -525,7 +525,7 @@ export function applySessionIsolation(
     return true;
   }
 
-  if (hasOtherSameProviderSessions(id, parsed.providerId, cwd)) {
+  if (isResume && hasOtherSameProviderSessions(id, parsed.providerId, cwd)) {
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Good catch, @Simonstorms. And thanks for opening the PR.
Merging.

PR #1449, opened 2026-03-12
feature: expandable mini terminal with bug fixes 

### LINKED ISSUE #1378: Feature: Expandable mini terminal — modal view for full-screen terminal access
## Feature Request

**Description**
The mini terminal is great for quick commands without taking up much space. But for more involved terminal work (e.g., interactive git rebase, reviewing logs), it's too small. Add a button to temporarily expand the mini terminal into a full-size modal, then collapse it back when done.

## Proposed UX
1. Mini terminal stays compact by default — perfect for quick commands (`git push`, `git commit --amend --no-edit`, etc.)
2. A small expand button (e.g., maximize icon) in the top-right corner of the mini terminal
3. Clicking it opens the terminal in a **full-size modal overlay** — same session, just bigger
4. User does their work (interactive rebase, scrolling through logs, etc.)
5. Closing the modal returns to the compact mini terminal — session preserved, no interruption

## Why This Matters
The mini terminal hits a sweet spot for quick commands, but interactive operations like `git rebase -i main` need more vertical space to see the full commit list and editor. Rather than permanently resizing the pane, a temporary expand-to-modal keeps the default layout clean while giving full access when needed.

## Keyboard Shortcut (Nice to Have)
A shortcut to toggle the expanded view (e.g., `Cmd+Shift+T` or double-clicking the terminal header) would make this even faster.

radhika-kapoor:
## Summary

Adds an expandable mini terminal modal to task and project terminals,
then fixes three bugs discovered during implementation.

Closes #1378 

## Feature: Expandable Mini Terminal
- Added an expand icon button to task and project `TaskTerminalPanel`
  components that opens the terminal in a full-screen modal overlay
- Keyboard shortcut: **Cmd+Shift+T** (macOS) / **Ctrl+Shift+T** (other
  platforms) toggles the expanded terminal from anywhere in the UI
- Extracted shortcut detection into `src/renderer/lib/terminalShortcuts.ts`
  for reusability and testability

## Bug Fixes

1. Empty terminal in expanded modal
- Root cause: Radix UI Portal uses a two-pass mount — the portal DOM node
  does not exist during the first React commit, so `useRef + useEffect`
  fired before the container was in the DOM
- Fix: Replaced with a callback ref (`ref={handleExpandedContainerRef}`)
  which fires only when the element is actually inserted, ensuring xterm
  attaches at the right time

2. Garbled / scrambled repeated prompt lines in task & project terminals
- Root cause: `SerializeAddon.serialize()` was capturing the corrupted
  terminal buffer into snapshots, which were then restored on every
  subsequent session attach — perpetuating the garble
- Fix: Added `disableSnapshots: true` on shell-based terminals
  (`TaskTerminalPanel`, global terminal `TerminalPane`) so stale buffer
  state is never persisted or restored

3. Multiple SIGWINCH / extra prompt lines on terminal attach
- Root cause: Each distinct PTY resize event during layout settling
  (panel open, modal animation, flex reflow) triggers a SIGWINCH, causing
  the shell to redraw the PS1 prompt — producing extra lines
- Fix: Added a 400ms post-attach stabilization window in
  `TerminalSessionManager`. All resize events within this window are
  collapsed into a single debounced PTY notification, sending only one
  SIGWINCH after layout settles

## Tests Added
- Extracted `computeResizeDelay` into `src/renderer/terminal/resizeUtils.ts`
  and added 6 tests covering the
  stabilization window logic
- Expanded `isTerminalExpandShortcut` tests:
  lowercase/uppercase T, both-modifiers set, missing Shift, non-T keys,
  Enter key, and altKey passthrough
  
 ## Mandatory Sanity Tests:
 - `pnpm run type-check`
 - `pnpm run lint`
 - `pnpm exec vitest run src/test/renderer/terminalShortcuts.test.ts`
 - `pnpm exec vitest run src/test/renderer/terminalSessionStabilize.test.ts`


arnestrickmann:
Thank you for taking this on - however it already has been implemented. 

Thank you, again.

PR #1454, opened 2026-03-13
fix(github): ensure local branch exists for fork PR checkout

jschwxrz:
summary:
- fix PR review worktree setup for forked PRs when direct fallback is needed
- remove --detach from fallback gh pr checkout
- after checkout explicitly verify refs/heads/<branch> exists
- If missing create the local branch ref from HEAD

this ensures downstream worktree creation can reliably use the expected local branch for fork PR review flows.


arnestrickmann:
Thanks! good catch

PR #1502, opened 2026-03-16
feat: add emdash account system with github oauth sign-in

jschwxrz:
summary:
- adds first party emdash accounts backed by an external auth server
- sign-in opens browser, runs a PKCE oauth flow via loopback server, exchanges code for session + github token
- if user has an active emdash account, github auth skips device flow and uses oauth instead

changes:
- new EmdashAccountService / OAuthFlowService / AccountCredentialStore / AccountProfileCache in main process, handles the full sign-in lifecycle, session token in keychain via keytar, profile cache on disk
- new EmdashAccountProvider context in renderer — loads session on mount, validates against server, exposes signIn/signOut/refreshSession
- GithubContextProvider and IntegrationsCard now try OAuth auth first when user has an active account, falling back to Device Flow
- new Account tab in settings page with signed-in/expired/no-account states
- changelog notification card shows "create account" CTA when no account exists, links to account settings
- IPC: account:getSession, account:signIn, account:signOut, account:checkServerHealth, account:validateSession, github:auth:oauth

### DIFF
--- a/src/main/services/OAuthFlowService.ts
+++ b/src/main/services/OAuthFlowService.ts
@@ -151,6 +151,7 @@ export class OAuthFlowService {
       method: 'POST',
       headers: { 'Content-Type': 'application/json' },
       body: JSON.stringify({ state, code, code_verifier: codeVerifier }),
+      signal: AbortSignal.timeout(30_000),
     });
 
     if (!response.ok) {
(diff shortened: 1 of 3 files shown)


jschwxrz:
@greptileai

PR #1504, opened 2026-03-16
feat(terminal): add shell selector with auto-detection

samuelpatro:
## Summary
- Adds a shell selector dropdown to Terminal Settings, allowing users to choose their preferred shell (bash, zsh, PowerShell, cmd, WSL distros, etc.)
- Auto-detects available shells on the system (reads `/etc/shells` on Unix, probes known paths + WSL on Windows)
- Persists the selected shell in app settings (`terminal.shellPath`)
- Wires the setting through to PTY spawning via `ptyManager.ts`

## Changed Files
- `src/main/ipc/appIpc.ts` — New `app:detectShells` IPC handler
- `src/main/preload.ts` — Expose `detectShells` to renderer
- `src/main/services/ptyManager.ts` — Use configured shell path when spawning PTYs
- `src/main/settings.ts` — Add `terminal.shellPath` to settings schema
- `src/renderer/components/TerminalSettingsCard.tsx` — Shell selector UI component
- `src/renderer/types/electron-api.d.ts` — Type declaration for `detectShells`

## Test plan
- [ ] Verify shell detection works on macOS/Linux (reads `/etc/shells`)
- [ ] Verify shell detection works on Windows (cmd, PowerShell, Git Bash, WSL)
- [ ] Select a non-default shell and confirm new PTY sessions use it
- [ ] Confirm the setting persists across app restarts
- [ ] Confirm "Default" option uses the system default shell

### DIFF
--- a/src/main/ipc/appIpc.ts
+++ b/src/main/ipc/appIpc.ts
@@ -1,5 +1,6 @@
 import { app, clipboard, ipcMain, shell } from 'electron';
 import { exec, execFile } from 'child_process';
+import { existsSync } from 'fs';
 import { readFile } from 'fs/promises';
 import { join } from 'path';
 import { ensureProjectPrepared } from '../services/ProjectPrep';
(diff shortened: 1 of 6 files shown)


arnestrickmann:
Hi @Samuell1
Thanks for the PR - could you run the formatting script?

samuelpatro:
@arnestrickmann i tried to run it and it found nothing just different line endings, not sure if its something wrong with my setup but i use bun to run cmds

rdadolf:
Hey @arnestrickmann and @Samuell1. I was interested in trying out emdash, but lack of WSL support was a blocker. I saw Samuel's attempt at getting this rolling a couple weeks back, but it seemed like maybe it wasn't pressing (understandable). I figured I'd just fix the rest of it and help finish it off. The [linked commit above](https://github.com/rdadolf/emdash/commit/50670795b24a13797e155b1b513c6318c23a0f43) from my fork addresses most of what's required here.

I figured I'd leave the decision on how you wanted to handle this up to you two. I could PR it to Samuel's fork and let his update here or I could just create a new PR on the main repo and shift things there. I've run the pre-commit checks locally, but we'd need PR checks run on GH again, obviously.

What's your preference?

rdadolf:
I should also note that full WSL support for emdash would be a bit more involved. Currently, all the tooling is done natively on the host, which can be a real problem for WSL environments. For instance, all your agents, git configurations, and the like are all going to be completely separate. The terminal is a great start, but the integrated tools would need to get shifted, too.

There's probably a way to fix this by wrapping the native tool calls in emdash with optional indirection. I poked around at it for a bit and had claude write up a summary of our findings. It's definitely not a 2-line fix, but it's probably still doable. The actual changes to direct WSL tool execution probably isn't the hard part. It'll probably be unexpected assumptions about where things live and how they behave that end up causing bugs, but that's just my guess.

In any case, content below. It's not necessary for this PR, but it's good to know what we might have to deal with down the road.

----

### Scope

The codebase executes external commands via `exec`, `execFile`, `execSync`, and `spawn` across **270 call sites in 29 files**:

| Category | Examples | Approx. call sites |
|----------|---------|---------------------|
| Git | `git fetch`, `git worktree`, `git diff`, `git status` | ~150 |
| Provider CLIs | `claude`, `codex`, `gemini`, etc. (via `resolveCommandPath`) | ~20 |
| SSH | `ssh`, `scp` | ~10 |
| Package managers | `brew`, `gh`, `winget` | ~5 |
| System | `ps`, `cmd`, `chmod`, `attrib` | ~5 |
| Shell/env probing | shell detection, env initialization | ~10 |

There is no centralized execution layer — commands are dispatched through per-file `execAsync = promisify(exec)` wrappers in at least 8 separate files (`gitIpc.ts`, `githubIpc.ts`, `GitService.ts`, `GitHubService.ts`, `WorktreeService.ts`, `RepositoryManager.ts`, `ForgejoService.ts`, `GitLabService.ts`).

### What a fix would require

1. **Command routing**: When the project `cwd` is a WSL path, prefix commands with `wsl.exe -d <distro> --`. E.g., `git fetch --all --prune` becomes `wsl.exe -d Ubuntu-24.04 -- git fetch --all --prune`.

2. **Path translation**: Every `cwd` passed to a routed command needs `\\wsl$\Ubuntu-24.04\home\user\project` → `/home/user/project`. This is mechanical (strip UNC prefix, flip separators).

3. **Centralized execution layer**: Something like:
   ```typescript
   function execForProject(command: string, args: string[], cwd: string): Promise<ExecResult> {
     const distro = extractWslDistro(cwd);
     if (distro) {
       const linuxCwd = wslPathToLinux(cwd);
       return execFileAsync('wsl.exe', ['-d', distro, '--', command, ...args], { cwd: linuxCwd });
     }
     return execFileAsync(command, args, { cwd });
   }
   ```

### Hard parts

- **Worktree paths**: Emdash creates worktrees in `../worktrees/`. On Windows that's a Windows path, but WSL git needs the Linux equivalent.
- **Mixed contexts**: Some operations touch both Windows state (SQLite DB in `%APPDATA%`) and WSL state (git repos on the Linux filesystem).
- **The `execAsync` string-command pattern**: About half the call sites use `exec('git fetch ...')` with command strings rather than `execFile('git', ['fetch', ...])` with arg arrays. The string pattern is harder to intercept cleanly.

### Impact without this

WSL projects in emdash currently work for terminal sessions (with the shell selector) but not for:
- Worktree creation/management (dubious ownership errors)
- Git operations from the UI (status, diff, branch, push)
- Provider CLI detection (finds Windows installs, misses WSL installs and their config)
- Task lifecycle scripts

PR #1538, opened 2026-03-19
fix: wait for setup script before spawning agent PTY

### LINKED ISSUE #1536: [bug]: skills copied over in setup script not available in first Claude Code session
### Bug Description

In our setup script, we copy over not version controlled project level skills from the root project's `.claude/skills` folder to the worktree.

But it looks like Claude Code is initialized before the setup script has finished, and thus the skills are not picked up.

We need to make sure that agents are only started _after_ the setup script has finished.

The current workaround is to open a second agent session upon which the skills will be picked up.

### Steps to Reproduce

1. In the setup script, copy agent skills into the worktree directory.
2. Create a new worktree/task.
3. Check the available skills in the agent - the copied skills won't be available (at least not for Claude Code).
4. Open a second agent session and check the available skills - the copied skills will be available _(this is our current workaround)_.

The copied over skills are not available, but they should be if the setup script had time to finish before the agent session was launched.

### emdash Version

v0.4.37

### Operating System

macOS 26.3.1

### Additional Context

_No response_

ubuntudroid:
## Summary

- Fixes skills (and other files) copied by a setup script not being available in the first agent session
- Root cause: agent PTYs were spawned via `pty:startDirect` before the in-flight setup script had finished — `pty:startDirect` is the path used by all provider agents (Claude, Codex, etc.)
- Adds `awaitSetup(taskId)` to `TaskLifecycleService` that awaits any in-flight setup promise for the given task, silently ignoring failures so a broken setup script never blocks the agent
- Both `pty:startDirect` and `pty:start` now call `awaitSetup` before spawning; if no setup is running the call resolves immediately with no overhead

Closes #1536

## Test plan

- [x] Create a new worktree task with a setup script that copies `.claude/skills` (or similar files) into the worktree
- [x] Open the first agent session — skills should be available immediately without needing to open a second session
- [x] Two new unit tests in `ptyIpc.test.ts` verify that `startDirectPty` / `startPty` are not called while `awaitSetup` is pending, and are called once it resolves

### DIFF
--- a/src/test/main/ptyIpc.test.ts
+++ b/src/test/main/ptyIpc.test.ts
@@ -19,6 +19,7 @@ const ipcOnHandlers = new Map<string, (...args: any[]) => any>();
 const appListeners = new Map<string, Array<() => void>>();
 const ptys = new Map<string, MockProc>();
 const notificationCtor = vi.fn();
+const awaitSetupMock = vi.fn(async (_taskId: string) => {});
 const notificationShow = vi.fn();
 const telemetryCaptureMock = vi.fn();
 const agentEventGetPortMock = vi.fn(() => 12345);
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Tested locally, this LGTM. Merging.

Thanks!

Issue #1531, opened 2026-03-19
[bug]: fast-forward buffering in terminal and agent


tcrasset:
### Bug Description

Let's say I have an agent running and thinking, or a long running command in the terminal that produces output (e.g. a docker compose up where there is a clock ticking, see screenshot).
If I leave the terminal/agent, either by switching to a different task, or by focusing/alt-tabbing to another program e.g Firefox, when coming back to the terminal/agent/emdash, it looks like at the things that happened between the moment I switched and now, things are getting 'fast-forwarded', e.g time moves much faster, but it does not show the final output right away.

I have a hard time describing the bug that I see, but it looks like some sort of 'fast-forward'. You would expect to come back to the agent/terminal and see the current output, but what you see is the previous output, and it fast-forwards until the current output is shown. This may take a few minutes depending on the 'fast-forward' speed as well as the time spent somewhere else.

### Steps to Reproduce

Have lots of tasks open tasks with lots of changes, I think.
Run a task, switch to another window, wait a few minutes, go back to the agent, and see the "thinking" icon of the agent tick MUCH faster than usual, or in the terminal, the time elapsed of a command tick much faster than 1s (see screenshot)

### Actual vs Expected Behavior

I expect to be shown the current output immediately.

### emdash Version

0.4.37

### Operating System

Ubuntu 25.10

### Additional Context

The right time indiciation was increasing much faster than once per second

<img width="1812" height="386" alt="Image" src="https://github.com/user-attachments/assets/b5cf43d1-e7b4-454c-8d10-49bfd598bcff" />

jordanmiguel:
+1 @tcrasset 

I can confirm this. Also relates to how the software starts getting slow on long running tasks. (https://github.com/generalaction/emdash/issues/1524)

nehaaprasad:
@tcrasset please check it out the PR, thankyou

arnestrickmann:
Thanks for raising this, @tcrasset, and thanks for the PR, @naaa760.

I tested `#1544` locally and I can still reproduce the issue there. When I switch away and come back, I still see the terminal/agent visibly replay buffered output instead of showing the current state immediately.

The `#1547` approach does seem to fix it for me. So I suspect the problem is in the drain scheduling around visibility/refocus, not just catch-up speed.

PR #1547, opened 2026-03-20
fix(terminal): use buffered write drain scheduler with visibility fallback

### LINKED ISSUE #1510: Bug: Terminal replays buffered output when switching back to Emdash
## Bug Report

**Description**
When switching away from Emdash (e.g., working on another desktop/screen on macOS) and then switching back, the agent terminal "replays" all the output that accumulated while the app was in the background. This replay can take 1-2 minutes of watching the terminal scroll through 10-15 minutes of buffered output before catching up to the current state.

## Steps to Reproduce
1. Start an agent task in Emdash
2. Switch to another application / desktop / screen on macOS
3. Let the agent work for 10-15 minutes in the background
4. Switch back to Emdash
5. Observe the terminal replaying all the accumulated output instead of jumping to the current state

## Expected Behavior
When switching back to Emdash, the terminal should immediately show the current state of the agent output — not replay everything that happened while the app was in the background. The user should be able to scroll up to see history if needed.

## Actual Behavior
The terminal slowly replays all buffered output sequentially, forcing the user to wait 1-2 minutes before they can see or interact with the current agent state.

## Likely Cause
When the app is not visible, xterm.js may be pausing rendering. When the app regains focus, the buffered PTY data is fed to xterm.js all at once, but it renders sequentially (line by line / chunk by chunk) instead of jumping to the end. This may be a recent regression.

## Suggested Fix
- Flush all buffered PTY data to xterm.js instantly when the window regains focus (batch write instead of streaming)
- Or keep rendering in the background even when the window is not focused
- Or skip intermediate rendering and only display the final state of the buffer

arnestrickmann:
## Summary

Replaces the raw `requestAnimationFrame` call in `TerminalSessionManager.scheduleWriteDrain()` with a new `scheduleTerminalWriteDrain` utility that handles background-tab scenarios.

## Problem

When the app tab is hidden, `requestAnimationFrame` callbacks are throttled or paused by the browser, causing terminal write drains to stall indefinitely.

## Changes

- Add `writeDrainScheduler.ts` — schedules drains via `requestAnimationFrame` when visible, with a `setTimeout` fallback (48ms) to guarantee delivery even when throttled. Falls back to `setTimeout(0)` when the document is hidden.
- Update `TerminalSessionManager` to use the new scheduler and track a `cancelPendingWriteDrain` handle, ensuring pending drains are properly cancelled on dispose/reset.

Fixes #1510

## Summary by CodeRabbit

* **Refactor**
  * Improved terminal write operation scheduling to enhance responsiveness when the window is visible, with optimized timing strategies and refined fallback behavior for edge cases.

* **Tests**
  * Added comprehensive test suite for terminal write scheduling with visibility state detection and timer fallback scenarios.

### DIFF
--- a/src/renderer/terminal/TerminalSessionManager.ts
+++ b/src/renderer/terminal/TerminalSessionManager.ts
@@ -29,6 +29,7 @@ import {
   type TerminalSearchBufferLike,
   type TerminalSearchMatch,
 } from './terminalSearch';
+import { scheduleTerminalWriteDrain } from './writeDrainScheduler';
 import { rpc } from '@/lib/rpc';
 import { APP_SHORTCUTS, normalizeShortcutKey } from '@/hooks/useKeyboardShortcuts';
 
@@ -145,6 +146,7 @@ export class TerminalSessionManager {
   private readonly pendingWriteQueue: string[] = [];
   private queuedWriteChars = 0;
   private writeDrainScheduled = false;
+  private cancelPendingWriteDrain: CleanupFn | null = null;
   private writeInFlight = false;
   private shouldScrollToBottomAfterWrites = false;
   private lastSlowInputLogAt = 0;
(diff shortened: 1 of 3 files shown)


arnestrickmann:
fixes https://github.com/generalaction/emdash/issues/1531

PR #1565, opened 2026-03-23
fix: suppress EPIPE during PTY shutdown on Windows

### LINKED ISSUE #1192: [bug]: Uncaught read EPIPE exception when archiving a task on Windows
### Bug Description

Archiving a task that has (or had) a running PTY session sometimes triggers an uncaught exception in the Electron main process, showing an error dialog. The error is non-fatal (clicking OK dismisses it), but it occurs intermittently on Windows. The intermittent nature is consistent with a race condition — the error only surfaces when the ConPTY pipe read happens to fire after the process tree is killed but before the socket is cleaned up.

Image

### Steps to Reproduce

1. Open Emdash on Windows
2. Create a task that spawns an agent (e.g. Claude Code)
3. Let the agent start (PTY is active)
4. Archive the task (the agent can be running or idle at the prompt)
5. The error dialog sometimes appears (not every time — it's a race condition)

**Actual:** An error dialog appears:

```
A JavaScript error occurred in the main process

Uncaught Exception:
Error: read EPIPE
    at Pipe.onStreamRead (node:internal/stream_base_commons:218:20)
```

**Expected:** The task is archived cleanly with no error dialog. Pipe errors during PTY shutdown should be suppressed, just as `EIO` errors already are.

### emdash Version

0.4.21

### Operating System

Windows 11 Pro 10.0.26200

### Archive flow (from renderer source)

The renderer's archive handler explicitly kills every PTY associated with the task before writing the archive to the DB:

1. Calls `window.electronAPI.ptyKill(mainPtyId)` for the main provider PTY
2. Iterates non-main conversations and calls `ptyKill` for each chat PTY
3. Calls `Xc()` (terminal state cleanup) which kills any remaining shell PTYs
4. Concurrently runs `lifecycleTeardown` and `lifecycleRunStop`
5. Then writes `archiveTask` to the DB and calls `lifecycleClearTask`

There is no `process.on('uncaughtException')` handler anywhere in the main process, so Electron's default behavior (showing the error dialog) applies.

### Root cause analysis (most likely)

The most likely cause is a race condition in `node-pty`'s ConPTY kill path on Windows:

1. `killPty(id)` in `ptyManager.js:970` calls `rec.proc.kill()` on the `WindowsTerminal` instance.
2. On Windows with ConPTY, `WindowsPtyAgent.kill()` (`windowsPtyAgent.js:156`) is **asynchronous**: it forks a child process (`conpty_console_list_agent.js`) to enumerate the console process list, kills each process, then calls `_ptyNative.kill()`. It sets `_outSocket.readable = false` but does **not** destroy the socket immediately.
3. During this async window, the ConPTY output pipe breaks (the process tree is being killed), and the `net.Socket` (`_outSocket`) receives an `EPIPE` error on `Pipe.onStreamRead`. On Windows, `ERROR_BROKEN_PIPE` during a pipe read is mapped to `UV_EPIPE` — unlike Unix where `EIO` is the typical error when a PTY child exits.
4. The socket error handler in `node-pty`'s `windowsTerminal.js` suppresses `EIO` errors (the Unix case) but does **not** suppress `EPIPE` (the Windows/ConPTY case):

   ```js
   // node-pty windowsTerminal.js (socket error handler)
   _this._socket.on('error', function (err) {
       _this._close();
       if (err.code) {
           if (~err.code.indexOf('errno 5') || ~err.code.indexOf('EIO'))
               return;  // Only EIO is suppressed, not EPIPE
       }
       if (_this.listeners('error').length < 2) {
           throw err;  // <-- EPIPE thrown as uncaught exception
       }
   });
   ```

5. `_this.listeners('error')` delegates to `_this._socket.listeners('error')` (via `terminal.js:157`). Since emdash does not register any additional error listeners on the PTY proc, the count is 1 (only the built-in handler), so the `< 2` check passes and the error is **thrown** from inside the event callback — becoming an uncaught exception.

### Disclaimer

The root cause analysis was produced by Claude Code (claude-opus-4-6) by extracting and reading the emdash `app.asar` source and `node-pty` native module code. The bug itself is real and reproducible, but the analysis may be incorrect — it is based entirely on source code reading with no empirical log data (emdash does not persist application logs to disk).

The stack trace is minimal (`Pipe.onStreamRead` at `node:internal/stream_base_commons:218:20`), so there's no filename pointing directly to `node-pty`. The attribution to `node-pty` is based on:

- `node-pty` is the only code in emdash that has a deliberate `throw err` inside a socket error handler for non-EIO pipe errors
- The error coincides exactly with PTY kill during archive
- No other pipe/socket in emdash lacks error handling in a way that would produce this during archive (TaskLifecycleService child processes only read from stdout/stderr — EPIPE is a broken-pipe condition, and their ChildProcess `error` event is handled; AgentEventService has a server-level error handler)
- EPIPE on a pipe **read** is a Windows-specific behavior (`ERROR_BROKEN_PIPE` → `UV_EPIPE`), consistent with ConPTY pipe teardown

### Suggested fix (emdash side)

Register an error listener on PTY procs in `ptyIpc.js` when `onData`/`onExit` listeners are set up (in both the `pty:start` and `pty:startDirect` handlers). Since `node-pty`'s `Terminal.on()` proxies to `_socket.on()`, this bumps the socket error listener count to 2, causing `node-pty` to swallow the error instead of throwing:

```js
// After proc.onData / proc.onExit registration:
proc.on('error', () => {
    // Suppress pipe errors (EPIPE/EIO) during PTY shutdown.
    // node-pty's built-in handler only suppresses EIO; adding this
    // listener brings the count to >= 2, skipping the throw.
});
```

### Suggested fix (node-pty upstream)

The socket error handler in `windowsTerminal.js` should also suppress `EPIPE`, which is the Windows equivalent of `EIO` when a ConPTY pipe is broken during process shutdown:

```js
if (~err.code.indexOf('errno 5') || ~err.code.indexOf('EIO') || ~err.code.indexOf('EPIPE'))
    return;
```

saschabuehrle:
Fixes #1192

On Windows with ConPTY, killing a PTY process tree breaks the output pipe with EPIPE. node-pty only suppresses EIO (the Unix equivalent), so the EPIPE bubbles up as an uncaught exception and shows an error dialog.

The fix registers an error listener on every spawned PTY proc. For EPIPE/EIO it silently returns, for anything else it logs a warning. This also bumps the node-pty socket listener count past the throw threshold.

Greetings, saschabuehrle

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of terminal/PTY error handling on Windows to reduce unexpected error surfacing during terminal sessions and lifecycle operations.

* **Tests**
  * Added a unit test verifying platform-specific attachment of PTY error handling to ensure behavior differs between Windows and non-Windows platforms.

### DIFF
--- a/src/main/services/ptyManager.ts
+++ b/src/main/services/ptyManager.ts
@@ -19,6 +19,8 @@ import { errorTracking } from '../errorTracking';
  * swallow the error instead of throwing.
  */
 function suppressPtyPipeErrors(proc: IPty): void {
+  if (process.platform !== 'win32') return;
+
   // IPty doesn't expose .on() in its type, but the underlying Terminal
   // class extends EventEmitter and proxies to the socket.
   (proc as any).on?.('error', (err: NodeJS.ErrnoException) => {
--- a/src/test/main/ptyManager.test.ts
+++ b/src/test/main/ptyManager.test.ts
@@ -383,6 +383,63 @@ describe('ptyManager provider command resolution', () => {
       })
     );
   });
+
+  it('attaches PTY pipe error suppression on Windows only', async () => {
+    const originalPlatformDescriptor = Object.getOwnPropertyDescriptor(process, 'platform');
+    const onWin32 = vi.fn();
+    const onDarwin = vi.fn();
+
+    try {
+      Object.defineProperty(process, 'platform', {
+        value: 'win32',
+        configurable: true,
+      });
+
+      nodePtySpawnMock.mockReturnValueOnce({
+        on: onWin32,
+        write: vi.fn(),
+        resize: vi.fn(),
+        kill: vi.fn(),
+        onData: vi.fn(),
+        onExit: vi.fn(),
+      });
+
+      const { startPty } = await import('../../main/services/ptyManager');
+      await startPty({
+        id: 'codex-main-task-win32',
+        cwd: '/tmp/task',
+        shell: '/bin/zsh',
+      });
+
+      expect(onWin32).toHaveBeenCalledWith('error', expect.any(Function));
+
+      Object.defineProperty(process, 'platform', {
+        value: 'darwin',
+        configurable: true,
+      });
+
+      nodePtySpawnMock.mockReturnValueOnce({
+        on: onDarwin,
+        write: vi.fn(),
+        resize: vi.fn(),
+        kill: vi.fn(),
+        onData: vi.fn(),
+        onExit: vi.fn(),
+      });
+
+      await startPty({
+        id: 'codex-main-task-darwin',
+        cwd: '/tmp/task',
+        shell: '/bin/zsh',
+      });
+
+      expect(onDarwin).not.toHaveBeenCalledWith('error', expect.any(Function));
+    } finally {
+      if (originalPlatformDescriptor) {
+        Object.defineProperty(process, 'platform', originalPlatformDescriptor);
+      }
+    }
+  });
 });
 
 describe('stale Claude session detection', () => {


arnestrickmann:
Thanks for the PR. The Windows ConPTY fix looks right, but I think the listener should be Windows-scoped only. As written originally, it changes `node-pty` error handling on macOS/Linux too. I narrowed it to `win32` and added a small regression test. With that, happy to merge:)

Issue #1518, opened 2026-03-18
zellij as first-class session backend: built-in resurrection, no plugins, cmd+q becomes safe by default


yigitkonur:
## the opportunity

emdash's tmux wrapping ([GEN-464](https://linear.app/general-action/issue/GEN-464/feature-optionally-back-pty-sessions-with-tmux-for-persistentresumable), shipped in v0.4.25) was a great move — it made session persistence possible. but it's opt-in, requires tmux to be installed, and relies on tmux-resurrect + tmux-continuum for full resurrection after reboots. most users never toggle it on, which means most users still lose everything on cmd+q.

zellij has **built-in session resurrection with zero plugins**. if emdash adopted zellij as its session backend, process persistence would work out of the box — no opt-in toggle, no plugin dependencies, no configuration. cmd+q becomes a safe operation by default. no other agentic development environment does this today.

## why zellij over tmux

| capability | tmux | zellij |
| -- | -- | -- |
| session resurrection | requires tmux-resurrect + tmux-continuum plugins | **built-in, zero plugins** |
| kitty keyboard protocol | not supported (open pr [#4068](<https://github.com/generalaction/emdash/issues/4068>)) | **supported** — enables shift+enter in claude code, codex |
| scrollback serialization | plugin-based, every 15min | built-in, configurable interval |
| layout persistence format | binary/opaque via plugin | **human-readable kdl files** |
| auto-attach | requires manual shell config | built-in `zellij attach --create` |
| force close behavior | kills session | configurable: `on_force_close "detach"` |
| command restart on resurrection | plugin handles it | built-in `--force-run-commands` flag |

the kitty keyboard protocol support is especially relevant — emdash users running claude code or codex through tmux-wrapped sessions currently can't use shift+enter and other extended key sequences. zellij supports these natively.

## how it maps to emdash's architecture

emdash's current tmux wrapping creates sessions named `emdash-{taskId}-{provider}`. the same pattern works with zellij, but better:

### local sessions

```
# current (tmux)
tmux new-session -d -s emdash-task123-claude 'claude'
tmux attach -t emdash-task123-claude

# proposed (zellij)
zellij attach --create emdash-task123-claude
```

`zellij attach --create` is idempotent — creates if new, reattaches if exists. no separate "detect existing session" logic needed in `RemotePtyService`.

### remote sessions (ssh)

[...]

PR #1571, opened 2026-03-24
Persistent session backends

### LINKED ISSUE #1215: Feature: Optionally back PTY sessions with tmux for persistent/resumable sessions
## Feature Request

**Description**
Allow Emdash to optionally wrap agent PTY sessions in named tmux sessions. This would make agent sessions persistent and resumable — users could SSH into the host machine from another device and `tmux attach` to pick up where they left off, while Emdash stays connected as a tmux client.

## Problem
Today, Emdash manages PTY sessions directly. Each connection starts its own independent session with no persistence:
- If you close Emdash or lose connection, the session is gone
- You can't pick up an existing agent session from another device via SSH
- There's no way to inject tmux via project config or lifecycle scripts
- Users are working around this by manually Ctrl+C-ing, creating tmux sessions, and re-running agents

### Settings toggle
A simple toggle in project settings or global settings:
> **"Wrap agent sessions in tmux"** — default: off

### Behavior when enabled
- Each agent session is started inside a named tmux session (e.g., `emdash-{taskId}-{provider}`)
- Emdash connects as a tmux client rather than owning the PTY directly
- Sessions persist independently of Emdash's connection state
- Users can `tmux attach -t emdash-{taskId}-{provider}` from any SSH session on the same machine
- Reconnecting Emdash to the project re-attaches to existing tmux sessions

### Benefits
- **Multi-device workflow**: Start a task on your desktop, check on it from your laptop via SSH
- **Session persistence**: Agent sessions survive Emdash crashes or disconnects
- **Power user flexibility**: Full tmux capabilities (split panes, scrollback, detach/reattach)
- **SSH remote synergy**: Especially valuable for SSH remote projects where the host runs headless

## Technical Considerations
- Wrap the PTY spawn command: instead of running `claude` directly, run `tmux new-session -d -s {name} 'claude ...' && tmux attach -t {name}`
- Handle session lifecycle: detect existing sessions on startup, reattach vs create new
- Ensure Emdash's terminal emulator (xterm.js) works correctly as a tmux client
- Consider tmux availability check and graceful fallback if tmux is not installed
- Could extend to support zellij as an alternative multiplexer in the future

## Community Interest
Multiple users have requested this — both for tmux and zellij-based persistent sessions.

DevDesai444:
## Summary

Closes #1215  
Refs #1518

This PR turns persistent PTY handling into a configurable session-backend abstraction instead of treating it as a tmux-only feature.

Both issues are really about the same underlying problem: Emdash currently owns PTY lifetime directly, which makes long-running agent sessions fragile across disconnects, app restarts, and multi-device workflows. This change moves session lifetime behind a backend layer so Emdash can work with persistent terminal multiplexers instead of managing persistence itself.

As a result, `tmux` and `zellij` are now supported as first-class session backends for agent PTYs.

## What This Solves

- Makes persistent/resumable PTY sessions a real product concept instead of a tmux-specific special case
- Preserves the existing optional `tmux` workflow from #1215
- Adds `zellij` as a supported backend to address the main architectural need from #1518
- Allows local and remote agent sessions to reconnect to an existing persistent session instead of always starting from scratch
- Improves the path toward safer app restarts and reconnects for long-running agents

## What Changed

- Added a shared `sessionBackend` abstraction with supported values:
  - `none`
  - `tmux`
  - `zellij`
- Added shared parsing/normalization helpers for session backend config
- Updated lifecycle/config resolution so projects can opt into a backend through config
- Preserved backward compatibility with legacy `tmux: true` config
- Reworked PTY startup and cleanup logic to operate on a generic persistent backend instead of only tmux
- Added local `zellij` session support
- Added remote `zellij` session support for provider-backed PTYs
- Updated preload, IPC, and renderer typings so the renderer knows whether a PTY is backed by a persistent session backend
- Updated terminal snapshot behavior so renderer-side snapshotting is disabled for any persistent backend, not just tmux
- Replaced the old tmux toggle in project settings with a backend selector
- Added tests covering local and remote zellij flows

## Behavior After This PR

- `none` keeps the current non-persistent PTY behavior
- `tmux` keeps the existing persistent-session model, but now through the shared backend abstraction
- `zellij` becomes a first-class backend option for persistent/resumable agent sessions

This means Emdash now has a clean architectural path for persistent PTY sessions regardless of which multiplexer is used.

## Why This Is The Right Abstraction

Issue #1215 asks for optional persistent tmux-backed sessions. Issue #1518 asks for zellij as a first-class backend with better persistence ergonomics.

Those are not two unrelated features; they are two versions of the same requirement: PTY session lifetime should be delegated to a resumable backend rather than being tied directly to the app connection. This PR addresses that shared boundary and makes the implementation extensible instead of hardcoding more tmux-specific logic.

## Compatibility

- Existing tmux-based configuration continues to work
- Legacy `tmux: true` config is still recognized
- New config can use `sessionBackend: "none" | "tmux" | "zellij"`

## Non-Goals In This PR

- This PR does not make `zellij` the default backend
- This PR does not auto-install `zellij` on remote hosts
- This PR does not claim full “safe by default on cmd+q” behavior for all users out of the box

Those are product/defaulting decisions that can build on top of this backend abstraction in follow-up work.

## Testing

- Added PTY manager coverage for local zellij startup
- Added PTY IPC coverage for remote zellij startup and wiring
- Updated tests to cover the new session-backend behavior instead of tmux-only branching

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable persistent session backends (tmux and zellij alternatives)
  * Tasks now track and display last activity time
  * Added multiple task sorting modes: by last activity, creation date, alphabetically, or manual reordering
  * Users can manually reorder tasks in the project sidebar

### DIFF
--- a/agents/risky-areas/pty.md
+++ b/agents/risky-areas/pty.md
@@ -14,7 +14,7 @@
 - PTY cleanup and exit handling
 - resize behavior
 - shell quoting and Windows command wrapping
-- tmux lifecycle
+- persistent session backend lifecycle (`tmux`, `zellij`)
 - provider-specific resume/session behavior
 - env passthrough safety
 
--- a/agents/workflows/worktrees.md
+++ b/agents/workflows/worktrees.md
@@ -23,6 +23,7 @@ Current supported keys:
 - `scripts.run`
 - `scripts.teardown`
 - `shellSetup`
+- `sessionBackend`
 - `tmux`
 
 ## Rules
@@ -30,4 +31,4 @@ Current supported keys:
 - do not hardcode worktree paths; use service helpers
 - use lifecycle config for repo-specific bootstrap and teardown behavior
 - `shellSetup` runs inside each PTY before the interactive shell starts
-- tmux wrapping is project-configurable and affects PTY lifecycle behavior
+- persistent session backends are project-configurable and affect PTY lifecycle behavior
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -114,6 +114,11 @@ contextBridge.exposeInMainWorld('electronAPI', {
   ptyResize: (args: { id: string; cols: number; rows: number }) =>
     ipcRenderer.send('pty:resize', args),
   ptyKill: (id: string) => ipcRenderer.send('pty:kill', { id }),
+  ptyKillPersistentSession: (id: string) =>
+    ipcRenderer.invoke('pty:killPersistentSession', { id }) as Promise<{
+      ok: boolean;
+      error?: string;
+    }>,
   ptyKillTmux: (id: string) =>
     ipcRenderer.invoke('pty:killTmux', { id }) as Promise<{ ok: boolean; error?: string }>,
 
--- a/src/main/services/DatabaseService.ts
+++ b/src/main/services/DatabaseService.ts
@@ -1,5 +1,5 @@
 import type sqlite3Type from 'sqlite3';
-import { and, asc, desc, eq, isNull, ne, or, sql } from 'drizzle-orm';
+import { and, asc, desc, eq, getTableColumns, isNull, ne, or, sql } from 'drizzle-orm';
 import { readMigrationFiles } from 'drizzle-orm/migrator';
 import { resolveDatabasePath, resolveMigrationsPath } from '../db/path';
 import { getDrizzleClient } from '../db/drizzleClient';
@@ -51,6 +51,7 @@ export interface Task {
   metadata?: any;
   useWorktree?: boolean;
   archivedAt?: string | null;
+  lastActivityAt?: string | null;
   createdAt: string;
   updatedAt: string;
 }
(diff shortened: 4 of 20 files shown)


arnestrickmann:
Hi @DevDesai444,

Thank you for your PR and for your interest in contributing to Emdash. 

We've made some major re-designs and implementations around how we are spawning and handling terminals in the new Emdash version. 

Which brings me to: 

---

Just wanted to flag - we're merging our new v1 into main today!

We launched the beta last week and already 25% of our users have switched over. 

Would love to hear what you think once you've had a chance to try it out and would love to know how we can make Emdash better for you. 

Drop your thoughts here or come chat on Discord if thats preferred. 
https://discord.com/invite/f2fv7YxuR2

V1 Preview:  
CleanShot 2026-04-24 at 16 47 09@2x

PR #1434, opened 2026-03-12
feat(terminal): detect and open URLs from terminal output

yashdev9274:
- Added functionality to extract URLs from terminal output during the 'run' phase.
- Implemented a dropdown to select and open detected URLs in the browser.
- Introduced a  function to standardize URL formats.
- Updated electron API types to include  for URL handling.

## Summary

This PR adds a feature that allows users to quickly open detected localhost URLs (from running dev servers) directly in their system browser from the run terminal toolbar.

## Fixes 
 Fixes #1369  

## Snapshot
 
https://github.com/user-attachments/assets/618b883b-6add-418a-b102-2b1b31719354

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] This change requires a documentation update
 
## Mandatory Tasks

- [x] I have self-reviewed the code
- [x] A decent size PR without self-review might be rejected

## Checklist

- [x] I have read the contributing guide
- [x] My code follows the style guidelines of this project (`pnpm run format`)
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have checked if my PR needs changes to the documentation
- [x] I have checked if my changes generate no new warnings (`pnpm run lint`)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked if new and existing unit tests pass locally with my changes

PR #1583, opened 2026-03-25
feat(config): add support for browser url

kchung:
## Summary
Add the ability to optionally define what URL the in-app browser opens, this removes a lot of the guess work about what to use because it'll be user provided. The Emdash env vars are exposed so you can do things like:

```
https://localhost:$EMDASH_PORT
```

### Use case
For this particular use case, we sync our development code to Kubernetes which would available at a specific url like:

```
https://${USER}-${EMDASH_TASK_NAME}.internal.dev
```

Exposing this configuration allows the user to define how they want to open the URL per worktree

## Fixes 
N/A

## Snapshot
image

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (refactoring code, technical debt, workflow improvements)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (does not change functionality, e.g. code style improvements, linting)
- [ ] This change requires a documentation update
 
## Mandatory Tasks

- [X] I have self-reviewed the code
- [X] A decent size PR without self-review might be rejected

## Checklist

- [X] I have read the contributing guide
- [X] My code follows the style guidelines of this project (`pnpm run format`)
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have checked if my PR needs changes to the documentation
- [X] I have checked if my changes generate no new warnings (`pnpm run lint`)
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] I haven't checked if new and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * Added an "Open Browser URL" project setting and editor field (supports env-variable expansion and task-specific interpolation); browser toggle now consults this configured URL (per-task) before falling back to auto-detection.
* **Documentation**
  * Project config docs updated with Open Browser URL usage, examples, and blank-to-auto-detect behavior.
* **Tests**
  * Added tests for preview-URL resolution, interpolation, and error cases.

### DIFF
--- a/src/main/services/fsIpc.ts
+++ b/src/main/services/fsIpc.ts
@@ -880,6 +880,42 @@ export function registerFsIpc(): void {
     }
   );
 
+  // Resolve the previewUrl from .emdash.json, expanding $VAR references using
+  // process.env merged with EMDASH task-specific vars supplied by the caller.
+  ipcMain.handle(
+    'fs:resolvePreviewUrl',
+    async (_event, args: { projectPath: string; taskEnvVars?: Record<string, string> }) => {
+      try {
+        const { projectPath, taskEnvVars } = args;
+        if (!projectPath) return { success: false, error: 'Missing projectPath' };
+        const configPath = path.join(projectPath, '.emdash.json');
+        if (!fs.existsSync(configPath)) return { success: true, url: null };
+        const raw = fs.readFileSync(configPath, 'utf8');
+        const config = JSON.parse(raw);
+        const template: string | undefined = config?.previewUrl;
+        if (!template || typeof template !== 'string' || !template.trim()) {
+          return { success: true, url: null };
+        }
+        // Expand $VAR and ${VAR} references: EMDASH vars take priority, then process.env
+        const env: Record<string, string> = {
+          ...(process.env as Record<string, string>),
+          ...(taskEnvVars || {}),
+        };
+        const resolved = template.replace(
+          /\$\{([^}]+)\}|\$([A-Za-z_][A-Za-z0-9_]*)/g,
+          (_m, braced, bare) => {
+            const key = braced || bare;
+            return Object.prototype.hasOwnProperty.call(env, key) ? env[key] : _m;
+          }
+        );
+        return { success: true, url: resolved.trim() };
+      } catch (error) {
+        console.error('fs:resolvePreviewUrl failed:', error);
+        return { success: false, error: 'Failed to resolve preview URL' };
+      }
+    }
+  );
+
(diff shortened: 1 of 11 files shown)


arnestrickmann:
Hi @kchung,

Really like the direction here. I think there are a couple things worth tightening up before merge, and curious what you think:

1. resolvePreviewUrl currently expands against full process.env, which feels a bit risky since openBrowserUrl is project-controlled.
2. when openBrowserUrl is set, BrowserToggleButton returns early, so the existing preview setup/start flow doesn’t run.
3. this also looks local-only right now, while ConfigEditorModal already supports remote .emdash.json over SSH.
4. BrowserToggleButton seems to be rebuilding the EMDASH_* vars manually instead of reusing the shared helper, so values like task name / default branch may drift.

Does that all sound right to you? If so, I’d lean toward tightening those up before merge.

yashdev9274:
> Hi @kchung,
> 
> Really like the direction here. I think there are a couple things worth tightening up before merge, and curious what you think:
> 
> 1. resolvePreviewUrl currently expands against full process.env, which feels a bit risky since openBrowserUrl is project-controlled.
> 2. when openBrowserUrl is set, BrowserToggleButton returns early, so the existing preview setup/start flow doesn’t run.
> 3. this also looks local-only right now, while ConfigEditorModal already supports remote .emdash.json over SSH.
> 4. BrowserToggleButton seems to be rebuilding the EMDASH_* vars manually instead of reusing the shared helper, so values like task name / default branch may drift.
> 
> Does that all sound right to you? If so, I’d lean toward tightening those up before merge.

I've already created a pr for this, do review it when you get time 

https://github.com/generalaction/emdash/pull/1434

kchung:
@arnestrickmann thanks yeah CodeRabbit caught some good ones, i dialed back exposing all env vars big concern there, i'll have to test the remote feature some more too I didn't know about that feature. pretty cool

@yashdev9274 ah interesting! i think we're solving for tangential things, specifically in this PR we are explicitly defining what URL we should expect the browser to open vs best-guessing via console output

src/main/services/fsIpc.ts
```diff
@@ -72,6 +72,8 @@ const ALLOWED_IMAGE_EXTENSIONS = new Set<string>([
 ]);
 const DEFAULT_ATTACHMENTS_SUBDIR = 'attachments' as const;
 
+const ALLOWED_HOST_VARS = ['USER'];
```
kchung, line 75: the only thing i found useful was the `USER` env var

PR #1602, opened 2026-03-26
fix(nix): fix Nix flake package build for NixOS

ratierd:
## Summary

The Nix flake package (`nix build .#emdash`) was not functional on NixOS. This PR fixes three issues:

- **Empty pnpm dependency hash**: The `fetchPnpmDeps`/`pnpm.fetchDeps` hash was set to `""`, causing the build to fail on the first attempt. Set the correct hash.
- **Native modules not compiled for Electron**: `keytar.node`, `sqlite3`, and `node-pty` were never rebuilt for the Electron runtime. Added `electron-rebuild` step with pre-fetched Electron headers so it works in the Nix sandbox (no network access).
- **Missing runtime shared libraries**: The Electron binary couldn't find `libGL.so.1`, GTK, X11, and other libraries at runtime. Added `autoPatchelfHook` to automatically patch ELF binaries and `makeWrapper` to set `LD_LIBRARY_PATH`.

## Question for maintainers

Was the empty `hash = ""` in `pnpmDeps` intentional — e.g. is the package output considered WIP, with the `devShell` being the primary supported output?

An empty hash in a Nix fixed-output derivation will always fail on first build (Nix then prints the correct hash in the error). There's no "unpinned" mode for FODs — Nix requires an exact hash for reproducibility. The tradeoff of pinning it is that it needs updating whenever `pnpm-lock.yaml` changes, but that's the standard pattern in nixpkgs. A CI check on PRs touching `pnpm-lock.yaml` or `flake.nix` could catch hash drift automatically.

Happy to adjust the approach based on your intent here.

## Test plan

- [x] `nix build .#emdash` completes successfully
- [x] `result/bin/emdash` launches without errors on NixOS
- [x] Native modules (keytar for credential storage, sqlite3, node-pty) work correctly

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -59,6 +59,19 @@
           cp ${electronLinuxZip} $out/electron-v${electronVersion}-linux-x64.zip
         '';
 
+        # Pre-fetch Electron headers for native module compilation (node-gyp)
+        electronHeaders = pkgs.fetchurl {
+          url = "https://www.electronjs.org/headers/v${electronVersion}/node-v${electronVersion}-headers.tar.gz";
+          sha256 = "sha256-Q+c8G4nIRoJL/0uAYVYY2hrnFgvmkKB6RC3nxJtFYzU=";
+        };
+
+        # Create a node-gyp cache directory with the Electron headers
+        electronHeadersDir = pkgs.runCommand "electron-headers" {} ''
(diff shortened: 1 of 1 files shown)


arnestrickmann:
Thanks for the fix! @ratierd

PR #1624, opened 2026-03-28
fix(opencode): preserve explicit config dir overrides

P-de-Jong:
## Summary

Emdash was overwriting an existing `OPENCODE_CONFIG_DIR` with its own per-PTY OpenCode config dir.

In my case, `OPENCODE_CONFIG_DIR` already points at a company-managed config repo, so that override caused OpenCode to ignore the shared config.

This change preserves an explicit `OPENCODE_CONFIG_DIR` if one is already set. If it cannot write the notify plugin there, it falls back to Emdash's per-PTY config dir. If no override is set, the existing per-PTY behavior stays the same.

## Testing

- `pnpm exec vitest run src/test/main/ptyManager.test.ts src/test/main/ptyIpc.test.ts`
- `pnpm exec prettier --check src/main/services/OpenCodeHookService.ts src/main/services/ptyManager.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenCode plugin install with clearer fallback behavior and additional error logging when an explicit config directory write fails.

* **Refactor**
  * Separated plugin-writing and runtime-env application to support explicit OPENCODE_CONFIG_DIR while retaining per-PTY defaults.

* **Tests**
  * Added and expanded tests covering explicit-config installs, fallback to per-PTY paths, and test isolation of OPENCODE_CONFIG_DIR.

### DIFF
--- a/src/main/services/OpenCodeHookService.ts
+++ b/src/main/services/OpenCodeHookService.ts
@@ -99,21 +99,38 @@ export class OpenCodeHookService {
     ].join('\n');
   }
 
-  static writeLocalPlugin(ptyId: string): string {
-    const configDir = OpenCodeHookService.getLocalConfigDir(ptyId);
+  static writePlugin(configDir: string): boolean {
     const pluginsDir = path.join(configDir, 'plugins');
     const pluginPath = path.join(pluginsDir, OPEN_CODE_PLUGIN_FILE);
 
     try {
       fs.mkdirSync(pluginsDir, { recursive: true });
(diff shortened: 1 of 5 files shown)


arnestrickmann:
Hi @P-de-Jong,

Thanks a lot for this PR; and for contributing to Emdash.

Good news: this change is still very relevant in v1. We’ve now shipped v1 and merged it into main, so the codebase has moved quite a bit since this PR was opened.

It would be amazing if you could re open this PR based on the new architecture. 

New Emdash:
CleanShot 2026-04-29 at 11 53 19@2x

Issue #1467, opened 2026-03-13
Bug: Text input fields randomly stop accepting keyboard input


arnestrickmann:
## Bug Report

**Description**
Text input fields in the app randomly stop accepting keyboard input. The user cannot type in text boxes at all — no characters appear when pressing keys. The issue has been observed in multiple locations and seems to occur intermittently.

## Known Occurrences
- **Commit message field** — Unable to type when opening the changes/diff view
- **General text boxes** — Reported across the IDE more broadly

## Steps to Reproduce
1. Use the app normally
2. At some point (seemingly random), text input fields stop responding to keyboard input
3. Observed to trigger when opening the changes/diff view page, but may happen in other contexts too

## Expected Behavior
All text input fields should consistently accept keyboard input at all times.

## Actual Behavior
Text fields become completely unresponsive to typing. The issue appears intermittently and may require an app restart to resolve.

## Likely Cause
Focus management issue — something (possibly the terminal, Monaco editor, or diff view) is capturing keyboard events and not releasing them, preventing input fields from receiving keystrokes. Could be:
- A focused terminal or xterm.js instance swallowing key events
- Monaco editor grabbing focus in the background
- An event listener not being cleaned up when switching views
- A focus trap that doesn't release properly when the diff/changes panel opens

dagss:
I can confirm this issue, it happens like once every 30 minutes for me. Rather annoying to have to restart emdash all the time.

When it happened just now I think it happened when I hit "Esc" to step out of git details view.

arnestrickmann:
Hi @dagss, 
Thank you for flagging this again. 

Could you, next time it happens, give me more context about the situation you’re in? 

Until now, I haven’t been able to reproduce this.

QuinnBast:
This is happening to me very frequently in 1.0. Maybe this is specific to linux users? In order to fix it, I need to click "Toggle Terminal" to open the lower terminal, and then close it to be able to type in the main chat window

PR #1631, opened 2026-03-29
fix: resolve 15-second SSH delay with fish shell prompt detection

### LINKED ISSUE #1597: [bug]: 15-second delay when connecting to SSH remote with fish shell
### Bug Description

When connecting to a remote host with fish as the default login shell, there is a consistent ~15-second hang before the CLI agent begins executing init commands. The SSH handshake completes successfully, and the PTY displays fish's standard greeting ("Welcome to fish") followed by a normal command prompt — but the emdash does not detect the prompt and waits until a 15-second timeout fires before proceeding. This makes connection to SSH projects with fish feel broken.

### Steps to Reproduce

1. Configure an SSH remote host with fish as the default login shell (`chsh -s /usr/bin/fish`).
2. Add the SSH connection in emdash.
3. Open a remote project on that host.

**Actual:** 15 seconds delay before the agent appeared.

**Expected:** The agent appeared right away.

### emdash Version

0.4.41

### Operating System

macOS 26.3.1 (local), Ubuntu 24.04 with fish shell (remote)

### Additional Context

_No response_

jasperan:
## Summary

Fixes #1597. The previous fix (#1615) addressed a different issue (focus-stealing, #1467) and didn't touch prompt detection.

- **Root cause**: `waitForShellPrompt` tested each PTY data chunk independently against the prompt regex. Fish shell renders its prompt in multiple small writes over SSH (colored segments, escape sequences, prompt character). When `>` arrived in its own TCP segment without preceding non-whitespace context, the regex failed and fell back to the 15-second timeout.
- **Secondary cause**: Fish's cursor save/restore escapes (`\e7`/`\e8`) and OSC-ST terminators weren't stripped, leaving non-whitespace bytes after `>` that broke the `\s*$` anchor.

**Fix (2 files, 3 changes):**
- Accumulate PTY chunks into a rolling buffer (4KB cap) so prompts split across TCP segments match against the full accumulated output
- Pass additional strip options (`stripOtherEscapes`, `stripOscSt`, `stripCarriageReturn`) to handle fish's escape sequences
- Expand `stripOtherEscapes` regex to cover Fe/Fp/Fs single-byte escapes (was only matching `[A-Za-z]`, now matches all except CSI/OSC openers)

## Test plan

- [x] 7 new test cases: fish prompt, TCP-split prompt, bare `>` rejection, Fe escapes, OSC-ST
- [x] All 27 `waitForShellPrompt` tests pass
- [x] All 9 `stripAnsi` tests pass (no regressions)
- [x] Full suite: 78 files, 782 tests pass
- [x] Type-check clean, format clean, lint clean on changed files
- [ ] Manual verification: connect to SSH remote with fish shell, confirm no 15s delay

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shell prompt detection for fish-style prompts (e.g., "user@host ~>"), including prompts split across multiple terminal/chunked inputs.
  * Expanded removal of noisy terminal escape bytes and OSC terminators to reduce false positives while preserving valid control sequences.

* **Tests**
  * Added tests covering split-chunk prompts, escape-sequence noise resilience, and a negative case ensuring a bare "> " alone does not trigger detection.

### DIFF
--- a/src/main/utils/__tests__/waitForShellPrompt.test.ts
+++ b/src/main/utils/__tests__/waitForShellPrompt.test.ts
@@ -379,7 +379,7 @@ describe('waitForShellPrompt', () => {
     expect(pty.write).toHaveBeenCalledWith('cd /foo\n');
   });
 
-  it('detects prompt when > arrives alone after accumulated context', () => {
+  it('detects prompt when $ arrives alone after accumulated context', () => {
     const pty = createMockPty();
     waitForShellPrompt({
       subscribe: pty.subscribe,


arnestrickmann:
@jasperan thank you for taking this on again! 
Would be amazing if you could resolve the merge conflicts in waitForShellPrompt.ts - happy to merge then!

jasperan:
done @arnestrickmann

arnestrickmann:
Thanks!

PR #1655, opened 2026-04-02
fix(ui): use Button component for View PR button consistency

jeremyhalin:
## Why

The "View PR" button had inconsistent styling compared to adjacent action buttons, appearing noticeably smaller due to custom inline styles (text-[10px], px-1.5, py-0.5) instead of using the standardized Button component.

## What

- Replaced the raw `<button>` element with the Button component
- Applied consistent `variant="outline"` and `size="sm"` props
- Updated className to match standard button styling (h-8, px-2, text-xs)
- Adjusted icon spacing (ml-1.5) to align with component padding

This ensures visual consistency across all action buttons in the FileChangesPanel.

**Before**
Capture d’écran 2026-04-02 à 12 18 20

**After**
Capture d’écran 2026-04-02 à 12 15 39

## Release Notes

* **Style**
  * Updated the appearance of the PR link button in the file changes panel with refined styling and icon sizing.


JiwaniZakir (commented):
The swap from a raw `<button>` to the `<Button variant="outline" size="sm">` component in `FileChangesPanel.tsx` is a solid consistency win — the old element was manually replicating styles that the design system already provides. One thing worth checking: the original button used `text-[10px]` and `py-0.5 px-1.5`, making it quite compact, while `size="sm"` on the `Button` component likely applies its own padding and a fixed height of `h-8` (32px). Verify that `h-8 shrink-0 px-2 text-xs` in the `className` override actually wins over the component's internal size styles — depending on how `Button` applies `cn()` merging, the `size="sm"` padding might not be fully overridden, resulting in a slightly larger button than intended in the tight toolbar layout. Also, `ArrowUpRight` went from `size-3` (12px) to `ml-1.5 h-3.5 w-3.5` (14px); confirm this doesn't visually crowd the icon against the label text at the PR state strings like `"PR Merged"`, which are already longer than `"View PR"`. Worth a quick visual pass across all three label branches (`isDraft`, `OPEN`, and the fallback state) to make sure none of them clip or overflow.

Issue #1472, opened 2026-03-14
Feature: Agent presets for per-task CLI configuration (model, flags, etc.)


arnestrickmann:
## Feature Request

**Description**
Allow users to define reusable agent "presets" that configure CLI flags and settings per task. Instead of manually interrupting an agent after launch to tweak settings (model, reasoning effort, fast mode, etc.), users should be able to select a preset at task creation time.

## Problem
Currently, to run a task with specific agent settings (e.g., Claude Code with Sonnet, low reasoning effort, fast mode), users have to:
1. Create the task and let the agent start
2. Immediately interrupt the agent
3. Manually update the CLI settings for that session
4. Send "Continue" to resume

This is tedious and error-prone, especially when frequently switching between different configurations for different types of tasks.

### Preset definition
Users can define named presets in settings, e.g.:

- **"Quick tasks"** — Sonnet 4.6, low reasoning effort, fast mode
- **"Deep work"** — Opus 4.6, high reasoning effort
- **"Budget mode"** — Haiku, low reasoning effort

Each preset maps to specific CLI flags for the target agent (e.g., `--model`, `--reasoning-effort`, etc.).

### Preset selection
- When creating a new task, a preset dropdown is available alongside the agent selector
- Default preset can be configured per project
- Presets are agent-specific (Claude Code presets vs Codex presets, etc.)

### Supported configuration per preset
For Claude Code:
- Model (`--model sonnet-4-6` / `opus-4-6` / `haiku-4-5`)
- Reasoning effort
- Fast mode
- Any other supported CLI flags

For Codex:
- Model selection
- Approval policy
- Sandbox mode

Extensible to other agents as they add CLI flags.

## Why This Matters
Different tasks have different complexity levels. Simple tasks (typo fixes, small refactors) don't need the most expensive model. Complex tasks (architecture changes, multi-file refactors) benefit from maximum capability. Presets let users match the right tool configuration to the task without manual CLI gymnastics every time.

glassworks-projects:
I don't think this was ever done. I am also requesting this feature—not just for Claude, but for all providers that support it (my default is Codex). It would also be nice to be able to save a default effort level so I don't have to change it each time.

PR #1656, opened 2026-04-02
feat(claude): add model, effort, and fast mode selectors to task/agent creation

ubuntudroid:
## Summary

Adds a collapsible **Claude options** section to the New Task dialog and the Add Agent dialog. Users can now configure three Claude Code CLI options before starting a session — without having to interrupt the agent after launch.

- **Model** — select from models fetched dynamically from the Anthropic `/v1/models` API (1-hour in-process cache, hardcoded fallback when the API is unavailable). Passed as `--model <id>`. Bracket-containing IDs like `claude-sonnet-4-6[1m]` are now correctly shell-quoted (fixes a zsh glob-expansion crash).
- **Effort** — `--effort low|medium|high|max` to control reasoning depth.
- **Fast mode** — `--settings '{"fastMode":true}'` at startup. Only shown when the selected model has `supportsFast: true` in the API response (derived from the model ID containing `"opus"`). Automatically hidden and cleared when switching to a non-Opus model.

The same options are available in the **Add Agent to Task** modal (stored in conversation metadata, read back when that conversation's PTY starts).

## Fixes

Related to #1472, but no preset support yet and only for Claude.

## Snapshot

| Create Task | Add Agent to Task |
|--------|--------|
| CleanShot 2026-04-02 at 12 27 23@2x | CleanShot 2026-04-02 at 12 28 09@2x | 

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Mandatory Tasks

- [x] I have self-reviewed the code

## Checklist

- [x] I have read the contributing guide
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have checked if my PR needs changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

---

### Implementation notes

**New files**
- `src/renderer/components/ClaudeModelSelect.tsx` — Radix Select wrapper; fetches model list on mount, exposes `onModelsLoaded` callback
- `src/renderer/components/ClaudeOptionsSection.tsx` — collapsible accordion (same style as "Advanced options") containing the three controls
- `src/test/main/ptyManager.modelFastMode.test.ts` — PTY arg-building tests
- `src/test/renderer/claudeModelSelect.test.ts` — sentinel round-trip tests
- `src/test/renderer/taskCreationService.modelFastMode.test.tsx` — metadata persistence tests

**Key design decisions**
- `supportsFast` is computed server-side from the model ID (`includes("opus")`) since the Anthropic API does not return capability metadata. The fallback list also has explicit values.
- Fast mode uses `--settings '{"fastMode":true}'` (not a `/fast` keystroke injection and not a `--fast` flag — neither of those exist as startup options).
- `--effort max` is valid per `claude --help`; all four levels are offered.
- Shell-path spawn now uses `quoteShellArg` (already used for remote SSH paths) instead of a partial regex, making it safe for any special characters in model IDs.

## Summary by CodeRabbit

* **New Features**
  - Per-task Claude model selector with live model listing (via Anthropic) and fallback list
  - "Effort" level setting (low/medium/high/max) for Claude tasks
  - "Fast mode" toggle surfaced where supported (Opus models)
  - Claude options available in task creation, chat UI, and terminal sessions; runtime/startup now respects chosen Claude settings

* **Documentation**
  - Added Claude configuration details to provider and task docs

### DIFF
--- a/docs/content/docs/providers.mdx
+++ b/docs/content/docs/providers.mdx
@@ -55,6 +55,12 @@ Not all providers support the same features. Here's what varies:
 
 **Initial prompt**: Pass instructions when starting the agent. Most providers support this.
 
+**Model selection** _(Claude Code only)_: Choose the Claude model to use per task. The list is fetched live from the Anthropic API.
+
+**Effort level** _(Claude Code only)_: Set reasoning depth at startup — low, medium, high, or max.
+
+**Fast mode** _(Claude Code, Opus models only)_: Enable fast streaming output.
+
 ## Adding a Provider
 
 Providers are detected automatically when you install them. Open Emdash, go to Settings, and click "Refresh" to scan for newly installed CLIs.
--- a/docs/content/docs/tasks.mdx
+++ b/docs/content/docs/tasks.mdx
@@ -46,6 +46,16 @@ Each agent gets its own worktree. Compare approaches and pick the best result. S
 
 If tasks start local servers, use `EMDASH_PORT` in your Project config script fields to avoid port collisions between parallel tasks. See [Project Configuration](/project-config).
 
+## Claude Options
+
+When Claude Code is selected, a **Claude options** section appears. Expand it to configure:
+
+- **Model**: Choose a specific Claude model (e.g. `claude-opus-4-6`, `claude-sonnet-4-6[1m]`). Models are fetched from the Anthropic API when an `ANTHROPIC_API_KEY` is set, with a hardcoded fallback list otherwise. Defaults to whatever Claude Code uses by default.
+- **Effort**: Control reasoning depth — `low`, `medium`, `high`, or `max`. Leave at Default to let Claude decide.
+- **Fast mode**: Enable Claude's fast streaming mode. Only shown when an Opus model is selected, as fast mode is currently supported for Opus models only.
+
+These options are also available in the **Add Agent to Task** dialog.
+
 ## Advanced Options
 
 Click **Advanced Options** when creating a task to access:
--- a/src/main/ipc/connectionsIpc.ts
+++ b/src/main/ipc/connectionsIpc.ts
@@ -45,7 +45,12 @@ const MODEL_CACHE_TTL_MS = 60 * 60 * 1000; // 1 hour
 
 /** Fetch available models from the Anthropic API. Returns null when unavailable. */
 async function fetchAnthropicModels(): Promise<ClaudeModel[] | null> {
-  const apiKey = process.env.ANTHROPIC_API_KEY;
+  // Prefer a key saved in the provider custom config (e.g. via Settings → Providers)
+  // so users who configure Claude that way also get live model lists.
+  const configuredKey = getProviderCustomConfig('claude')?.env?.['ANTHROPIC_API_KEY'];
+  const apiKey =
+    (typeof configuredKey === 'string' && configuredKey.trim() ? configuredKey.trim() : null) ??
+    process.env.ANTHROPIC_API_KEY;
   if (!apiKey) return null;
 
   return new Promise((resolve) => {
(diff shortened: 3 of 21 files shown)


JiwaniZakir (commented):
In `fetchAnthropicModels` (connectionsIpc.ts), when a non-2xx status code is received, `resolve(null)` is called and the function returns early without draining the response body. In Node.js HTTP, failing to consume the response stream can prevent socket reuse and leak resources — at minimum, `res.resume()` should be called before returning.

The module-level `claudeModelCache` has no invalidation path tied to the API key. If a user updates their `ANTHROPIC_API_KEY` in provider settings mid-session, `getProviderCustomConfig` will return the new key, but the cached model list (keyed only by TTL) will be served for up to another hour. Consider busting the cache when the provider config is updated via `providers:updateCustomConfig`.

The fallback model ID `claude-sonnet-4-6[1m]` in `CLAUDE_FALLBACK_MODELS` contains square brackets, which is an unusual format. If this string is passed directly as a `--model` CLI argument to Claude Code without validation or quoting, it could be misinterpreted by the shell. It's worth verifying this ID is the canonical form Claude Code accepts, and adding a note or normalization step if the API returns a differently formatted ID for the same model.

ubuntudroid:
Thanks for the thorough review @JiwaniZakir!

**Point 1 (res.resume())** — Fixed in 702d27c. Added `res.resume()` before `resolve(null)` on non-2xx responses so the socket is properly drained and the connection can be reused.

**Point 2 (cache invalidation)** — Fixed in 702d27c. `claudeModelCache` is now set to `null` inside the `providers:updateCustomConfig` handler whenever `providerId === 'claude'`, so a changed API key takes effect on the next fetch rather than waiting out the 1-hour TTL.

**Point 3 (square brackets in model ID)** — `claude-sonnet-4-6[1m]` is the canonical ID returned by the Anthropic `/v1/models` API for the 1M-context Sonnet variant. The shell quoting issue it caused (zsh glob-expanding `[1m]`) was actually an original bug this PR fixed — the PTY layer now passes all CLI args through `quoteShellArg` so this is handled correctly. Added a comment to `CLAUDE_FALLBACK_MODELS` to make this explicit for future readers.

JiwaniZakir:
The coderabbit flags are all valid — the `||` collapsing `false` and `undefined` for fast mode is the most subtle one and will cause unexpected behavior for conversations that explicitly disable it; that needs a `?? taskDefault` instead. The stale cache write is also a real race condition worth fixing before this lands. I'll push a fix for the provider env key lookup and the effort/fastMode validation whitelist in the spawn paths as well.

ubuntudroid:
@JiwaniZakir Following up on the cache invalidation race condition you identified — CodeRabbit caught it too in the latest round of reviews, so we went ahead and fixed it in 388ceb09.

The fix uses a generation counter: `claudeModelCacheGeneration` is incremented whenever `providers:updateCustomConfig` clears the cache for the `claude` provider. The `providers:listModels` handler snapshots the generation before `await fetchAnthropicModels()` and only writes the result back if the generation is unchanged, so any in-flight fetch started under a superseded API key silently discards its result rather than repopulating the cache.

JiwaniZakir:
The double-click issue on the collapsible section is likely because the first click updates state and triggers a re-render before the collapse logic fires — wrapping the toggle in a single handler that updates both the display state and the saved state atomically should fix it. For the cursor, adding `cursor: pointer` to the collapsible header element is a one-liner CSS fix. Both of these are straightforward to address before merging.

ubuntudroid:
> The double-click issue on the collapsible section is likely because the first click updates state and triggers a re-render before the collapse logic fires — wrapping the toggle in a single handler that updates both the display state and the saved state atomically should fix it. For the cursor, adding `cursor: pointer` to the collapsible header element is a one-liner CSS fix. Both of these are straightforward to address before merging.

@JiwaniZakir I'm not sure I get your last comment. It seems like you are referring to previously mentioned issues, but I couldn't find those - and I couldn't reproduce any issues around pointer or the collapsible section. Could you elaborate please?

PR #1659, opened 2026-04-04
fix: remove max-w-4xl and drop Ctrl+Z to let terminal fill available width

### LINKED ISSUE #1650: [bug]: Chat terminal has fixed max-width (max-w-4xl/896px), does not expand with window
### Bug Description

The main chat terminal container in `ChatInterface.tsx` and `MultiAgentTask.tsx` uses Tailwind's `max-w-4xl` class (896px max-width). This caps the terminal at roughly ~110 columns regardless of how wide the Emdash window is.

When using a wide monitor or full-screen window, the terminal stays capped at 896px and the extra horizontal space is wasted. The sidebar (Context, MCP, Todo panel) already takes ~280px, so on a 1920px screen the terminal only uses about 46% of available width.

### Affected code

- `src/renderer/components/ChatInterface.tsx:1252` — terminal container div: `max-w-4xl`
- `src/renderer/components/ChatInterface.tsx:1164` — tab bar container: `max-w-4xl`
- `src/renderer/components/MultiAgentTask.tsx:640` — multi-agent terminal: `max-w-4xl`
- `src/renderer/components/MultiAgentTask.tsx:723` — multi-agent input: `max-w-4xl`

The FitAddon and ResizeObserver chain works correctly — the terminal *would* resize to fill the container, but the CSS caps the container width.

### Expected Behavior

The chat terminal should scale with the window width, either:
- Remove `max-w-4xl` entirely and let the terminal fill available space
- Use a much wider cap (e.g., `max-w-[1600px]` or `max-w-screen-xl`)
- Make it configurable in settings

### Environment

- emdash v0.4.42
- macOS

yuzhichang:
## Summary

Two fixes for terminal UX issues:

1. **Terminal width**: Removes `max-w-4xl` from `ChatInterface.tsx` and `MultiAgentTask.tsx` terminal containers. The terminal was capped at 896px regardless of window width, wasting horizontal space on wide monitors.

2. **Ctrl+Z deadlock**: Drops `\x1a` (SIGTSTP) at the PTY write level in `writePty()`. When Ctrl+Z is pressed in emdash's terminal running an agent, the shell chain (`bash -ilc "claude; exec bash -il"`) would deadlock — the agent stops but `exec bash -il` never runs, leaving no foreground process to handle `fg`. Also removes `exec` from the shell chain so the resume shell runs as a proper foreground child. Ctrl+C (`\x03`) is preserved so users can still interrupt agent actions.

Fixes generalaction/emdash#1650

## Test plan

- [ ] Terminal fills available width on wide monitors
- [ ] Ctrl+Z does not hang the terminal or cause deadlocks
- [ ] Ctrl+C still interrupts agent actions within Claude

## Release Notes

* **Bug Fixes**
  * Fixed PTY input handling to prevent unwanted signal generation when pressing Ctrl+Z in the terminal.

* **UI Improvements**
  * Terminal containers now expand to full available width for optimal screen space utilization.

* **Improvements**
  * Enhanced shell command execution flow for better process management.

### DIFF
--- a/src/main/services/ptyManager.ts
+++ b/src/main/services/ptyManager.ts
@@ -1534,14 +1534,13 @@ export async function startPty(options: {
 
         const shellBase = (defaultShell.split('/').pop() || '').toLowerCase();
 
-        // After the provider exits, exec back into the user's shell (login+interactive)
         const resumeShell =
           shellBase === 'fish'
             ? `'${defaultShell.replace(/'/g, "'\\''")}' -i -l`
             : `'${defaultShell.replace(/'/g, "'\\''")}' -il`;
         const chainCommand = shellSetup
-          ? `${shellSetup} && ${commandString}; exec ${resumeShell}`
-          : `${commandString}; exec ${resumeShell}`;
+          ? `${shellSetup} && ${commandString}; ${resumeShell}`
+          : `${commandString}; ${resumeShell}`;
 
         // Always use the default shell for the -c command to avoid re-detecting provider CLI
         useShell = defaultShell;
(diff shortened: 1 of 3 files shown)


yuzhichang:
v1 does fill available width.

PR #1677, opened 2026-04-06
fix(terminal): add padding to expanded terminal modal content

arnestrickmann:
## Summary

Adds inner padding to the expanded terminal modal so the terminal content doesn't render flush against the modal edge.

## Changes

- Wrapped the terminal container `ref` in a padding wrapper with `p-2 pt-1` to match the embedded terminal inset
- Applied theme-aware background color (`bg-zinc-900` for dark, `bg-background` for light) to the padding wrapper
- Moved the click-to-focus handler to the outer wrapper so clicking the padding area also focuses the terminal

## Summary by CodeRabbit

* **Style**
  * Updated the expanded terminal modal's visual presentation with improved padding and background styling to ensure better visual consistency and alignment within the modal container.

### DIFF
--- a/src/renderer/components/ExpandedTerminalModal.tsx
+++ b/src/renderer/components/ExpandedTerminalModal.tsx
@@ -109,15 +109,19 @@ const ExpandedTerminalModal: React.FC<Props> = ({ terminalId, title, onClose, va
           </Button>
         </div>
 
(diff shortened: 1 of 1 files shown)


leandroalves-sc:
was the terminal supposed to be full size in the main panel?
because for me it's just using container size instead of the full width available

PR #1752, opened 2026-04-22
feat: auto-detect open PRs for task worktrees

casassg:
Tasks now auto-detect open GitHub PRs against their worktree branch and surface them in the titlebar, without the user having to open the repo-level PR list first.

- Added a `Resource<void>` in `PrStore` that calls `rpc.pullRequests.getPullRequestsForTask(..., invalidate=true)` so the existing service actually runs on its own.
- Triggers: task mount, `gitWorkspaceChangedChannel` (HEAD), `gitRefChangedChannel` (remote-refs, i.e. first push), and a 60s poll that pauses when the window is hidden.
- Threaded `taskId` through `workspaceRegistry.acquire` → `WorkspaceStore` → `PrStore` so the RPC can resolve the branch.
- Added a `<PrBadge>` chip in `ActiveTaskTitlebar` bound to `workspace.pr.currentPr`; renders only when a PR exists.
- Data propagation is unchanged: the RPC upserts and `onPrUpserted` emits `taskPrUpdatedChannel`, which `task-manager` already patches into `task.data.prs`.

## Example usage

1. Open a task whose branch already has an open PR on GitHub — the titlebar chip appears within a second of mount.
2. Push a task branch (`Push` / `Publish` in the titlebar popover), then `gh pr create --head <branch>` — the chip appears within ~1s without any UI refresh.
3. Open a PR for the branch externally via the GitHub web UI — the chip appears within ≤60s via the poll.

### DIFF
--- a/src/renderer/features/tasks/stores/pr-store.ts
+++ b/src/renderer/features/tasks/stores/pr-store.ts
@@ -20,6 +20,7 @@ export type MergeResult = { success: true } | { success: false; error: string };
 
 export class PrStore {
   readonly commitHistory: Resource<{ commits: Commit[]; aheadCount: number }>;
+  private readonly _taskPrLookup: Resource<void>;
 
   private _prFiles = new Map<string, Resource<GitChange[]>>();
   private _prCheckRuns = new Map<string, Resource<PrCheckRun[]>>();
@@ -28,6 +29,7 @@ export class PrStore {
   constructor(
     private readonly projectId: string,
     private readonly workspaceId: string,
+    private readonly taskId: string,
     private readonly repositoryStore: RepositoryStore,
     private readonly getPrs: () => PullRequest[]
   ) {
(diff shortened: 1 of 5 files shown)


arnestrickmann:
Hi @casassg, thanks for opening this PR - LGTM. great addition. 

Just as an FYI: later this week we’ll be switching from this older Emdash codebase to the new v1, which we’ve been actively working on and launched in beta last week. Around 25% of our user base has already moved over.

Would be great to hear your thoughts once you’ve had a chance to try it.

You can share feedback here or on Discord: https://discord.com/invite/f2fv7YxuR2

casassg:
> Hi @casassg, thanks for opening this PR - LGTM. great addition.

Was mostly a thing I was missing for me XD

> Just as an FYI: later this week we’ll be switching from this older Emdash codebase to the new v1, which we’ve been actively working on and launched in beta last week. Around 25% of our user base has already moved over.

I pushed to the v1 branch already, been using the beta this week already 😄 

> Would be great to hear your thoughts once you’ve had a chance to try it.

So far so good. I do have separate feedback though not specifically for the beta:

- Would love some way to pull comments from my GitHub PR into an agent (for example greptile comment into an agent).
- Would be nice to have some quick way to comment on diff and send to the agent as well.

Aside from those, I would say Emdash is a great app. The Linear integration and Github integration is great. Pushing branches on creation is a bit weird imho but can be disabled, so that's fine. 

The beta still has some rough edges:
- for example if you try to create a workspace for a branch which is already checked out in some other path, it errors but gets stuck 
- opencode integration hooks are not working I think

but nothing major

PR #1763, opened 2026-04-25
fix(pty): resolve race condition between direct-spawn exit handlers

### LINKED ISSUE #1519: Bug: Terminal does not accept keyboard input after agent process exits
## Bug Description

After an agent (e.g. Codex) finishes its session and exits back to the shell prompt, the terminal becomes unresponsive to keyboard input. The shell prompt is visible and the cursor is blinking, but no keystrokes are delivered to the terminal. This occurs in the multi-agent task view.

## Steps to Reproduce

1. Create a multi-agent task with Codex (or any agent)
2. Let the agent run and complete its work, or exit the agent session (e.g. via `/exit` or the agent hitting a rate limit)
3. The agent process exits and the shell prompt appears in the terminal
4. Try to type any command in the terminal

## Actual Behavior

The terminal does not accept any keyboard input. The shell prompt is displayed correctly but keystrokes are completely ignored. The only workaround is to restart the task or the app.

## Expected Behavior

After the agent process exits, the terminal should return to a normal interactive shell where the user can type commands freely.

## Possible Cause

The PTY session may not be re-enabling stdin after the agent process exits, or the xterm.js terminal instance may still have `disableStdin` set. It could also be a focus management issue where the terminal fails to reclaim keyboard focus after the agent exit event.

Related issues (similar symptoms but different triggers):

* [GEN-432](https://linear.app/general-action/issue/GEN-432/bug-terminal-is-unresponsive-to-keyboard-input-on-initial-startup) — Terminal unresponsive on initial startup
* [GEN-571](https://linear.app/general-action/issue/GEN-571/bug-text-input-fields-randomly-stop-accepting-keyboard-input) — Text input fields randomly stop accepting input

## Environment

* **emdash version:** latest (built from main @ d8cb465a)
* **OS:** macOS
* **Agent:** Codex (observed in multi-agent task view)

## Screenshot

The screenshot below shows the terminal after the Codex agent exited. The shell prompt is visible at the bottom but keyboard input is not accepted:

> Shell prompt is displayed (`~/De/B/w/ai-image-local-codex-6m0 on emdash/ai-image-local-codex-6m0`) but no keystrokes register.

andrassmuk:
## Summary

Fixes a bug where keyboard input stops working in the terminal after a direct-spawned agent (e.g. Codex) exits.

## Root cause

Two `onExit` handlers fire when the agent exits:

1. `ptyManager` starts an async shell respawn via `startPty`
2. `ptyIpc` runs synchronously right after — calls `removePtyRecord(id)` and emits `pty:exit` to the renderer

Because the async respawn hasn't completed yet, the renderer receives `pty:exit` first and tears down its listeners. By the time the new shell is ready, nothing is listening anymore — so keyboard input has no effect.

## Fix

- Skip cleanup in the `ptyIpc` `onExit` handler when the exiting PTY belongs to a direct spawn, so it doesn't race with the respawn
- Let the respawn callback own the full lifecycle (including cleanup if respawn fails)

Fixes #1519

### DIFF
--- a/src/main/services/ptyIpc.ts
+++ b/src/main/services/ptyIpc.ts
@@ -622,7 +622,12 @@ export function registerPtyIpc(): void {
   // When a direct-spawned CLI exits, spawn a shell so user can continue working
   setOnDirectCliExit(async (id: string, cwd: string) => {
     const wc = owners.get(id);
-    if (!wc) return;
+    if (!wc) {
+      // No owner — clean up since the ptyIpc onExit handler skipped
+      // cleanup for direct spawns.
+      removePtyRecord(id);
+      return;
+    }
 
     try {
       // Spawn a shell in the same terminal
@@ -636,6 +641,11 @@ export function registerPtyIpc(): void {
       if (!proc) {
         log.warn('ptyIpc: Failed to spawn shell after CLI exit', { id });
         killPty(id); // Clean up dead PTY record
+        safeSendToOwner(id, `pty:exit:${id}`, { exitCode: 1, signal: undefined });
+        sendPtyExitGlobal(id);
+        owners.delete(id);
+        listeners.delete(id);
+        removePtyRecord(id);
         return;
       }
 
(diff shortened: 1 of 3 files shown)


arnestrickmann:
Hi @andrassmuk, thanks for this contribution, and especially for the clear root-cause analysis of the two racing onExit handlers. That write-up made the bug easy to understand.

Unfortunately this PR has become outdated - the PTY subsystem has since been substantially refactored, and the files this change touches (src/main/services/ptyIpc.ts and ptyManager.ts) no longer exist — that code now lives under src/main/core/pty/ with a different structure. Since the patch can't be rebased onto the current code, I'm going to close this PR.  If the issue from #1519 still reproduces for you on latest main, we'd genuinely welcome a fresh PR against the new PTY code. Thanks again for taking the time!

PR #1766, opened 2026-04-26
feat(projects): add local project icons

BeelixGit:
## Why

Had a small itch — agent-harness UIs are taking off and none of them (GitHub included) seem to support per-project icons. This adds a local-only, opt-in version: if you want one, you upload an image; if you don't, nothing about emdash changes.

## Summary

Per-project icon shown in the sidebar and project header. Users pick an image in the project Config modal; the main process normalizes it to a 256×256 PNG via Electron's `nativeImage` and stores it under `userData/project-icons/`, indexed by a `project-icons.json` sidecar.

Fully opt-in: projects without a custom icon render exactly as upstream — same `FolderOpen` / `FolderClosed` pair, same click-to-collapse, byte-identical default behavior.

## Screenshots

Sidebar with project icons

Config modal — Project icon section

## Implementation notes

- **No DB schema changes** — icon state is a JSON sidecar alongside the DB.
- **No new dependencies.**
- Renderer transport: base64 data URLs computed in `mapDrizzleProjectRow`, cached by absolute path keyed on mtime so subsequent `getProjects()` calls don't re-encode.
- Source-file open uses `O_NOFOLLOW` (Windows-safe fallback) + buffer-based decode to close a TOCTOU symlink-swap window. Managed-dir check uses `realpathSync`. IPC args validated at the boundary. Index writes are mutex'd + atomic (`.tmp` + `rename`); a failed write rolls back the new PNG.
- Filename stems are sanitized; an 8-char SHA-1 suffix is appended only when sanitization had to alter the raw id, so existing UUID-keyed installs need no migration.
- Project deletion already calls `clearStoredProjectIconForProject`, so icon files don't outlive their rows.

## Behavior

- **Local + remote projects supported.** SSH-attached remote projects can have icons; they're stored on the local install only (the modal helper text says "Local only. Stored in Emdash app data, not committed to the repo.").
- **Accepted formats:** PNG, JPG/JPEG, WEBP. SVG is excluded because Electron's `nativeImage.createFromBuffer` doesn't decode it; users with SVG sources can convert externally.
- **Out of scope:** theme adaptation (images stored as-is), syncing across machines, a built-in glyph picker (potential follow-up).

## Test plan

- [x] `pnpm run format`, `pnpm run lint`, `pnpm run type-check`, `pnpm exec vitest run` all pass — 912/912 tests including 14 new storage tests + 2 renderer tests.
- [x] Upload / replace / clear flows through Config modal verified on macOS.
- [x] Sidebar correctly conditional-renders avatar vs default folder.
- [x] Icons persist across restarts.
- [ ] _(Maintainer)_ Windows / Linux verification welcome.

### DIFF
--- a/src/main/core/projects/controller.ts
+++ b/src/main/core/projects/controller.ts
@@ -1,4 +1,5 @@
 import { createRPCController } from '@shared/ipc/rpc';
+import { clearProjectIcon } from './operations/clearProjectIcon';
 import {
   createLocalProject,
   createSshProject,
@@ -10,6 +11,7 @@ import { getProjectBootstrapStatus } from './operations/getProjectBootstrapStatu
 import { getLocalProjectByPath, getProjects, getSshProjectByPath } from './operations/getProjects';
 import { getProjectSettings } from './operations/getProjectSettings';
 import { openProject } from './operations/openProject';
+import { setProjectIcon } from './operations/setProjectIcon';
 import { updateProjectConnection } from './operations/updateProjectConnection';
 import { updateProjectSettings } from './operations/updateProjectSettings';
 
@@ -27,4 +29,6 @@ export const projectController = createRPCController({
   updateProjectConnection,
   getProjectBootstrapStatus,
   openProject,
+  setProjectIcon,
+  clearProjectIcon,
 });
--- a/src/main/core/projects/icons/storage.test.ts
+++ b/src/main/core/projects/icons/storage.test.ts
@@ -1,0 +1,247 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { MAX_PROJECT_ICON_BYTES } from '@shared/projects';
+import {
+  clearStoredProjectIconForProject,
+  getStoredProjectIconDataUrl,
+  setStoredProjectIconForProject,
+} from './storage';
+
+const getPathMock = vi.fn();
+const createFromBufferMock = vi.fn();
+const cropMock = vi.fn();
+const resizeMock = vi.fn();
+const toPNGMock = vi.fn();
+
+vi.mock('electron', () => ({
+  app: {
+    getPath: (name: string) => getPathMock(name),
+  },
+  nativeImage: {
+    createFromBuffer: (bytes: Buffer) => createFromBufferMock(bytes),
+  },
+}));
+
+describe('projectIconStorage', () => {
+  let tempDir: string;
+  let userDataDir: string;
+  let sourceDir: string;
+  const iconDir = () => path.join(userDataDir, 'project-icons');
+  const indexPath = () => path.join(userDataDir, 'project-icons.json');
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'project-icon-storage-test-'));
+    userDataDir = path.join(tempDir, 'user-data');
+    sourceDir = path.join(tempDir, 'source');
+
+    fs.mkdirSync(sourceDir, { recursive: true });
+    getPathMock.mockReturnValue(userDataDir);
+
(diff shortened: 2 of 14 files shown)


BeelixGit:
Updated for Greptile's findings:

- **Type duplication (`ProjectIconPayload`)** — moved the type to a new
  `src/shared/projectIcons.ts` module and imported it from both
  `electron-api.d.ts` and `global.d.ts` instead of redeclaring locally.

  One note: emdash already has the same duplication pattern for the
  existing `ProjectSettingsPayload` — it's redeclared in both files in
  upstream main. This PR only deduplicates the type I added; the upstream
  one is untouched.

- **Redundant per-row `readIndex()`** — added a sibling mtime-keyed cache
  (`readIndexCached`) for the JSON sidecar, mirroring the existing
  data-URL cache. Writers stay on the uncached path so their R-M-W loop
  never sees a stale snapshot. The first read still hits disk; subsequent
  reads inside the same `getProjects()` are free.

- **Lost collapse-state visual indicator on custom-icon projects** —
  considered a few options (chevron next to the icon, ring around the
  avatar, opacity dim, row left-edge bar). Each one either competed
  visually with the `FolderOpen`/`FolderClosed` swap on no-icon rows or
  introduced a state-cue inconsistent with that existing pattern.
  Opting for no extra indicator on this PR — when you expand a project
  by clicking its icon, the chats appear below; when collapsed they
  don't.

BeelixGit:
Rebased onto current `main` after the v1 release (`7fc7e860`).
The original PR was opened against `main` a few hours before the v1
merge, so it went conflicting almost immediately — this push moves the
single squashed commit (formerly `d1aefd19`) onto the v1 layout.

Functionality and hardening are unchanged from the prior Greptile pass.
Only the file layout shifted to match v1's feature-folder structure:

- Storage moved: `src/main/services/projectIconStorage.ts` → `src/main/core/projects/icons/storage.ts`
- IPC handlers split into operations: `src/main/core/projects/operations/setProjectIcon.ts` and `clearProjectIcon.ts` (registered in `projectController` via `createRPCController`)
- `iconDataUrl` now lives directly on `LocalProject` / `SshProject` in `@shared/projects` (the typed RPC client surfaces it through `rpc.projects.*` automatically — no `electron-api.d.ts` edits, no preload edits, no separate payload type)
- Avatar moved: `src/renderer/components/ProjectAvatar.tsx` → `src/renderer/features/projects/components/project-avatar.tsx`
- Sidebar swap moved: `src/renderer/components/sidebar/LeftSidebar.tsx` → `src/renderer/features/sidebar/project-item.tsx`
- Header avatar in `src/renderer/features/projects/components/project-titlebar.tsx`
- Settings UI ported from the old Config modal to the new Project Settings panel (`src/renderer/features/projects/components/settings-view/project-settings-form.tsx`) with icon staging integrated into the existing dirty/save flow
- Renderer state propagation moved from the React-Query `useProjectManagement` hook to a new `applyProjectIconUpdate(projectId, iconDataUrl)` action on `ProjectManagerStore` (MobX, matching v1's renderer pattern)

Notes on what changed in semantics:

- File picker switched from a dedicated dialog IPC to an HTML `<input type="file">` + the existing `electronAPI.getPathForFile` (already exposed for pty drag-drop). No new IPC, no new dependency.
- `MAX_PROJECT_ICON_BYTES` moved into `@shared/projects` so storage and the form share one constant.
- `ProjectAvatar.test.tsx` dropped (v1 has no `@testing-library/react` and no React-component test paradigm; the 13 storage tests cover behavior).

Validation: `pnpm format`, `lint`, and the project test suite (42/42 in `src/main/core/projects/`) all pass. Pre-existing typecheck/test failures in upstream main (better-sqlite3 native-binding mismatch in `legacy-port` tests, unrelated TS errors in `McpView`/`add-project-modal`/etc.) are unchanged on this branch.

arnestrickmann:
Hi @BeelixGit,
Thank you for this pr and your interest in contributing! 

This hasn't been on the roadmap, therefore closing it for now.

PR #1770, opened 2026-04-26
feat(settings): toggle back from settings view

janburzinski:
go into the settings with cmd+, but now you can also leave them and go back to where you were (also the anon analytics toggle was flickering when quickly switching so i fixed that too)

https://github.com/user-attachments/assets/a85c0e90-34b6-4906-a616-ac7d1b1a73f6


jschwxrz:
closing as already implemented

PR #1776, opened 2026-04-27
fix(ci): code consistency checks

Davidknp:
- upgraded ESLint 
- fixed gh workflow to run lint and use cache 
- fixed major lint and type issues in codebase
- got rid of tests in CI (should be added back later when we clean up tests)


arnestrickmann (commented):
LGTM

PR #1780, opened 2026-04-27
fix: emdash repeatedly requests GitHub ssh key access

jschwxrz:
issue:
- active + polling calls to github triggered ssh key access requests

fix:
- use gh access token authenticated http remotes instead of the ssh remotes for emdash initiated gh calls


Davidknp (approved):
Looks good!

PR #1788, opened 2026-04-27
fix: resolve Windows path resolution and drive scoping issues

JManou:
## Description
A regression introduced in commit `032fdb07` causes `ENOENT` errors and path traversal security blocks on Windows when:
1. A project is located on a different drive than the application's current working directory (e.g., project on `D:`).
2. Absolute paths are used in worktree settings.

## Environment
- OS: Windows

## Technical Details
- **Drive Scoping:** The `rootFs` was previously hardcoded to `/`, which resolves to the current drive root on Windows.
- **Path Joining:** `LocalFileSystem.resolvePath` used `path.join`, which incorrectly prefixes absolute paths starting with a drive letter (e.g., `C:\D:\...\`).

## Steps to Reproduce
1. Open a project located on a non-system drive (e.g., `D:`).
2. Attempt to provision a task.
3. Observe `ENOENT` error.

## The Fix
- **Drive Scoping:** Updated `ProjectManager` and `LocalProjectProvider` to correctly scope the filesystem to the drive root where the worktrees reside.
- **Path Resolution:** Updated `LocalFileSystem.resolvePath` to use `path.resolve()`, which correctly handles absolute paths on Windows (preventing corrupted `C:\D:\` paths).
- **Settings Validation:** Updated `LocalProjectSettingsProvider` to use `defaultLocalWorktreeFs` (native Node `fs`) for validation, ensuring settings can be saved across drive boundaries without 'jail' restrictions.

### DIFF
--- a/src/main/core/fs/impl/local-fs-windows.test.ts
+++ b/src/main/core/fs/impl/local-fs-windows.test.ts
@@ -1,0 +1,48 @@
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+import { LocalFileSystem } from './local-fs';
+
+describe('LocalFileSystem (Windows specific behavior simulation)', () => {
+  it('should correctly resolve absolute Windows paths when they are within the project root', async () => {
+    // This test simulates the case where projectPath is absolute (e.g., C:\project)
+    // and we try to resolve an absolute path (e.g., C:\project\file.txt)
+    // We use path.win32.resolve to simulate the logic since the actual resolve call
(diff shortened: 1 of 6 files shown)


jschwxrz:
Hi, thank you for opening the PR and wanting to contribute!

The rootFs being hardcoded to `/` was an issue, but your fix only addressed part of the problem. Worktree directories on a different drive than the project root would still not be handled correctly on Windows.

I pushed a broader fix in #1799 that separates project file access from absolute host worktree operations, alongside more fixes for windows.

PR #1806, opened 2026-04-28
feat: render images in diff viewer

arnestrickmann:
## Summary
- Render raster images in single-file and stacked diff views instead of showing the binary placeholder.
- Add git-backed image loading for refs and the staged index so before/after panes can show PNG changes.
- Keep working-tree image loading on the existing filesystem image RPC path.

## Root Cause
PNG files were correctly previewable in the normal editor through fs.readImage, but the diff viewer only had text/Monaco model loading for git refs and treated images as binary.

## Validation
- pnpm run format
- pnpm run lint
- pnpm run typecheck
- pnpm run test


arnestrickmann:
will resolve the conflicts

arnestrickmann:
@Davidknp I tightened this so the diff viewer only renders static image formats (png, jpg, jpeg).

Animated formats like gif and webp, plus all other binary files, stay on the existing binary placeholder path and are not loaded/rendered as image diffs. The normal file editor preview behavior is unchanged.

Will resolve the merge conflicts

PR #1807, opened 2026-04-28
feat: auto-rename task from first prompt

### LINKED ISSUE #409: [feat]: Auto-rename workspaces based on first prompt
## Feature Summary

Automatically rename workspaces based on the first prompt(s) submitted, replacing the auto-generated name with something more meaningful and descriptive.

## Problem or Use Case

When creating a new workspace, an auto-generated name is assigned. This makes it difficult for users to find and identify their workspaces later, especially when working on multiple tasks. Users have to manually remember what each auto-generated workspace name corresponds to.

## Proposed Solution

After the user submits their first prompt (or first few prompts), automatically generate a concise, descriptive name based on the content and update the workspace name. This could:

- Analyze the first prompt to extract key topics/intent
- Generate a short, descriptive title (e.g., "Fix auth bug", "Add dark mode", "Refactor API")
- Update the workspace name in the database and UI
- Optionally show a brief notification that the workspace was renamed

## Alternatives Considered

- Manual renaming: Users can already rename workspaces, but this adds friction
- Better auto-generated names at creation: Doesn't help since we don't know the task yet
- Prompt user to name workspace before starting: Adds friction to the quick-start flow

## Additional Context

This is similar to how chat applications (ChatGPT, Claude) auto-title conversations based on the first message. The rename could happen after the first prompt is sent or after the first agent response is received.

junjizhi:
## Summary

Adds a configurable `tasks.autoRenameFromFirstPrompt` setting (default on, alongside the existing `autoGenerateName` toggle) that replaces a freshly auto-generated task slug like `sunny-loops-warn` with one derived from the user's first prompt — addressing #409. Two trigger paths feed a shared helper: `createConversation` covers tasks created with an explicit `initialPrompt`, and a new `tasks.tryAutoRenameFromPrompt` RPC fires from the renderer when `SubmittedInputBuffer` captures the first `isRealTaskInput`-passing message typed into the running TUI; an in-memory dedup set in the helper guarantees a task is renamed at most once per app session even if both paths fire. The rename reuses the existing `generateTaskName` heuristic (no LLM, no network) and the existing `renameTask` RPC, so the underlying git branch follows when the task owns its own branch. A new `taskRenamedChannel` event lets `TaskManagerStore` reactively update the MobX store on server-driven renames so the sidebar label flips without a refresh; this also dropped some dead-code raw-keystroke `onFirstMessage` plumbing in `use-pty.ts` that had no upstream consumer. Tests cover the helper's branch matrix (toggle off, not-first-in-task, empty prompt, dedup, error swallow) and the RPC's conversation-count gating.

## Test plan

- [ ] `pnpm run format && pnpm run lint && pnpm run typecheck && pnpm test`
- [ ] Create a fresh task with no name → confirm slug like `happy-cat-jumps`
- [ ] Type a first prompt in Claude Code → sidebar label flips, git branch follows
- [ ] Send a second prompt in the same task → no further rename
- [ ] Toggle Settings → General → "Auto-rename task from first prompt" off → repeat → name stays
- [ ] Repeat with the create-task-with-initial-prompt flow (createConversation path)
- [ ] Sibling task sharing a branch → DB name updates, branch is not renamed (existing renameTask guard)

Closes #409.

### DIFF
--- a/src/main/core/conversations/autoRenameTaskFromPrompt.test.ts
+++ b/src/main/core/conversations/autoRenameTaskFromPrompt.test.ts
@@ -115,6 +115,28 @@ describe('autoRenameTaskFromPrompt', () => {
     expect(mocks.logWarnMock.mock.calls[0][0]).toMatch(/auto-rename/i);
   });
 
+  it('does not consume the dedup slot when the toggle is off', async () => {
+    mocks.getMock.mockResolvedValueOnce({ autoRenameFromFirstPrompt: false });
+
+    await autoRenameTaskFromPrompt({
+      projectId: 'p1',
+      taskId: 't1',
+      isFirstInTask: true,
+      initialPrompt: 'first prompt',
+    });
+    expect(mocks.renameTaskMock).not.toHaveBeenCalled();
+
+    mocks.getMock.mockResolvedValueOnce({ autoRenameFromFirstPrompt: true });
+
+    await autoRenameTaskFromPrompt({
+      projectId: 'p1',
+      taskId: 't1',
+      isFirstInTask: true,
+      initialPrompt: 'second prompt after toggling on',
+    });
+    expect(mocks.renameTaskMock).toHaveBeenCalledTimes(1);
+  });
+
   it('only renames the same task once per app session', async () => {
     await autoRenameTaskFromPrompt({
       projectId: 'p1',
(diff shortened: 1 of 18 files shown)


arnestrickmann:
Hi @junjizhi, 
thank you for taking this on and for your interest in contributing to Emdash. 

We are closing this for now as we are tackling this internally. 

Thank you again!

PR #1813, opened 2026-04-29
(fix): MCP modal save clone error

arnestrickmann:
## Summary

  Fixes GEN-1065 by keeping modal payloads as plain reference data instead of allowing MobX to deep-observe them.

  ## Root cause

  The MCP edit modal passes server config data through the global modal store. Because the store used
  `makeAutoObservable(this)` with default deep observability, nested modal args could become MobX observable proxies.
  When saving, those proxied MCP server objects were passed through Electron IPC, where structured cloning failed
  with:

  `An object could not be cloned`

  ## Changes

  - Store `activeModalArgs` using `observable.ref` so modal payloads remain plain objects.
  - Add a regression test that verifies MCP-like modal args remain non-observable and structured-cloneable.

  ## Validation

  - `pnpm run format`
  - `pnpm run lint`
  - `pnpm run typecheck`
  - `pnpm run test`
  - Manual UI check: saving an existing MCP server no longer shows the clone error and the modal closes.


arnestrickmann:
@Davidknp can you review before next release pls?

Davidknp (approved):

PR #1822, opened 2026-04-29
fix(conversations): parse configured agent cli arguments

janburzinski:
when using my alias "cc" for the agent claude code and then going to a new conversation (i had cc aliased to claude --dangerously....) i wuold get this erorr "clang: error: unknown argument: '--session-id'
clang: error: unknown argument: '--dangerously-skip-permissions'
clang: error: no such file or directory: 'acfc78e3-acac-4d2a-9a23-fe571efd3acd'
clang: error: no input files
clang: error: unknown argument: '--session-id'
clang: error: unknown argument: '--dangerously-skip-permissions'
clang: error: no such file or directory: 'acfc78e3-acac-4d2a-9a23-fe571efd3acd'
clang: error: no input files
clang: error: unknown argument: '--session-id'
clang: error: unknown argument: '--dangerously-skip-permissions'
clang: error: no such file or directory: 'acfc78e3-acac-4d2a-9a23-fe571efd3acd'
clang: error: no input file"

this fixes that :D

### DIFF
--- a/src/main/core/conversations/impl/agent-command.test.ts
+++ b/src/main/core/conversations/impl/agent-command.test.ts
@@ -61,7 +61,7 @@ describe('buildAgentCommand', () => {
     });
   });
 
-  it('keeps custom Claude command names generic when they do not resolve to the system compiler', async () => {
+  it('keeps the custom cli when resolution fails', async () => {
     getItem.mockResolvedValue({
       cli: 'c --already-configured',
       autoApproveFlag: '--dangerously-skip-permissions',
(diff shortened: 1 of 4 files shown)


Davidknp:
Thanks for raising the issue and the fix suggestion. I don't think this is the right fix though we should not add manual handling and routing of shell commands like this since it will make this more unpredictable and it is not sustainable for upcoming cases. If you've set up 'cc' as an alias for Claude code and its working in your regular terminal but it is not in emdash it is a symptom that something is off about how we resolve the user shell environment or how we invoke the commands.

PR #1823, opened 2026-04-29
fix: surface step on task provision timeout

devinmatte:
## Summary
- Capture the most recent step from `taskProvisionProgressChannel` while a task is provisioning, so timeout errors can say which step was running instead of just "Operation timed out after Nms".
- Add a dedicated `provision-timeout` variant to `CreateTaskError` and render a matter-of-fact, step-aware message in the renderer (e.g. "Task setup timed out after 600s while resolving the worktree.").

## Test plan
- [ ] Create a task on a small repo — provisions normally.
- [ ] Force a timeout (e.g. point at a huge repo / artificially slow `git worktree add`) and confirm the renderer shows the step-aware message.
- [ ] `pnpm run typecheck`, `pnpm run lint`, `pnpm run format` (all clean).

### DIFF
--- a/src/main/core/tasks/operations/createTask.ts
+++ b/src/main/core/tasks/operations/createTask.ts
@@ -31,6 +31,8 @@ function mapProvisionError(error: ProvisionTaskError): CreateTaskError {
         branch: error.branch,
         message: error.message,
       };
+    case 'timeout':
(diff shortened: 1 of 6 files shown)


Davidknp (approved):
Looks great! Thanks a lot!

PR #1827, opened 2026-04-29
fix(agent-hooks): detect pi completion events

janburzinski:
# summary
this fixes that pi is always marked as running

### DIFF
--- a/src/main/core/agent-hooks/hook-config.test.ts
+++ b/src/main/core/agent-hooks/hook-config.test.ts
@@ -9,7 +9,7 @@ vi.mock('@main/core/dependencies/probe', () => ({
   resolveCommandPath: mockResolveCommandPath,
 }));
 
-class MemoryFs implements Pick<FileSystemProvider, 'exists' | 'read' | 'write'> {
+class MemoryFs {
   readonly files = new Map<string, string>();
 
   async exists(path: string): Promise<boolean> {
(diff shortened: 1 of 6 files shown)


src/main/core/agent-hooks/hook-config.test.ts
```diff
@@ -0,0 +1,81 @@
...
+import type { ExecFn } from '@main/core/utils/exec';
+import { HookConfigWriter } from './hook-config';
+
+const mockResolveCommandPath = vi.hoisted(() => vi.fn());
+
+vi.mock('@main/core/dependencies/probe', () => ({
+  resolveCommandPath: mockResolveCommandPath,
+}));
+
+class MemoryFs {
```
Davidknp: MemoryFs is a test helper and should be made reusable and moved to a location that signifies that. I would suggest colocating it with the Fs implementations in main/core/fs and put it into a folder called 'test-helpers'

src/main/core/agent-hooks/hook-config.ts
```diff
@@ -136,3 +159,41 @@ export class HookConfigWriter {
     });
   }
 }
+
+const PI_EMDASH_EXTENSION = `import type { ExtensionAPI } from '@mariozechner/pi-coding-agent';
```
Davidknp: Let's not put ts code into a string literal. Instead we should put it into a separate file and you can then import it as a string using vites import as string functionality using ?raw (https://vite.dev/guide/assets.html#importing-asset-as-string)

Davidknp (changes requested):
Other than that lgtm!

Davidknp (approved):
Nicee, thank you!!

PR #1828, opened 2026-04-29
feat(tasks): add fresh source branch option

janburzinski:
little feature idea: auto pull from base branch (that you select) so that you always have the latest stuff on your worktree 

image


Davidknp:
This should already be possible when you select the remote main branch instead of the local one, no? We already have an issue tracking this to make it more intuitive.

PR #1831, opened 2026-04-30
Fix generic Linux runtime deps for RPM packaging

yuzhichang:
## Summary
- extract shared electron-builder config into a single base factory for stable and canary builds
- inject Linux runtime production dependencies after pack so rpm/deb/appimage builds work with hoisted pnpm node_modules layouts
- add package homepage metadata required by rpm packaging

## Validation
- `mise exec node@24.14.0 -- pnpm run build:main`
- `mise exec node@24.14.0 -- pnpm run typecheck`
- `mise exec node@24.14.0 -- node --experimental-strip-types scripts/release/rebuild-native.ts --arch x64`
- `mise exec node@24.14.0 -- pnpm exec electron-builder --linux rpm --x64 --publish never --config electron-builder.config.ts --config.npmRebuild=false`
- installed rebuilt rpm locally and verified app startup no longer fails with `ERR_MODULE_NOT_FOUND` for `better-sqlite3`

## Notes
- local `pnpm test` is not fully green on current upstream main due to pre-existing baseline issues unrelated to this packaging change, including the `pr-merge-line` expectation mismatch and an existing legacy-port task import test failure

### DIFF
--- a/electron-builder.base.ts
+++ b/electron-builder.base.ts
@@ -1,0 +1,84 @@
+import type { Configuration } from 'electron-builder';
+import { injectLinuxRuntimeNodeModules } from './scripts/release/linux-runtime-node-modules';
+
+interface IdentityConfig {
+  appId: string;
+  artifactPrefix: string;
+  productName: string;
+  r2BaseUrl: string;
+  updateChannel: string;
+  macIcon: string;
+  winIcon: string;
+}
+
+export function createElectronBuilderConfig(identity: IdentityConfig): Configuration {
+  return {
+    appId: identity.appId,
+    productName: identity.productName,
+    directories: { output: 'release' },
+    artifactName: `${identity.artifactPrefix}-\${arch}.\${ext}`,
+    publish: [
+      {
+        provider: 'generic',
+        url: identity.r2BaseUrl,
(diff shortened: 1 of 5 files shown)


yuzhichang:
Addressed the open review threads in `c02d7f27`:

- https://github.com/generalaction/emdash/pull/1831#discussion_r3165533650
  Fixed `copyPackageFiles` to compute the filter relative path from the source path passed by `fs.copy`, which preserves package contents while still excluding nested `node_modules`.

- https://github.com/generalaction/emdash/pull/1831#discussion_r3165533675
  Added an `AfterPackContext` sentinel so Linux runtime dependency injection exits early on repeated `afterPack` invocations in the same packaging run.

- https://github.com/generalaction/emdash/pull/1831#discussion_r3165533705
  Split top-level optional dependencies into a separate guarded loop so missing optional packages no longer abort the hook.

arnestrickmann:
Hi @yuzhichang, thanks for the PR and for following up on the review feedback. Unfortunately the packaging setup has changed quite a bit since April and this branch is now outdated with conflicts, so I'm going to close it. If RPM packaging is still broken for you on latest main, a fresh PR would be very welcome. Thanks again!

PR #1833, opened 2026-04-30
fix: make command in lifecycle scripts optional

jschwxrz:
issue:
- lifecycle setup/run tabs prepared their PTY with command: '', which local spawning resolved to shell -c "".
- that shell exited immediately, so later writes like pnpm I had nowhere to run

fix:
- made lifecycle spawn command optional
- omitted command now explicitly means “prepare an interactive shell"


Davidknp (approved):
Nice!

PR #1840, opened 2026-04-30
feat: add opencode hook plugin

janburzinski:
# summary
adds a local plugin for opencode so that emdash knows when the agent is actually done

issue: GEN-1017


Davidknp (approved):
Nice, thanks!

PR #1850, opened 2026-05-01
fix: clear stale diff models

janburzinski:
when quickly switching between files it could happen that you get stuck in the previous file while wanting to be in the next file (because of the file still loading) 

issue: GEN-1006


Davidknp (approved):
Nice fix, thanks!

PR #1855, opened 2026-05-02
Fix and simplify flake

giodamelio:


giodamelio:
Sorry that was opened accidentally. It was supposed to be a pull against my own forks `main`....

PR #1856, opened 2026-05-03
feat(settings): improve terminal font picker

janburzinski:
# summary
"old" font picker displayed fonts that i didnt have installed and that broke the terminal. this splits into popular and installed fonts and in the popular section only displays fonts you have installed

image


src/renderer/features/settings/components/TerminalSettingsCard.tsx
```diff
@@ -25,17 +44,51 @@ const POPULAR_FONTS = [
...
+  value: '',
+  label: 'Default (Menlo)',
+};
 
 const dedupeAndSort = (fonts: string[]) =>
   Array.from(new Set(fonts.map((font) => font.trim()).filter(Boolean))).sort((a, b) =>
     a.localeCompare(b)
   );
 
+const queryInstalledFonts = async (): Promise<string[]> => {
```
Davidknp: Why are we using 2x strategies here to retrieve the installed fonts?

src/renderer/features/settings/components/TerminalSettingsCard.tsx
```diff
@@ -25,17 +44,51 @@ const POPULAR_FONTS = [
...
+    if (result?.success && Array.isArray(result.fonts)) {
+      return dedupeAndSort(result.fonts);
+    }
+  } catch {
+    // Swallow — UI shows just the default option.
+  }
+  return [];
+};
+
+let installedFontsCache: string[] | null = null;
```
Davidknp: I would prefer we don't use these DIY module level caches. Instead we can just create a hook using tan stack query for data that is rarely mutated and not accessed often (which is the case here)

Davidknp (changes requested):
the combobox implementation looks good 👍

janburzinski:
@greptileai

Davidknp (approved):

Issue #1857, opened 2026-05-03
[feat]: first-class ProxyJump (jump host/bastion) support for SSH connections


Bnjoroge1:
## What i want
I want first-class `ProxyJump` support in emdash ssh connections.

right now a lot of us connect through a bastion/jump host, and this should be a normal supported flow in the app, not something fragile or hidden behind manual ssh config behavior.

## why this matters
- many real remote setups are private-network only and require a jump host
- terminal ssh can do this with `-J`, so emdash should have equivalent support
- without clear support, connection setup feels inconsistent and confusing

## expected product behavior
- in add/edit ssh connection, let me configure jump host directly
- support common forms: host alias, `user@host`, optional port
- use jump host consistently for:
  - test connection
  - saved connection
  - remote project browsing and task/session flows
- show clear debug text when jump host is being used

## nice to have
- allow either:
  - explicit UI fields (preferred), or
  - explicit "use ~/.ssh/config ProxyJump" toggle
- validation/error messages that call out jump host failures vs target-host failures

## quick example
```sshconfig
Host my-target
  HostName 10.0.1.25
  User ubuntu
  ProxyJump jumpuser@bastion.example.com:22
```

if i enter `my-target` in emdash, the app should reliably honor that jump host behavior end-to-end.

PR #1858, opened 2026-05-04
Add ProxyJump support for SSH connections

Bnjoroge1:
Implements first-class ProxyJump support for SSH connections.

- Parses \ from \
- Applies host alias resolution for saved/test SSH connections
- Uses a local \ proxy socket for jump host transport
- Adds regression tests for config parsing and proxy-jump failures

Refs #1857


Bnjoroge1:
Will address Greptile's comments. I was planning on leaving multi-hop impl for a different PR since my main use cases was a single hop but I'll add that.

arnestrickmann:
Hi @Bnjoroge1,
Thanks for opening the PR and for your interest in contributing. 
We'll have a look soon.

Bnjoroge1:
nit: 
1) right now the PR assumes you have a jumphost entry/ref in your ssh config. doesnt allow you to add one in emdash's UI - can try to open a separate PR for that, but I imagine most cases where you have some kind of bastion, it's probably defined in your config file so just wanted to keep things simple, and has been my primary way of using it.

arnestrickmann:
Hi @Bnjoroge1,
Thank you for your interest in contributing and for opening the PR - closing this because we are already working on this internally.

PR #1860, opened 2026-05-04
fix: feature flags and add byoi testing setup

Davidknp:
- fixed feature flags not working in canary build because they have been substituted at build time
- add byoi testing setup: copy the files from testing/byoi into a repo where you want to test out the BYOI feature and it will spin up tasks in a docker container when the setting and the FF is enabled
- fix adding conversation with empty initial prompt


Davidknp:
@greptileai

PR #1868, opened 2026-05-04
fix(linear): keep connection through transient check failures

janburzinski:
# summary
linear connection auth error and timeout was being handled the same this would cause a disconnect just because i lost internet


janburzinski:
@greptile review

janburzinski:
@greptileai

janburzinski:
@greptileai

arnestrickmann:
we already fixed this in a more recent PR, right? @janburzinski - closing this one.

PR #1897, opened 2026-05-06
feat(agent-hooks): improve amp,claude state detection 

janburzinski:
# summary
improves reliablity of the working/error state for amp and claude

amp would now have a plugin that tells emdash its state
claude uses hook events


janburzinski:
@arnestrickmann this is very old but i think ill extract the amp plugin atleast, or?

PR #1913, opened 2026-05-06
feat: add Trae open-in options

arnestrickmann:
## Summary
- add Trae and Trae Solo to the Open In app registry
- add the Trae icon asset
- use Windows-safe raw path quoting for Trae launch commands

## Testing
- Not run: node_modules are missing in this worktree, so pnpm run format fails with prettier not found


arnestrickmann:
Windows ecosystem

PR #1928, opened 2026-05-08
feat: db backed project settings

jschwxrz:
This refactor moves project settings to a DB-backed model while keeping .emdash.json as optional shared config. Without .emdash.json, settings configured in the UI are simply the project settings. When .emdash.json exists, it provides the shared baseline and user-defined DB values take precedence per property as local overrides. The UI now labels those overrides and lets users remove individual overrides to fall back to shared config.

Key changes:
- Added DB-backed project settings with migration
- added legacy import support for data from old .emdash.json files into the new model
- Introduced per-property resolution for shareable settings across DB values and optional .emdash.json
- Added UI override labeling and per-field override removal
- Added sharing flow that writes selected shareable settings into .emdash.json
- Consolidated project settings main-process logic into a dedicated service and settings modules
- Refactored renderer settings state into a single canonical settings page resource
- split the project settings form into smaller sections, a form model, and a dedicated settings hook
- Updated save/share flows to refresh from canonical server state instead of deriving local optimistic state
- Split project creation/inspection logic into local, SSH, and shared helper modules
- Added supporting worktree/file APIs needed for resolving and sharing settings across local and SSH projects


jschwxrz:
@greptileai

Davidknp (approved):

PR #1947, opened 2026-05-09
fix(projects): prefill new project owner

janburzinski:
# summary
i had multiple gh accounts up for selection so when i went to create a new project inside of the modal no accounts were prefilled and i had to select one, this makes it so that the first one is always selected by default


janburzinski:
@greptile

PR #1964, opened 2026-05-11
Fix DeepSec security findings

janburzinski:
## Summary
- Fixes the DeepSec high-severity findings for repo-controlled BYOI command migration and SSH/remote-editor argument injection.
- Hardens Git, SSH, skill filesystem, Electron navigation, release-script, SQLite migration, and Claude trust-file paths against injection, traversal, and stale-write risks.
- Moves GitHub auth tokens out of argv and removes implicit credential fallbacks for BYOI provisioning.

## What Was Done
- Blocked silent migration of `.emdash.json` `workspaceProvider` commands so repo-controlled shell commands are not imported without user intent.
- Added SSH host/username validation and percent-encoded vscode-remote authorities/paths to prevent `ProxyCommand`/URL injection.
- Added allowlists and separators for release CLI flags and Git positional arguments, including clone, branch operations, PR fetch inputs, refs used in show/diff, and GitHub-controller clone flow via shared clone utils.
- Added skill-id validation and strict `SKILLS_ROOT` symlink checks for install/uninstall/sync paths.
- Passed GitHub auth via `GIT_CONFIG_*` environment variables instead of `git -c ...` argv so tokens are not exposed in process listings locally or remotely.
- Required BYOI provision output to explicitly include `username` and either `password` or `useAgent: true`, avoiding silent fallback to local `$USER`/`SSH_AUTH_SOCK`.
- Vacuumed copied app secrets and made the app-secret database migration atomic via a staging path before rename.
- Default-denied non-internal Electron navigations/window opens and removed `file://` from internal URL handling.
- Replaced shell-string env probing with `execFileSync` plus conservative `$SHELL` validation.
- Added optimistic concurrency checks and retries for `.claude.json` trust updates to avoid TOCTOU clobbering.
- Tightened bundled migration lookup to exact migration filenames and restored `foreign_keys` reliably on legacy import failures.

## Why
DeepSec identified multiple places where untrusted repo, renderer, GitHub API, CLI, filesystem, or environment values could be reinterpreted as shell/Git/SSH options, escape expected paths, leak credentials, or leave sensitive deleted bytes recoverable. These changes fail closed at trust boundaries, prefer argv/env-safe APIs, and make migration/update operations atomic where needed.

## DeepSec Coverage
- HIGH: 2/2 fixed
- MEDIUM: 16/16 fixed
- HIGH_BUG: 1/1 fixed
- BUG: 8/9 fixed
- Not fixed: migration SHA-256 verification, because the report itself marks it as not a security vulnerability and this PR intentionally leaves it as-is.

## Validation
- `pnpm run format` passed
- `pnpm run lint` passed
- `pnpm run typecheck` passed
- `pnpm run test` failed: existing Featurebase test mock issue in `src/main/core/featurebase/featurebase-issue-provider.test.ts` where the mocked `./featurebase-connection-service` does not export `NOT_CONFIGURED_ERROR`.

## Commits
- `fix: harden git and ssh arguments`
- `fix: avoid implicit credential exposure`
- `fix: harden filesystem and release paths`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ Thumbs.db
 
 src/main/appConfig.json
 .pi/extensions/emdash-hook.ts
+.opencode/plugins/emdash-notifications.js
--- a/scripts/release/build.ts
+++ b/scripts/release/build.ts
@@ -20,6 +20,9 @@ if (!platform || !['mac', 'linux', 'win'].includes(platform)) {
 }
 
 const archInput = values.arch ?? 'both';
+if (!['x64', 'arm64', 'both'].includes(archInput)) {
+  fail(`Invalid --arch: ${archInput} (expected one of: x64, arm64, both)`);
+}
 const archs: string[] = archInput === 'both' ? ['x64', 'arm64'] : [archInput];
 
 const defaultTargets: Record<string, string> = {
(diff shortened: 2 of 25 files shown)


janburzinski:
lol i just wanted to test something why did it already make a pr for this hahhaha

PR #1977, opened 2026-05-11
fix(ui): increase inactive tab contrast

janburzinski:
image


jschwxrz:
I think we should avoid changing --background-secondary* globally for this? this key is also used outside the tab bars. is intend to only change tab bars or general theme?

arnestrickmann:
Thanks Jan.

I would propose to not change the classes of the title bar:
CleanShot 2026-05-11 at 14 07 58@2x

old: 
CleanShot 2026-05-11 at 14 08 36@2x

arnestrickmann:
What I meant with my previous comment is, that we dont want to change the color of the title bar. 

In the screenshot you can see 
1) unchanged app 
2) this pr branch

Lets make sure that the titlebar's color stays the same

CleanShot 2026-05-13 at 13 46 53@2x

janburzinski:
aaaahhhhh okok mb i misundertsood lol sowy

janburzinski:
image
like this?

PR #1983, opened 2026-05-12
feat: fix codex hook

janburzinski:


src/main/core/conversations/impl/local-conversation.ts
```diff
@@ -142,7 +142,8 @@ export class LocalConversationProvider implements ConversationProvider {
 
     const hookActive = port > 0;
     const provider = getProvider(conversation.providerId);
-    const useHooksOnly = hookActive && provider?.supportsHooks;
+    const shouldKeepClassifierFallback = conversation.providerId === 'codex';
```
Davidknp: Shouldn't we always fallback to classifier when hook doesn't work independent of the provider

janburzinski: agree!! :)

jschwxrz:
could we either keep the emdash hooks in .codex/config.toml or explicitly clean up the old notify key when migrating to .codex/hooks.json? codex supports both inline hooks in config.toml and separate hooks.json files: https://developers.openai.com/codex/hooks


keeping hooks in .codex/config.toml would have two benefits:
- users who already gitignore .codex/config.toml from the previous emdash behavior would not also need to gitignore .codex/hooks.json
- we could replace/remove the old invalid notify key in the same file, preventing the codex startup warning

the tradeoff is that writing .codex/hooks.json is less invasive than mutating config.toml, since it avoids rewriting users existing config. If we keep hooks.json, maybe we should still remove notify from .codex/config.toml when it looks like the emdash managed value, so users do not continue seeing the deprecated key warning after the migration


image

janburzinski:
agreed, makes sense i kept the hooks.json to avoid rewriting users codex.toml and the cleanup 100% makes sense (when the notify contains something about emdash)

janburzinski:
@greptileai

janburzinski:
@greptileai

jschwxrz:
sorry for pushing back again, but I've tested it and have found some blocking issues: 

- hooks need to be reviewed + accepted on a per task level
- the classifier is wired unconditionally, even when hooks are configured -> duplicate events from hooks + classifier
- classifier for codex is not very strong, fires for every codex conversation for me when opening a task that has just been provisioned (even for those that have been sitting unopened in the sidebar for a couple days)

I think we have two options:
- either auto accept the hooks on a per task level
- stop writing to local config and just write to global config once

I think global would make more sense, we should check out if global would work for our case. Also the classifier should only run, if hooks are  unavailable

wdyt @Davidknp 

image

janburzinski:
oh thanks for flagging.

we could add something like this in the global codex config 
```json
{
  "hooks": {
    "Stop": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "curl -sS -X POST \"http://127.0.0.1:$EMDASH_HOOK_PORT/hook\" -H \"Content-Type: application/json\" -H \"X-Emdash-Hook-Token: $EMDASH_HOOK_TOKEN\" -H \"X-Emdash-Event-Type: notification\" --data-binary @-"
          }
        ]
      }
    ],
    "PermissionRequest": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "curl -sS -X POST \"http://127.0.0.1:$EMDASH_HOOK_PORT/hook\" -H \"Content-Type: application/json\" -H \"X-Emdash-Hook-Token: $EMDASH_HOOK_TOKEN\" -H \"X-Emdash-Event-Type: notification\" --data-binary @-"
          }
        ]
      }
    ]
  }
}
```

jschwxrz:
yes, I think that is the right approach. I just put your snippet in ~/.codex/hooks.json and then the hooks only need to be approved once by the user. however, they do not seem to work. I also put the content of the old local config.toml into the global config.toml and that seems to work... though I think we should try to make a global hooks.json work, since they already deprecated notify in local configs, they might deprecate it in the global one as well

jschwxrz:
Have addressed it in #2047

PR #1990, opened 2026-05-12
fix: show tooltip for truncated sidebar task names

janburzinski:
adds
image

fixes
Screenshot 2026-05-12 at 14 14 03


arnestrickmann:
This generally makes a lot of sense - but I also think that adding a tooltip for this is "too much" - after you opened this PR, we added the context menu option to copy the branch name of a task, which is already a "good enough" improvement imo. 

Closing this PR.

PR #2002, opened 2026-05-13
fix: project settings + lifecycle scripts

jschwxrz:


jschwxrz:
@greptileai

src/main/core/projects/settings/project-settings-service.ts
Davidknp, line 1: If we make this a class instead we can use the Hookable interface and also subscribe to project settings services from other services. The renderer events could then be implemented as a 'bridge'. on top the Hookable hooks. `src/main/lib/hookable.ts`

Davidknp (approved):

PR #2019, opened 2026-05-14
fix: duplicate terminal creation on last one close

janburzinski:
# summary
in the temrinal drawer if you closed the last terminal that was open 2 more would be created


jschwxrz (approved):

jschwxrz:
nice!

PR #2036, opened 2026-05-15
feat: add prompt library view

arnestrickmann:
## Summary

This adds a Library entry to the left sidebar and uses it as the shared home for reusable agent assets. The Library opens on a new Prompts tab, with Skills and MCP moved into sibling tabs that reuse their existing views and behavior.

The Prompts tab lets users manage prompt presets that can be inserted from task prompt menus. It includes the built-in review prompt, supports editing/resetting that review prompt, and adds custom prompt presets with create, edit, delete, and search flows. Custom prompts are persisted in app settings so they are available across tasks without introducing a new storage path.

The task bottom bar and create-task modal now expose prompts through a compact Prompts popover. Long prompt bodies are truncated by default, can be expanded inline for review, and the menu scrolls when many presets are available. The trigger intentionally stays compact and does not show a prompt count badge, which keeps the bottom bar stable as users add more presets.

https://github.com/user-attachments/assets/dca88893-99bd-43ff-8f22-2f4e0bf72397

## Implementation Notes

- Registers a new `library` view and prompt modal using the existing renderer registries.
- Adds `promptLibrary` to app settings schema/defaults and updates the settings hook so array settings replace cleanly instead of being object-merged.
- Extends context actions with custom prompt entries, then shares the prompt popover between the task context bar and create-task modal.
- Keeps the existing Skills and MCP implementations intact by rendering them inside the Library tab shell.

## Validation

- `pnpm run format`
- `pnpm run lint` passes with the existing `command-palette-modal.tsx` warning
- `pnpm run typecheck`
- `pnpm rebuild better-sqlite3`
- `pnpm test`


arnestrickmann:
still need to address the greptile comments @jschwxrz

src/renderer/features/library/prompts/prompt-library-view.tsx
```diff
@@ -0,0 +1,256 @@
...
+      title: string;
+      prompt: string;
+      canReset: boolean;
+    }
+  | {
+      kind: 'custom';
+      id: string;
+      title: string;
+      prompt: string;
+    };
```
jschwxrz: I think this type can be simpler. We should just have 
type PromptListItem = {
   id: string;
   title: string;
   prompt: string;
}

and then just seed 1-3 prompts like the review prompt. But after seeding them, they should be treated as regular prompts that the user can edit and delete. Reset functionality is not necessary

src/renderer/features/library/prompts/prompt-library-view.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+export function PromptLibraryView() {
+  const {
+    value: reviewPromptValue,
+    defaults: reviewPromptDefault,
+    update: updateReviewPrompt,
+    reset: resetReviewPrompt,
+    isLoading: isReviewPromptLoading,
+    isSaving: isReviewPromptSaving,
+  } = useAppSettingsKey('reviewPrompt');
```
jschwxrz: those and the other reviewPrompt related things can then also be removed

jschwxrz:
We should also remove the review prompt from the project settings: 
image

jschwxrz:
Also now that the prompt library is its own concept now and not part of the project settings anymore, we should not save the review prompts in the app settings KV table. We should either create a new table prompt_presets or store the presets in the regular KV table.

src/renderer/index.css
```diff
@@ -405,6 +405,20 @@
...
+  @container (max-width: 520px) {
+    .prompt-library-new-prompt-label {
+      display: none;
+    }
+
+    .prompt-library-new-prompt-icon {
+      margin-right: 0;
+    }
+  }
+
```
jschwxrz: those should probably also be just local to the JSX and not global. It's feature specific and not an app wide primitive

arnestrickmann:
@greptile

arnestrickmann:
@greptile review

arnestrickmann:
CleanShot 2026-05-15 at 16 52 00@2x

arnestrickmann:
CleanShot 2026-05-15 at 16 52 55@2x

jschwxrz (approved):
nice!! lgtm

Issue #1936, opened 2026-05-08
[feat]: Make the random branch suffix optional


arthurficialS:
## Feature Summary

A per-project setting in `.emdash.json` to make the random 5-character branch suffix optional, so teams whose convention is one ticket ID → one branch name can use clean names like `ZN-6671` instead of `ZN-6671-aj3jg`.

## Problem or Use Case

Task branches always get a 5-character random suffix appended (e.g., `ZN-6671-aj3jg`). For teams that follow strict ticket-to-branch naming conventions (JIRA, Shortcut, Plane, etc.), this breaks the convention and requires a manual rename before every push.

We name tasks after JIRA ticket IDs (`ZN-6671`) so the branch matches the ticket. Our CI auto-links PRs, runs status checks, and routes deploys based on the ticket-key prefix in the branch name. The current suffix breaks all of these unless we manually rename before each push.

The suffix is hardcoded — there's no setting in `.emdash.json` or the UI:

`src/main/core/tasks/operations/createTask.ts:46`:
```ts
const suffix = Math.random().toString(36).slice(2, 7);
```

`src/main/core/tasks/resolveTaskBranchName.ts`:
```ts
return branchPrefix ? \`\${branchPrefix}/\${rawBranch}-\${suffix}\` : \`\${rawBranch}-\${suffix}\`;
```

The only existing escape is the Linear-linked-issue path (`linkedIssue?.provider === 'linear'`), which doesn't help users on JIRA, Shortcut, Plane, or other trackers.

## Proposed Solution

A per-project setting in `.emdash.json`:

**Option A — boolean, simplest:**
```json
{ "branchSuffix": false }
```

**Option B — mode, preserves the existing uniqueness safety net:**
```json
{ "branchSuffix": "always" | "on-conflict" | "never" }
```

`"on-conflict"` would only append a suffix if the desired branch name already exists locally, so duplicate task names still don't clobber each other but the common case stays clean.

I'd advocate for **Option B with `"always"` as the default** for full backward compatibility.

Touchpoints in code:
- `src/main/core/settings/schema.ts` — new optional field on the project settings schema
- `src/main/core/tasks/resolveTaskBranchName.ts` — consume the new setting
- `src/main/core/tasks/operations/createTask.ts:46` — suffix generation becomes conditional
- `src/main/core/tasks/operations/renameTask.ts` — same logic applies to renames

## Alternatives Considered

Manual rename before push:
```bash
[...]

PR #2037, opened 2026-05-15
feat: make random branch name suffix optional

jschwxrz:
- add toggle to app settings to disable/enable random branch suffix
- branch suffix remains disabled for linear branch names
- make rename task return typed result in case renaming a task fails

Screenshot 2026-05-14 at 20 19 46


jschwxrz:
closes #1936

arthurficialS:
good stuff @jschwxrz thank you for this

PR #2038, opened 2026-05-15
feat: add migration for project configuration

jschwxrz:
summary:
- add stateless project config, best effort migration flow
- detect repo configs from codex, conductor, paseo and superset
- map to emdash concepts, ignore settings for concepts that do not exist in emdash
- let users import into local project settings or write .emdash.json
- show import once on project creation if no .emdash.json exists and at least one supported external config exists
- if user skips import, keep on showing import option in settings as long as no .emdash.json exists, no shareable emdash settings are configured and at least one supported external config exists

limits:
- convenience variables may be used in the configs, but are not mapped

Screenshot 2026-05-14 at 20 44 11

Screenshot 2026-05-14 at 20 43 17


Davidknp (approved):
Nice!!

PR #2045, opened 2026-05-15
fix: prevent duplicate terminal creation

janburzinski:
fixes that cmd+shift+t would create 2 terminals at once


jschwxrz:
@greptileai

jschwxrz (approved):

PR #2052, opened 2026-05-16
fix(pty): direct-spawn Windows .cmd shims when path contains spaces

### LINKED ISSUE #1927: [bug]: Codex provider fails when npm global codex.cmd is under Program Files
### Bug Description

On Windows, the Codex provider fails to start when the npm global `codex.cmd` shim is installed under `%ProgramFiles%\nodejs`.

The same Codex CLI works from the terminal, so this appears related to how Emdash launches Windows `.cmd` shims through its PTY layer.

### Steps to Reproduce

1. Install Node.js on Windows so npm global shims are under `%ProgramFiles%\nodejs`.
2. Install Codex with `npm install -g @openai/codex`.
3. Confirm `codex --version` works in a terminal.
4. Open Emdash and start a Codex conversation.
5. Observe that Codex fails to start.

Actual: Emdash fails to launch Codex and the terminal shows:

```text
'"C:\Program Files\nodejs\codex.CMD"' is not recognized as an internal or external command,
operable program or batch file.
```

On a Chinese Windows locale, this appears as:

```text
'"C:\Program Files\nodejs\codex.CMD"' 不是内部或外部命令，也不是可运行的程序
或批处理文件。
```

Expected: Emdash should start the Codex provider successfully when `codex.cmd` is installed in the normal Node.js global install location on Windows.

### emdash Version

1.1.10

### Operating System

Windows

### Additional Context

Environment:

- Emdash install path: `%LOCALAPPDATA%\Programs\Emdash`
- Node/npm global prefix: `%ProgramFiles%\nodejs`
- Codex CLI path:
  - `%ProgramFiles%\nodejs\codex`
  - `%ProgramFiles%\nodejs\codex.cmd`
  - `%ProgramFiles%\nodejs\codex.ps1`
- Codex version: `codex-cli 0.129.0`

These commands work successfully from the terminal:

```powershell
where.exe codex
```

Output:

```text
C:\Program Files\nodejs\codex
C:\Program Files\nodejs\codex.cmd
```

```powershell
cmd.exe /d /s /c '"C:\Program Files\nodejs\codex.CMD" --version'
```

Output:

```text
codex-cli 0.129.0
```

Directly starting the `.cmd` shim through `node-pty` also works:

```js
const pty = require("node-pty");

const p = pty.spawn(
  "C:\\Program Files\\nodejs\\codex.cmd",
  ["--version"],
  {
    cwd: process.cwd(),
    cols: 80,
    rows: 24,
    env: process.env,
  }
);

p.onData((d) => process.stdout.write(d));
p.onExit((e) => console.log("EXIT", e));
```

Output:

```text
codex-cli 0.129.0
EXIT { exitCode: 0 }
```

However, launching via `cmd.exe` through `node-pty` fails:

```js
const pty = require("node-pty");

const p = pty.spawn(
  "C:\\Windows\\System32\\cmd.exe",
  ["/d", "/s", "/c", "\"C:\\Program Files\\nodejs\\codex.CMD\" --version"],
  {
    cwd: process.cwd(),
    cols: 80,
    rows: 24,
    env: process.env,
  }
);

p.onData((d) => process.stdout.write(d));
p.onExit((e) => console.log("EXIT", e));
```

Observed output:

```text
The syntax of the command is incorrect.
EXIT { exitCode: 1 }
```

The issue may be in the Windows `.cmd` / `.bat` handling path around:

```text
src/main/core/pty/pty-spawn-platform.ts
```

Current behavior appears to resolve extensionless `codex` to a `.CMD` file and then wrap it with:

```text
cmd.exe /d /s /c "<resolved codex.CMD>" ...
```

This seems fragile when the resolved command path is under `Program Files`.

Possible fix directions:

- Avoid wrapping Windows `.cmd` / `.bat` shims through `cmd.exe /s /c` when `node-pty` can launch the `.cmd` file directly.
- Or change the wrapper form to something like `cmd.exe /d /c call "C:\Program Files\nodejs\codex.CMD" ...`.
- Add a regression test for a provider command resolved to `C:\Program Files\nodejs\codex.CMD`.

Workaround:

```text
"C:\Program Files\nodejs\node.exe" "C:\Program Files\nodejs\node_modules\@openai\codex\bin\codex.js"
```

serhiizghama:
## Problem

Fixes #1927.

On Windows, when the npm global `codex.cmd` shim is installed under `%ProgramFiles%\nodejs` (or any path with a space), Codex provider startup fails with:

```
'"C:\Program Files\nodejs\codex.CMD"' is not recognized as an internal or external command,
operable program or batch file.
```

The same `codex` command works fine from a regular terminal.

## Root cause

`resolveWindowsSpawn` wraps `.cmd` / `.bat` shims through `cmd.exe /d /s /c "<quoted path> <args>"`. When the resolved shim path contains whitespace (e.g. `C:\Program Files\nodejs\codex.CMD`), `quoteForCmdExe` adds quotes around it, and the resulting `/c` argument needs to round-trip through node-pty's Windows argv escaping into `cmd.exe`'s own non-standard quote parsing. That round-trip mangles the inner quotes — `cmd.exe` ends up treating `"C:\Program Files\nodejs\codex.CMD"` (with the quotes) as the literal command name and fails to resolve it.

The issue author confirmed empirically that `pty.spawn("C:\\Program Files\\nodejs\\codex.cmd", ["--version"], …)` works directly — node-pty handles `.cmd` / `.bat` invocation on Windows correctly when given the absolute shim path.

## Solution

In `pty-spawn-platform.ts`, when the resolved `.cmd` / `.bat` path contains whitespace, skip the `cmd.exe /d /s /c` wrapper and spawn the shim directly. Paths without whitespace keep the existing wrapping behavior, so nothing changes for `pnpm.cmd`, `npm.cmd`, or shims under standard user-profile install locations (e.g. `C:\Users\<user>\AppData\Roaming\npm`).

This is the surgical fix the issue author suggested ("Avoid wrapping Windows `.cmd` / `.bat` shims through `cmd.exe /s /c` when `node-pty` can launch the `.cmd` file directly") scoped to the path-with-spaces case only.

## Testing

- `pnpm exec vitest run src/main/core/pty/pty-spawn-platform.test.ts` — all 15 tests pass (13 existing + 2 new regression tests for the Program Files case).
- `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` — clean.

Added two regression tests covering both code paths into the `.cmd` branch:
- extensionless `codex` resolved via PATH to `C:\Program Files\nodejs\codex.CMD`
- explicit absolute `C:\Program Files\nodejs\codex.cmd` argv command

Manual Windows verification would still be valuable — I don't have a Windows environment, but the new test mirrors the exact path the issue reporter pasted, and the resulting spawn shape matches their working `pty.spawn(...)` example.

### DIFF
--- a/src/main/core/pty/pty-spawn-platform.test.ts
+++ b/src/main/core/pty/pty-spawn-platform.test.ts
@@ -70,6 +70,55 @@ describe('resolveLocalPtySpawn - Windows', () => {
     });
   });
 
+  it('direct-spawns extensionless commands that resolve to .cmd shims under paths with spaces', () => {
+    // Regression for npm globals installed under `C:\Program Files\nodejs`:
+    // the cmd.exe wrapper used to fail with `'"C:\Program Files\nodejs\codex.CMD"'
+    // is not recognized as an internal or external command`.
+    const result = resolveLocalPtySpawn({
+      platform: 'win32',
+      env: {
+        ...winEnv,
+        Path: 'C:\\Program Files\\nodejs',
+      },
+      fileExists: (candidate) => candidate === 'C:\\Program Files\\nodejs\\codex.CMD',
+      intent: {
+        kind: 'run-command',
+        cwd: 'C:\\repo',
+        command: { kind: 'argv', command: 'codex', args: ['--version'] },
+      },
+    });
+
+    expect(result).toEqual({
+      command: 'C:\\Program Files\\nodejs\\codex.CMD',
+      args: ['--version'],
+      cwd: 'C:\\repo',
+      warnings: [],
+    });
+  });
+
+  it('direct-spawns .cmd argv commands when the absolute command path contains spaces', () => {
+    const result = resolveLocalPtySpawn({
+      platform: 'win32',
+      env: winEnv,
+      intent: {
+        kind: 'run-command',
+        cwd: 'C:\\repo',
+        command: {
+          kind: 'argv',
+          command: 'C:\\Program Files\\nodejs\\codex.cmd',
+          args: ['--version'],
+        },
+      },
+    });
+
+    expect(result).toEqual({
+      command: 'C:\\Program Files\\nodejs\\codex.cmd',
+      args: ['--version'],
+      cwd: 'C:\\repo',
+      warnings: [],
+    });
+  });
+
   it('direct-spawns extensionless commands that resolve to exe files', () => {
     const result = resolveLocalPtySpawn({
       platform: 'win32',
--- a/src/main/core/pty/pty-spawn-platform.ts
+++ b/src/main/core/pty/pty-spawn-platform.ts
@@ -156,6 +156,15 @@ function resolveWindowsSpawn(
   const ext = path.win32.extname(resolvedCommand).toLowerCase();
 
   if (ext === '.cmd' || ext === '.bat') {
+    // When the resolved shim path contains whitespace (e.g. npm globals under
+    // `C:\Program Files\nodejs`), wrapping through `cmd.exe /d /s /c "..."`
+    // breaks because node-pty's argv escaping mangles the inner quotes and
+    // cmd.exe ends up trying to run `'"C:\Program Files\nodejs\codex.CMD"'`
+    // (quotes preserved literally). Spawn the shim directly — node-pty handles
+    // .cmd/.bat invocation on Windows correctly when given the absolute path.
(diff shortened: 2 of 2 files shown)


jschwxrz:
Thanks for the fix! I just merged #2058 though, tackling the same issue, so closing this now.

PR #2070, opened 2026-05-17
feat(sidebar): inline "+" button to add project from Projects header

mezotv:
Adds a `+` button next to the sort filter so you can create a project straight from the sidebar header without scrolling to the footer button. Opinionated tweak from me — totally fine to close if it's not the direction you want.

Preview:

8oAprrb6oAS


jschwxrz:
thanks! decided for a slightly different way to implement it though

PR #2073, opened 2026-05-17
fix: stop Codex loader when turn completes

mezotv:
## Summary
- clear the Codex conversation loader when Codex reports an idle prompt after working
- keep the terminal classifier active for Codex as a fallback when hooks are skipped


janburzinski:
thanks for the update, this then lgtm! :D

janburzinski (approved):

arnestrickmann (approved):

PR #2074, opened 2026-05-17
fix(terminals): focus new terminal from lifecycle drawer

jschwxrz:
previously cmd+shift+t created and activated the new regular terminal in the terminal store, but the drawer kept rendering the lifecycle terminal because its active item lived in local component state. the drawer active item now lives in WorkspaceViewModel so command driven terminal creation can switch the drawer to the newly created terminal


arnestrickmann:
machine

PR #2076, opened 2026-05-17
ui: add issue selector affordances

arnestrickmann:
## Summary
- show a compact chevron on the issue provider selector when multiple trackers are configured
- add an accessible label for the icon-only provider selector
- add a `Link to issue` tooltip to the task header popover trigger

## Validation
- pnpm run format
- pnpm run lint
- pnpm run typecheck
- pnpm run test


arnestrickmann:
CleanShot 2026-05-17 at 15 12 07@2x

PR #2088, opened 2026-05-18
feat: add cursor pointer setting

janburzinski:
very controversial 

https://github.com/user-attachments/assets/2f557ffc-88ad-47aa-8b41-e07188869256


mezotv:
finally!

PR #2091, opened 2026-05-18
fix: open-in paths with spaces on Windows

janburzinski:


janburzinski:
@arnestrickmann could you test pweaseee :D i got 1gb disk left on my mac lol

PR #2095, opened 2026-05-18
fix: render deleted file diffs

janburzinski:
# summary
instead of displaying an empty file we now display the real diff of a completely deleted file 

image


janburzinski:
@greptileai

arnestrickmann (approved):

PR #2099, opened 2026-05-18
fix(updater): add action to update toast

janburzinski:
little improvement with a direect cta 

image


janburzinski:
another very bad demo: https://streamable.com/dxhg4m

janburzinski:
new demo: https://streamable.com/2r685k

janburzinski:
@greptileai

janburzinski:
https://streamable.com/6ory7a

arnestrickmann:
How I think about it: Let’s keep it simple. The user clicks the toast, gets routed to Settings, and updates the app there. I wouldn’t show download progress in the toast, and I’d dismiss it after the redirect.

janburzinski:
also very valid! would def simplify it alot i just thought the feedback was that the user wanted to download through the toast directly (dont 100% remember anymore) but yes imo the simpler/less code the better :D

janburzinski:
image
so basically like this? @arnestrickmann

janburzinski:
https://streamable.com/g00nnc

PR #2104, opened 2026-05-18
fix: shortcuts

jschwxrz:
- standardized shortcut display across the renderer, including keycap styling, tooltip badges, enter glyph handling and macOS modifier ordering
- updated keyboard shortcut settings so reset appears to the left of clear when both actions are available
- fixed the terminal drawer scripts settings button so it opens the project settings tab directly
- changed Mod+T to create a new conversation
- removed the direct theme-toggle global shortcut
- added toggle theme as a searchable command palette action


jschwxrz:
@greptileai

PR #2108, opened 2026-05-19
feat: support external session rename via hook server

OliPetry:
## Summary

Adds two new hook event types — `rename-conversation` and `rename-task` — that let CLI agents rename conversations (tab bar) and tasks (sidebar) through the existing hook server HTTP interface.

When an agent POSTs to `/hook` with `X-Emdash-Event-Type: rename-conversation` (or `rename-task`) and a JSON body `{"title": "..."}`, the handler:
1. Resolves the conversation from the pty ID header
2. Delegates to the existing `renameConversation()` / `renameTask()` service functions
3. Emits IPC events so the renderer updates the UI in real time

**Use case:** Agents managing multiple sessions (orchestrators, task managers) can label conversations and tasks programmatically without needing direct DB access.

## Files Changed

- `src/shared/events/taskEvents.ts` — new `taskRenamedChannel` and `conversationRenamedChannel` IPC events
- `src/main/core/agent-hooks/rename-handler.ts` — new handler that parses the request, looks up the conversation, and delegates to service functions
- `src/main/core/agent-hooks/agent-hook-service.ts` — routes `rename-conversation` and `rename-task` event types before the generic agent event path
- `src/renderer/.../conversation-manager.ts` — listens for `conversationRenamedChannel` and updates conversation title in the store
- `src/renderer/.../task-manager.ts` — listens for `taskRenamedChannel` and updates task name in the store

## Test Plan

- [x] `pnpm run typecheck` passes
- [x] `pnpm run lint` passes

### DIFF
--- a/agents/architecture/main-process.md
+++ b/agents/architecture/main-process.md
@@ -7,7 +7,7 @@ The main process is organized into domain modules under `src/main/core/`. Each d
 ## Domain Modules (`src/main/core/`)
 
 - **account** — Emdash account service, credential store, provider token registry
-- **agent-hooks** — HTTP hook server for agent callbacks, event enrichment, OS notifications, hook config writer (Claude/Codex)
+- **agent-hooks** — HTTP hook server for agent callbacks, event enrichment, OS notifications, hook config writer (Claude/Codex), external session rename (conversation title and task name)
 - **app** — App lifecycle service and controller
 - **conversations** — Conversation CRUD, session start, agent event classifiers (per-provider terminal output parsers)
 - **dependencies** — CLI agent detection, probing, dependency management
@@ -35,7 +35,7 @@ The main process is organized into domain modules under `src/main/core/`. Each d
 - `src/main/lib/` — Logger, telemetry, events, result type, updater error
 - `src/main/db/` — Database schema and initialization
 - `src/main/utils/` — Shell environment, shell escaping, child process env, external links
-- `src/main/core/agent-hooks/` — Hook server, event enrichment, OS notifications, hook config writer for Claude/Codex
+- `src/main/core/agent-hooks/` — Hook server, event enrichment, OS notifications, hook config writer for Claude/Codex, external session rename
 
 ## IPC / RPC Structure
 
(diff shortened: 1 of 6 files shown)


arnestrickmann:
Thanks for the PR, @OliPetry! We’re reviewing it now and will respond here with feedback/questions shortly.

Davidknp:
Looks interesting! We should definitely add a rename functionality like this. I am trying this out right now: Which agents are you using with this? How do you configure the agent to discover and use this endpoint?

OliPetry:
Thanks David! I use this with Claude Code (CLI) sessions — when you run multiple conversations or tasks in parallel, it's useful to label them programmatically so you can tell them apart in the tab bar / sidebar.

For discovery: fair question — right now the agent doesn't discover the endpoint dynamically. I have a small shell script that wraps the `curl` call to `/hook` with the right headers (`X-Emdash-Event-Type: rename-conversation`, `X-Emdash-Pty-Id`, auth token). The agent just calls that script early in a session to set a meaningful title based on what it's working on. I put the script on `PATH` and mention it in the agent's system prompt / instructions so it knows when to use it.

That said, I think discovery is the real question here — not just for rename but for the hook server capabilities in general. My setup works because I wrote the script and told the agent about it, but that doesn't scale to other users. A couple of options that might make sense:

1. **Documentation** — document the available hook event types and their payloads somewhere agents can be pointed to (a AGENTS.md section, a man page, etc.)
2. **Skill / plugin** — package hook server interactions into a bundled skill or MCP plugin that agents can discover and use out of the box, without needing a wrapper script or manual instructions

Option 2 seems like the more compelling long-term answer — agents could discover "I can rename sessions, emit events, etc." without users having to wire anything up.

Oli

Davidknp:
That sounds like a cool setup! The implementation via the rename channel also is aligned with the patterns we are currently using in the app.

As you're pointing out the feature currently requires the user to customise the system prompt and the shell script. Documentation can be a way to mitigate the user effort here however in this case I think this is such a fundamental and important feature that we should try to integrate it more natively without users having to configure it which is why I think Option 2 would be the right call here. 

We are currently looking into leveraging the plugin infrastructure most agents are providing so that users would just install an '[agent] emdash plugin' (or we do it automatically somehow) and wire up the integration globally for the whole machine. The benefit of this would be that we could potentially use more advanced apis and the integration would be more transparent to the user compared to the currently sometimes unexpected modifications to agent settings in the users repo. Arguably this is also the less intrusive option.

For the rename feature specifically we could then, as you're suggesting, register an emdash MCP server that serves a rename endpoint which would let agents rename the conversations at points where the agent determines it makes sense. 

Additionally we have to refactor the agent integrations in general in our codebase as well and centralise agent specific logic into 'agent integrations / plugins'. While reviewing your PR and trying out how we could enable this rename feature I started some experiments on [here](https://github.com/generalaction/emdash/pull/2160). 

I would love to hear your thoughts on this! 

David : )

PR #2113, opened 2026-05-19
fix(editor): render file tree from normalized children

### LINKED ISSUE #1952: [feat]: add collapsible folder tree to file explorer
### Feature Summary

The file explorer currently displays all nested folders and files expanded by default. I’d like the explorer to support a collapsible tree view so folders behave like dropdowns and only reveal contents when expanded.

This would make navigating large projects significantly cleaner and easier.

### Problem or Use Case

When working in projects with deeply nested directories, the current file explorer becomes cluttered because every folder and file is rendered recursively at once.

This creates several issues:

difficult navigation
excessive scrolling
reduced readability
poor scalability for large repositories

For example, generated folders like .svelte-kit/generated/client-optimized/nodes and nested .kiro/skills/* directories are all visible immediately, making the explorer noisy and hard to use.

Image

### Proposed Solution

Implement a recursive collapsible tree-view for the file explorer.

Expected UX
Only root-level files/folders are shown initially
Clicking a folder expands/collapses its contents
Nested folders remain hidden until opened
Expanded state should persist while navigating
Suggested Behavior
Folder rows display expand/collapse indicators (▶ / ▼)
Child nodes are indented based on depth
Recursive rendering supports unlimited nesting
Optional animation for smoother UX

### Alternatives Considered

Current workaround is manually ignoring the visual clutter or relying on search/navigation shortcuts instead of the explorer.

However, this becomes increasingly difficult in larger projects with generated files and deeply nested directories.

### Additional Context

Current behavior renders every nested directory expanded automatically, which makes the file explorer difficult to scan quickly.

A collapsible tree structure similar to:

VS Code
GitHub file explorer
JetBrains IDEs

would greatly improve usability and navigation.

Drakaniia:
File tree QA
Fixes #1952.

## Summary
- Refactor the task editor file tree to use a normalized hierarchical node model with stable POSIX paths and per-folder `children`.
- Keep initial file loading root-only, with directory contents loaded lazily on expansion or reveal.
- Normalize local filesystem relative paths on Windows so nested files do not render as duplicated root-level wrappers.

## Root Cause
The renderer tree was backed by a flat parent index and Windows local file paths could arrive with backslashes. That allowed nested entries such as `src\routes\+page.svelte` to be treated as a single root identity instead of `src -> routes -> +page.svelte`, producing repeated folder/file-looking rows.

## Implementation
- Add `children` to `FileNode` and derive visible rows from `rootNodes` by walking each expanded directory's own children.
- Normalize file-tree paths before storing, loading, revealing, adding, removing, or processing watch events.
- Preserve task-scoped expansion state through existing `expandedPaths` snapshots.
- Update `LocalFileSystem.relPath()` to return POSIX-style relative paths.
- Add focused unit tests for path normalization, collapsed/expanded visibility, directory-first sorting, leaf file behavior, lazy loading, and reveal loading.

## Validation
- `pnpm run format` passed
- `pnpm run lint` passed
- `pnpm run typecheck` passed
- `pnpm exec vitest run --project node src/renderer/features/tasks/editor/stores/files-store.test.ts src/renderer/features/tasks/editor/stores/files-store-utils.test.ts src/main/core/fs/impl/local-fs.test.ts` passed: 61 tests
- Electron rendered QA passed: root-only initially, `src` expands to direct children, `routes` expands to direct children, and no duplicated folder/file wrappers appeared.

## Notes
- `pnpm run test` still has unrelated existing Windows/environment-sensitive failures in migration fixture paths, MCP config path separator expectations, dependency probe mocks, user env mocks, PTY env, and worktree path assertions.

### DIFF
--- a/src/main/core/fs/impl/local-fs.ts
+++ b/src/main/core/fs/impl/local-fs.ts
@@ -190,7 +190,7 @@ export class LocalFileSystem implements FileSystemProvider {
    * Get relative path from absolute path
    */
   private relPath(fullPath: string): string {
-    return relative(this.projectPath, fullPath);
+    return relative(this.projectPath, fullPath).replace(/\\/g, '/');
   }
 
   /**
--- a/src/renderer/features/tasks/editor/editor-file-tree.tsx
+++ b/src/renderer/features/tasks/editor/editor-file-tree.tsx
@@ -346,9 +346,7 @@ export const EditorFileTree = observer(function EditorFileTree() {
   const editorView = taskView.editorView;
   const [isDragOverRoot, setIsDragOverRoot] = useState(false);
 
-  const visibleRows = files
-    ? buildVisibleRows(files.nodes, files.childIndex, editorView.expandedPaths)
-    : [];
+  const visibleRows = files ? buildVisibleRows(files.rootNodes, editorView.expandedPaths) : [];
 
   const parentRef = useRef<HTMLDivElement>(null);
 
--- a/src/renderer/features/tasks/editor/stores/files-store-utils.test.ts
+++ b/src/renderer/features/tasks/editor/stores/files-store-utils.test.ts
@@ -1,0 +1,68 @@
+import { describe, expect, it } from 'vitest';
+import { buildVisibleRows, makeNode, sortFileNodes } from './files-store-utils';
+
+describe('file tree utils', () => {
+  it('normalizes Windows paths into stable POSIX node identities', () => {
+    const node = makeNode('src\\components\\Button.tsx', 'file');
+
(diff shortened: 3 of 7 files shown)


jschwxrz:
thank you for the PR!!

I pushed a follow up commit that addresses the issues greptile flagged and switches the file tree back to eager background loading.

eagerly warming the full tree does cost more upfront, but in practice that work is usually hidden during task startup because users rarely open the file viewer in the first second or two. pure lazy loading made expansion noticeably slower on remote workspaces, which is a much more visible UX regression

jschwxrz (approved):

PR #2126, opened 2026-05-19
fix: allow empty branch prefix

janburzinski:
# summary
- adds the ability to have no prefix


arnestrickmann:
Thanks!

I am not running into this issue when leaving the branch prefix field empty. 

Instead, I will add a hint into the settings, saying that the user can also leave it empty. 
CleanShot 2026-05-19 at 15 12 45@2x

PR #2138, opened 2026-05-20
fix: run resource monitor only while open

janburzinski:
# summary
- sample only when open
- clean up listeners
- prevent older snapshots from overwriting newer data
- track open monitor from renderer


janburzinski:
@greptileai

arnestrickmann (approved):
great improvement

PR #2149, opened 2026-05-20
fix: enable opencode auto approve

janburzinski:
# summary
before auto approve did nothing for opencode

https://opencode.ai/docs/permissions/


arnestrickmann (approved):
thanks for the fix!

PR #2157, opened 2026-05-21
fix: skip letta version probe

janburzinski:
# summary
- dont do "letta --version" it will ALWAYS create a .lettaignore nomatter what; checked the "which letta" should hopefully be enough for a lively check


arnestrickmann (approved):
thanks! - will stop creating the .letta files automatically on version check @janburzinski

PR #2160, opened 2026-05-21
feat: improve agent handling

Davidknp:


OliPetry:
Thanks David! I took a look at #2160 (with some help from my agent doing the code analysis 😄). The plugin architecture looks like the right direction — consolidating all the per-provider logic into self-contained plugins with scoped I/O is clean, and it's a natural home for capabilities like rename.

Happy to have the rename PR folded into that refactor rather than merged separately. It was always more of a proof-of-concept for the capability than a final form.

Looking forward to seeing how the plugin system evolves — especially the MCP server angle you mentioned. That feels like it could unlock a lot for agent-emdash integration without users having to configure anything.

PR #2170, opened 2026-05-21
fix(plain): restore thread context injection, show all threads, and prefix task names with ref

tobi110289:
## Summary

Three regressions vs. the original Plain integration ([#1328](https://github.com/generalaction/emdash/pull/1328)):

1. **`listIssues` was hard-filtered to `statuses: ['TODO']`** — snoozed and done threads were unreachable from the issue selector. The original PR had status filter pills (All/Todo/Snoozed/Done) that the unified issue-selector flow does not surface. Since the selector is now a single list, the only reasonable fix is to show all statuses.
2. **Thread context never reached the agent.** The original PR's `taskCreationService` auto-prefixed the initial prompt with a seed context (customer, priority, labels, description). The unified create-task flow stopped doing this — a user typing `/triage` and selecting a thread ended up sending only `/triage`.
3. **Task names lost the thread-ref prefix.** The original generated names like `t-1070-fix-login`; the unified flow produced LLM-only names with no ref.

## Changes

- **`src/main/core/plain/plain-issue-provider.ts`**
  - `listIssues` now passes `statuses: ['TODO', 'SNOOZED', 'DONE']` instead of just `TODO`.
  - Implements `getIssueContext` (mirrors the Linear pattern): fetches the thread by ref/id, awaits the customer, and packs customer name/email, priority (Urgent/High/Normal/Low), labels, and the full description into `issue.context`. `Title:` and `Status:` are intentionally omitted because `buildIssueContextText` already emits them from the top-level Issue fields; the description is only appended when it differs from the preview text already on `Issue.description`.
  - Customer formatting branches on which fields are present so an email-only customer renders as just the email rather than the malformed `Customer: user@example.com>` (no opening `<`).
  - `toIssue` synthesizes a branch-style `branchName` (`${ref}-${title}`) so the existing `getIssueTaskName` can produce `t-1070-fix-login-bug`. This field is never used as the actual git branch — `resolveTaskBranchName` only honors Linear's branchName.

- **`src/renderer/features/tasks/issue-context/refresh-linked-issue-context.ts`**
  - Generalized from a hardcoded `'linear'` check to a `PROVIDERS_WITH_CONTEXT` set (`linear`, `plain`). This file was dead code before this PR — it had no callers.

- **`src/renderer/features/tasks/create-task-modal/use-from-issue-mode.ts`**
  - When a user selects an issue, refresh its context once and replace local state with the enriched issue. Cancels gracefully on rapid selection changes.

- **`src/renderer/features/tasks/create-task-modal/create-task-modal.tsx`**
  - In `handleCreateTask`'s from-issue branch, auto-append the rendered context text to `initialPrompt` unless the user already typed the identifier. The existing `ModalContextBar` chip still works for manual insertion; this just guarantees the data reaches Claude even without clicking it. The change applies to any provider that exposes issue context.

- **`src/renderer/features/tasks/create-task-modal/issue-task-name.ts`**
  - Generalized the provider guard so Plain participates in the same "use branchName as task name" path as Linear. Tests added covering the Plain ref-prefix case and the no-branch-name fallback.

## Test plan

- [x] `pnpm run format`
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm run test` (825 passed)
- [x] Manually verified: snoozed and done Plain threads now appear in the issue selector
- [x] Manually verified: typing `/triage` and selecting a Plain thread sends both the slash command and the full thread context to the agent
- [x] Manually verified: Plain-linked task names are pre-filled as `t-XXXX-{slugified title}`


arnestrickmann (commented):
Updated this PR by merging the latest `main` and resolving the create-task modal conflict.

arnestrickmann (approved):

PR #2173, opened 2026-05-21
fix(renderer): enable Ctrl+V paste in terminal on Windows

abierhaus:
# summary
- Windows terminals now accept `Ctrl+V` (and still `Ctrl+Shift+V`) to paste from the clipboard.
- Linux behaviour is unchanged (`Ctrl+Shift+V` only); macOS continues to rely on the native `Cmd+V` handling.
- Updates the `shouldPasteToTerminal` tests to cover all three platforms.

# why
On Windows, pressing `Ctrl+V` inside the PTY pane previously dropped through to xterm.js, which sent the raw `0x16` control character to the running CLI agent instead of pasting clipboard text. The comment above `shouldPasteToTerminal` already documented the intended Windows behaviour, but the implementation only intercepted `Ctrl+Shift+V` on non-Mac platforms.

### DIFF
--- a/src/renderer/lib/pty/pty-keybindings.ts
+++ b/src/renderer/lib/pty/pty-keybindings.ts
@@ -76,24 +76,33 @@ export function shouldKillLineFromTerminal(event: KeyEventLike, isMacPlatform: b
 }
 
 /**
- * Detect Ctrl+Shift+V paste shortcut on Linux.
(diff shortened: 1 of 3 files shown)


arnestrickmann (approved):
lgtm

arnestrickmann (approved):

PR #2180, opened 2026-05-22
fix: paste into Claude on Windows; route clipboard paste through provider-aware injection

### LINKED ISSUE #1901: [bug]: cannot paste text into Claude
### Bug Description

In a Claude chat, pressing `Ctrl + V` does not paste text. Need to use `Shift + Insert`.

### Steps to Reproduce

1. Create a project
2. Start a new task
3. Start conversation with Claude Code
4. Copy some text
5. Try to paste it with `Ctrl + V`

Actual: Need to use `Shift + Insert`
Expected: `Ctrl + V` just works

### emdash Version

1.1.10

### Operating System

Windows 11

### Additional Context

_No response_

Aarekaz:
## Summary

Fixes #1901 — pasting into Claude does not work on Windows.

There were two overlapping bugs in the terminal paste pipeline:

1. **Ctrl+V on Windows did nothing.** The renderer only recognized `Ctrl+Shift+V` (Linux convention) as a terminal paste shortcut. On Windows, users naturally press `Ctrl+V`, which fell through to xterm.js's default and never reached our paste handler.

2. **Multiline pastes broke for non-Claude providers.** `pasteFromClipboard` wrote raw text to the PTY (`sendInput(text)`) with no provider-aware formatting. For agents that expect bracketed paste, every newline was interpreted as Enter — so the first line submitted the prompt before the rest of the text arrived. This is the "Ctrl+Shift+V seems broken at times" symptom in the issue thread.

The codebase already had a provider-aware paste function (`pastePromptInjection` in `src/shared/prompt-injection.ts`) that correctly sends raw text to Claude and bracketed paste to others. It was wired into drag-and-drop and the context bar, but **not** into the clipboard-paste shortcut.

This PR:

- Routes `pasteFromClipboard` through `pastePromptInjection` so the keyboard shortcut and the `appPasteChannel` listener share the same provider-aware formatting.
- Threads `providerId` from the conversation through `PtyPane` → `usePty` (via a ref so swapping agents mid-session does not remount xterm.js).
- Adds `Ctrl+V` as a terminal paste shortcut on Windows. Linux keeps `Ctrl+Shift+V` (so `Ctrl+V` remains readline `quoted-insert`); macOS keeps its `Cmd+V` menu accelerator.
- Updates the keybinding tests with a per-platform matrix.

## Files changed

- `src/renderer/lib/pty/pty-keybindings.ts` — `shouldPasteToTerminal` now takes `isWindowsPlatform`.
- `src/renderer/lib/pty/use-pty.ts` — accepts `providerId`, paste callback uses `pastePromptInjection`, adds `IS_WINDOWS_PLATFORM`.
- `src/renderer/lib/pty/pty-pane.tsx` — forwards `providerId` to `usePty`.
- `src/renderer/features/tasks/conversations/conversations-panel.tsx` — passes `activeConversation?.data.providerId`. Plain shell terminals leave it undefined, so they keep the existing bracketed-paste-for-multiline behavior, which is correct for a shell.
- `src/renderer/tests/terminalKeybindings.test.ts` — replaces the Linux-only test with a Mac/Windows/Linux matrix.

## Test plan

- [x] `pnpm run format`
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm test` (823/823 passed, including new platform-matrix cases)
- [ ] **Manual verification on Windows 11**: Ctrl+V pastes single- and multi-line text into Claude.
- [ ] Manual verification on Linux: Ctrl+Shift+V still pastes; bare Ctrl+V still reaches the PTY as `quoted-insert`.
- [ ] Manual verification on macOS: Cmd+V still pastes via the Edit menu; nothing regresses.

The logic and unit tests pass on macOS; I was not able to verify the end-to-end Windows flow locally.

Closes #1901

### DIFF
--- a/src/main/core/conversations/impl/keystroke-injection.ts
+++ b/src/main/core/conversations/impl/keystroke-injection.ts
@@ -24,6 +24,7 @@ export function scheduleInitialPromptInjection(args: {
   const payload = buildPromptInjectionPayload({
     providerId: args.conversation.providerId,
     text: args.initialPrompt,
+    mode: 'initial-prompt',
   });
 
   let injected = false;
--- a/src/renderer/features/tasks/conversations/conversations-panel.tsx
+++ b/src/renderer/features/tasks/conversations/conversations-panel.tsx
@@ -118,6 +118,7 @@ export const ConversationsPanel = observer(function ConversationsPanel() {
                     className="h-full w-full"
                     onInterruptPress={onInterruptPress}
                     mapShiftEnterToCtrlJ
+                    providerId={activeConversation?.data.providerId}
                     remoteConnectionId={remoteConnectionId}
                   />
                 </div>
--- a/src/renderer/lib/pty/prompt-injection.ts
+++ b/src/renderer/lib/pty/prompt-injection.ts
@@ -14,6 +14,7 @@ export async function pastePromptInjection(args: InjectPromptArgs): Promise<void
   const payload = buildPromptInjectionPayload({
     providerId: args.providerId,
     text: args.text,
+    mode: 'paste',
     forceBracketedPaste: args.forceBracketedPaste,
   });
   if (!payload) return;
--- a/src/renderer/lib/pty/pty-keybindings.ts
+++ b/src/renderer/lib/pty/pty-keybindings.ts
@@ -76,11 +76,20 @@ export function shouldKillLineFromTerminal(event: KeyEventLike, isMacPlatform: b
 }
 
 /**
- * Detect Ctrl+Shift+V paste shortcut on Linux.
- * Linux terminals use Ctrl+Shift+V as the standard paste shortcut,
- * unlike Windows/macOS which use Ctrl+V/Cmd+V.
+ * Detect clipboard-paste shortcuts in the terminal pane.
+ *
+ * - Ctrl+Shift+V on Linux/Windows: standard terminal paste shortcut.
+ * - Ctrl+V on Windows: native Windows paste shortcut. Intercepted here so
+ *   paste flows through our provider-aware pipeline instead of xterm.js's
+ *   built-in (bracketed) paste, which some CLI agents (e.g. Claude Code)
+ *   do not handle correctly.
+ * - Cmd+V on macOS: routed via the Electron Edit menu, not handled here.
  */
-export function shouldPasteToTerminal(event: KeyEventLike, isMacPlatform: boolean): boolean {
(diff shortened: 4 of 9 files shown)


Aarekaz:
Thanks for the catch on the trim regression. Fixed in 7d50ae27:

- Dropped the unconditional `text.trim()` in `buildPromptInjectionPayload` — verified no caller actually depended on it (drag-drop joins single-quoted paths with spaces, no leading/trailing whitespace; the prompt-injection path didn't need it either).
- Changed the bracketed-paste heuristic from "any newline = multiline" to "internal newlines = multiline." A single trailing newline now means "execute this line" and passes through raw, so `ls -la\n` pasted into a shell auto-executes again.
- Added two regression tests covering both cases (trailing-newline auto-execute and leading-whitespace preservation on multiline pastes).

The Python REPL / indented multiline case is still bracket-wrapped on non-Claude providers, which is the correct path — bracketed paste preserves the leading whitespace as bytes inside the wrap sequence.

Aarekaz:
Following up on issue #1901's full comment thread — Robin Mennens reported a second case I hadn't fully addressed: multi-paragraph pastes (Lorem Ipsum example) breaking when pasted into Claude via Ctrl+Shift+V, while Shift+Insert worked.

Root cause: the \`providerId !== 'claude'\` exemption in \`buildPromptInjectionPayload\` was added by 20610a16 ("fix(opencode): submit initial prompt reliably") for keystroke injection of the *initial task prompt* at session start, where Claude's TUI apparently mishandled bracketed paste. Routing user-driven clipboard paste through the same helper inherited that exemption — but Robin's evidence (Shift+Insert works, which routes through xterm.js's bracketed paste) shows Claude does handle bracketed paste fine for ordinary clipboard pastes.

Fixed in b41adeed by adding a \`PromptInjectionMode = 'paste' | 'initial-prompt'\` discriminator. The Claude exemption now applies only to \`mode: 'initial-prompt'\`. The renderer's \`pastePromptInjection\` wrapper hardcodes \`mode: 'paste'\` so all three of its callers (clipboard, drag-drop, context bar) route Claude multiline content through bracketed paste.

Added a regression test that mirrors the Lorem Ipsum scenario from the issue.

Issue thread coverage:
- ✅ Ctrl+V on Windows
- ✅ Multiline pastes into non-Claude providers
- ✅ Multiline pastes into Claude (this commit)
- 🛑 In-app paste hint / docs (separate UX work, out of scope)

PR #2187, opened 2026-05-22
fix: Fix environment variable configuration for opencode MCPs

DMarby:
This PR fixes environment variable handling when reading/writing opencode MCP configuration: opencode expects `environment`, but we were setting `env`. Additionally, it doesn't support setting either for remote MCPs. See McpLocalConfig and McpRemoteConfig on https://opencode.ai/config.json for reference


arnestrickmann (approved):
lgtm

arnestrickmann:
Thank you @DMarby for your interest in contributing to Emdash. 
Great fix - thanks for catching this as well.

PR #2188, opened 2026-05-22
fix: stale awaiting input indicator on tab close

janburzinski:
# summary
- conversation tab would still be marked as unseed even after you closed it


arnestrickmann (approved):
good catch, thank you

PR #2189, opened 2026-05-22
feat: add skills.sh search backend

janburzinski:
# summary
- adds support for skills.sh in skills
- Parse multiline skill descriptions so YAML block markers do not appear in the skills catalog. (@mezotv )

### DIFF
--- a/src/main/core/skills/SkillsService.test.ts
+++ b/src/main/core/skills/SkillsService.test.ts
@@ -1,0 +1,94 @@
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { afterEach, describe, expect, it } from 'vitest';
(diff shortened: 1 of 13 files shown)


arnestrickmann:
I think this still needs a reliability pass before merging: the implementation mixes Skills.SH remote IDs with Emdash local skill names. Some live Skills.SH results contain IDs like `react:components`, and some generated install names exceed the existing 64-char `isValidSkillName` limit. Installed Skills.SH results can also still show as installable because the UI filters by `catalogSkillId` while installs are stored as `skillssh-owner-repo-skill`.

janburzinski:
@greptileai

janburzinski:
@arnestrickmann would love if you could look over this again :)

arnestrickmann (approved):
Approved. Thank you for also adding the focused regression tests for the risky Skills.SH install paths.

PR #2198, opened 2026-05-22
fix(main): stop hidden conversation sessions

janburzinski:
# summary 
- hidden conversations now get stopped after grac eperiod (30s)
- track visible conversations
- prevents runaway background agents


janburzinski:
@greptileai

janburzinski:
@greptileai

PR #2199, opened 2026-05-23
fix(renderer): sanitize persisted snapshots

janburzinski:
# summary
- workspace creation failed on existing folders
- git failed before provisioning finished


janburzinski:
@greptileai

PR #2200, opened 2026-05-23
feat(ssh): prefill default private key

janburzinski:
# summary
- autofills the default ssh key when connecting to a remote


janburzinski:
@greptileai

PR #2203, opened 2026-05-23
fix: enter in cmdk doing the wrong action

janburzinski:
# summary
when in the cmdk and "beneath" the cmdk is a different keybind selected that would get used, but that doesnt maek sense since you're in the cmdk 

this fixes this bug:
https://github.com/user-attachments/assets/91d9a231-c6af-4594-87c7-70f4806c2a3a


arnestrickmann (approved):
thank you jan, and good catch

PR #2204, opened 2026-05-23
fix(github): route Octokit by host to support GitHub Enterprise

### LINKED ISSUE #2181: [bug]: PR creation fails for Github Enterprise Instance
### Bug Description

Hi team,

it seems to me that there are still issues with Github Enterprise instance. I want to create a Draft PR but it gives me following error message. `gh auth` is properly setup and authenticated. Also, it did work with 0.* version.

Image

Thanks,
Felix

### Steps to Reproduce

1. Run `gh auth` with enterprise instance.
2. Create PR --> error message (it also does not find existing PRs, and Project settings also seems to be broken).

Github Enterprise via `gh`cli is supported.

### emdash Version

1.1.24

### Operating System

Mac

### Additional Context

_No response_

Aarekaz:
Closes #2181.

## Summary

- Routes Octokit lookups by host so Enterprise URLs hit `https://{host}/api/v3` instead of always hitting public `api.github.com`
- Generalizes the URL parser to capture host (was hardcoded to `github.com`)
- Propagates host through `pr-sync-engine`, `issue-service`, and `gh-cli-token` so PR sync, PR creation, and merge all use the right API
- Preserves existing behavior for `github.com` users — the parser still defaults the shorthand `owner/repo` to `github.com`, and the create-PR target picker is still `github.com`-only (broadening it to known Enterprise hosts is a follow-up)

## Root cause

Two hardcodes:

1. `Octokit` was always constructed without a `baseUrl`, so every API call hit `api.github.com` regardless of the actual repo host
2. The URL parser regexes only matched `github.com`, so GHE remotes parsed to `null` — surfacing as "invalid GitHub repository"

That's why the reporter sees PR creation fail _and_ existing PRs not getting found _and_ Project settings appear broken: they all funnel through the same parser + Octokit.

## Design notes

- The parser is intentionally permissive about host. Enterprise instances can live on any domain, so there's no reliable URL-only way to tell "this is a GitHub-compatible server." Verifying that a host actually speaks the GitHub API moves to the call layer (auth failure becomes a clearer error than silent rejection).
- For callers that _do_ want a strict `github.com`-only check (currently the create-PR target picker), a new `isGitHubDotCom(ref)` helper is exported. Used in `target-remote.ts` to preserve previous behavior.
- `getOctokit(host)` caches per `(host, token)` so multiple Enterprise hosts coexist cleanly and token rotation only invalidates the affected entry.
- For non-`github.com` hosts, `getToken(host)` reads directly from `gh auth token --hostname <host>` because emdash's encrypted secure storage only holds one token under a single key. This means GHE users rely on `gh auth login --hostname …` being configured, which is the standard GHE workflow.

## Test plan

- [x] `pnpm run format`
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm test` — 834 passing
- [x] New parser tests cover github.com SSH/HTTPS, GHE SSH/HTTPS, ports, case normalization, shorthand
- [x] New `pr-sync-engine.host-routing.test.ts` proves the engine routes a GHE URL to `getOctokit('ghe.example.com')` and a github.com URL to `getOctokit('github.com')`
- [x] Manual: `github.com` PR creation still works locally

## What I couldn't validate

I don't have access to a GitHub Enterprise instance, so the end-to-end "create a PR against GHE" path is not directly verified. The unit + integration tests cover the routing logic, and the `github.com` path is fully validated, but the actual `api/v3` round-trip against a real GHE 3.x server would benefit from a second pair of eyes.

@bondfelix — if you have a moment, would you mind pulling this branch and confirming PR creation works on your GHE instance? `gh pr checkout` from this PR should pick up the branch.

## Out of scope (intentional follow-ups)

- Broadening the create-PR target picker to include Enterprise remotes (currently filters to `github.com` to keep this PR minimal)
- A GHE host setting in the UI (auto-detection from the repo remote should be enough for the bug fix)
- Multi-host repo browsing in the "Add Project" picker

### DIFF
--- a/src/main/core/github/github-issue-provider.test.ts
+++ b/src/main/core/github/github-issue-provider.test.ts
@@ -32,6 +32,7 @@ describe('githubIssueProvider', () => {
 
     expect(mockIssueService.listIssues).toHaveBeenCalledWith(
       {
+        host: 'github.com',
         owner: 'owner',
         repo: 'repo',
         nameWithOwner: 'owner/repo',
@@ -52,6 +53,7 @@ describe('githubIssueProvider', () => {
 
     expect(mockIssueService.searchIssues).toHaveBeenCalledWith(
       {
+        host: 'github.com',
         owner: 'owner',
         repo: 'repo',
         nameWithOwner: 'owner/repo',
--- a/src/main/core/github/services/gh-cli-token.ts
+++ b/src/main/core/github/services/gh-cli-token.ts
@@ -1,17 +1,28 @@
 import type { IExecutionContext } from '@main/core/execution-context/types';
+import { GITHUB_DOT_COM_HOST } from '@shared/github-repository';
 
-export async function isGhCliAuthenticated(ctx: IExecutionContext): Promise<boolean> {
+function hostnameArgs(host: string): string[] {
+  return host === GITHUB_DOT_COM_HOST ? [] : ['--hostname', host];
+}
+
+export async function isGhCliAuthenticated(
+  ctx: IExecutionContext,
+  host: string = GITHUB_DOT_COM_HOST
+): Promise<boolean> {
   try {
-    await ctx.exec('gh', ['auth', 'status']);
+    await ctx.exec('gh', ['auth', 'status', ...hostnameArgs(host)]);
     return true;
   } catch {
     return false;
   }
 }
 
-export async function extractGhCliToken(ctx: IExecutionContext): Promise<string | null> {
+export async function extractGhCliToken(
+  ctx: IExecutionContext,
+  host: string = GITHUB_DOT_COM_HOST
+): Promise<string | null> {
   try {
-    const { stdout } = await ctx.exec('gh', ['auth', 'token']);
+    const { stdout } = await ctx.exec('gh', ['auth', 'token', ...hostnameArgs(host)]);
     const token = stdout.trim();
     return token || null;
   } catch {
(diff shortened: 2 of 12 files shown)


rabanspiegel:
Thanks for your contribution, @Aarekaz ! We'll have a look 🙏

Aarekaz:
Welcome! Let me know if I am doing anything wrong and how I can improve!!

Aarekaz:
Pushed `81a81f78` to address both findings:

1. Token subprocess on every `getOctokit` call. Added a per-host TTL cache in the connection service mirroring the existing one for github.com. The subprocess now only fires on a true cache miss or after the 5-minute TTL expires.

2. `www.` strip being too broad. The regex now captures the host verbatim, and only `www.github.com` is collapsed to `github.com` (since that's its actual redirect). Added a test for `https://www.ghe.example.com` to lock the behavior in.

arnestrickmann:
Thanks a lot for the contribution and for iterating on the feedback so quickly, @Aarekaz really appreciated 🙏

We discussed this internally and decided to implement the fix internally, so we’re going to close this PR for now. That said, your implementation and the direction you took were genuinely helpful inspiration for solving this issue.

Really appreciate you taking the time to work on this and engage with the review comments so thoughtfully. Hope to see more contributions from you in the future!

Aarekaz:
It makes complete sense. Glad I was able to give some value.

PR #2213, opened 2026-05-24
feat(diff-view): add tree view for changes and pull request files

shoto290:
## Summary

- Add a button to switch between a flat list and a tree view of the changes.
- Available in the Changed, Staged, and Pull Request files panels.
- Remember the preference globally across all tasks and sessions.
- Unify the section header used across the changes panel for consistency.


jschwxrz:
nice, thank you for the PR! lgtm, just one issue I found:

if a file is replaced by a directory e.g. deleted "foo" plus added "foo/bar.ts" the changed count is "2" but tree view only renders "foo". "foo/bar.ts" does not appear. this comes from buildChangesTree keying nodes only by path prefix. this needs to be fixed before we can merge it.

shoto290:
> nice, thank you for the PR! lgtm, just one issue I found:
> 
> if a file is replaced by a directory e.g. deleted "foo" plus added "foo/bar.ts" the changed count is "2" but tree view only renders "foo". "foo/bar.ts" does not appear. this comes from buildChangesTree keying nodes only by path prefix. this needs to be fixed before we can merge it.

Good catch! Thanks for spotting that. Just pushed a fix !

jschwxrz (approved):

PR #2214, opened 2026-05-24
feat(editor): compact single-child folder chains in file tree

shoto290:
## Summary

- Fuse chains of single-child folders into one row in the file tree (e.g. `src/components/ui` shown together instead of three separate rows).
- Truncate the path label with priority: always keep the last segment visible, then the first, dropping middles first.
- Show chains automatically as the file tree loads, without requiring a click on the parent folder.

## Test plan

- [ ] Open a project with deep single-child folder chains; verify they render as one row.
- [ ] Resize the sidebar narrow; verify path truncates through the documented stages.
- [ ] Open a fresh project; verify chains appear without clicking.
- [ ] Expand a chain; verify the deepest folder's children show.
- [ ] Drag-drop a file onto a chain row; verify it lands in the chain's deepest folder.

### DIFF
--- a/src/renderer/features/tasks/editor/compacted-path-label.test.ts
+++ b/src/renderer/features/tasks/editor/compacted-path-label.test.ts
@@ -1,0 +1,72 @@
+import { describe, expect, it } from 'vitest';
+import { pickCompactedDisplay } from './compacted-path-label';
+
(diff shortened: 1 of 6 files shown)


jschwxrz:
nice addition! thank you!

jschwxrz (approved):

PR #2216, opened 2026-05-24
fix(agent-hooks): detect Claude permission prompts

janburzinski:
# summary
- detect claude permissions prompt
- add claude pretooluse start hook


janburzinski:
@arnestrickmann this adds:
- claude nwo registers "pretooluse" as start event so that we dont mark it was not running anymore when e.g. a "ask questions" tool gets used
- claude notificaiton payload normalized into permissions_prompt and idle_prompt
- permission prompt stays visible until explicit start event

janburzinski:
we dont rely on regex purely anymore and now use hooks more :D

janburzinski:
@greptileai

PR #2220, opened 2026-05-25
fix(workspaces): format Windows lifecycle scripts

janburzinski:
# summary
- fix windows lifecycle script execution
- joining newline-separated commands with a valid cmd separator

### DIFF
--- a/src/main/core/terminals/format-lifecycle-script-input.test.ts
+++ b/src/main/core/terminals/format-lifecycle-script-input.test.ts
@@ -56,14 +56,18 @@ describe('formatLifecycleScriptInput', () => {
     );
   });
 
-  it('ignores blank lines', () => {
+  it('ignores blank lines for Windows only', () => {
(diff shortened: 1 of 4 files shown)


janburzinski:
@greptileai

arnestrickmann (approved):
thanks jan, looks good to me. isolated to Windows - low risk

arnestrickmann:
This is close, but the formatter currently normalizes POSIX lifecycle scripts too. `splitScriptLines()` runs before the `shellKind === 'cmd'` branch, so macOS/Linux/SSH scripts lose blank lines and trailing whitespace compared to the previous behavior.

Since this PR is intended to be Windows-only, can we keep POSIX input unchanged and only split/join lines for `cmd`? For example, return `exit ? `${script}; exit` : script` in the POSIX branch, and apply the newline-to-` & ` conversion only inside the cmd branch.

janburzinski:
posix scripts now stay untouched :D

PR #2221, opened 2026-05-25
fix(git): explain missing local executable

janburzinski:
# summary
- before it was "spawn git EOENT" and this makes it more friendly


arnestrickmann (approved):
thanks! a lot more user friendly.

PR #2222, opened 2026-05-25
fix: lifecycle scripts getting recognized in resource monitor

janburzinski:
# summary
- before those lifecycle scripts were not grouped correclty in the resourcemonitor

this is after:
Screenshot 2026-05-25 at 11 18 44


arnestrickmann (approved):
good addition

arnestrickmann:
good addition, @janburzinski

PR #2227, opened 2026-05-25
feat: keep agent running during e.g. update

janburzinski:
# summary
- agents keep running on quit/update via tmux
- quit/update detaches instead of killing sessions when tmux is there

### without tmux/windows
- plain pty -> terminate on quit

### open questions
- maybe put this in experiemntal
- have to find a better way because this would be tmux only
- every session by default?


arnestrickmann:
closing this draft for now to clean up a bit

PR #2230, opened 2026-05-26
fix: Copilot initial prompt flag

janburzinski:
# summary
- fixes that you couldn't have an initial prompt in copilot

### DIFF
--- a/src/main/core/agent-hooks/claude-trust-service.test.ts
+++ b/src/main/core/agent-hooks/claude-trust-service.test.ts
@@ -10,6 +10,7 @@ import { ClaudeTrustService } from './claude-trust-service';
 
 const mockReadFile = vi.hoisted(() => vi.fn());
 const mockWriteFile = vi.hoisted(() => vi.fn());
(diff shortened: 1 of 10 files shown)


janburzinski:
@greptileai

arnestrickmann:
Hi, when pasting text into the initial prompt field and then clicking "Create", the initial prompt is not passed to Copilot because I always get the "Confirm folder trust dialog"

CleanShot 2026-05-26 at 17 36 54@2x

janburzinski:
could you check again? @arnestrickmann https://streamable.com/aqzdg8

arnestrickmann:
I keep running into this @janburzinski 

CleanShot 2026-05-28 at 12 11 53@2x

janburzinski:
and it doesnt work after allowing?

arnestrickmann (approved):
this looks good to me - I flagged one edge case directly in Linear @janburzinski which is worth addressing. (related to the Cloudflare MCP server). 
Will update the user on this fix.

PR #2232, opened 2026-05-26
Add shortcut to delete selected tasks

janburzinski:
# summary
- keybind cmd+delete when in the task overview to delete quicker (using same modal)


arnestrickmann (approved):
great addition, thanks Jan!

PR #2233, opened 2026-05-26
feat(git): expand submodule changes into individual files

### LINKED ISSUE #1104: [feat]: Support Submodules
### Feature Summary

I would like the diff/PR view to support changes in submodules.

### Problem or Use Case

I just tested emdash in a monorepo with submodules.
The diff only showed a dirty submodule, but not the changes in the submodule itself.
This results in a problem when creating the PR (it didn't work).

### Proposed Solution

Show multiple repos in the diff/pr view and let them handle seperately.

### Alternatives Considered

_No response_

### Additional Context

_No response_

codyjhsieh:
## Summary

- Expand dirty submodule working trees and committed pointer moves into individual file changes in the Changes panel (e.g. `lib/child/a.txt` instead of just `lib/child`).
- Route stage, unstage, commit, push, and revert operations to the correct submodule repository, then update the parent gitlink automatically.
- Commit submodule changes before the parent and push submodule commits before the parent push to avoid dangling gitlink references.
- Expand submodule pointer diffs in PR review so file-level changes are visible.
- Add `submodule` metadata to `GitChange` so the renderer displays nested paths normally while the main process routes git operations to the correct repo.

## How it works

When `GitService` sees a changed submodule entry, it queries both repos:

1. **Parent repo** — "Did the submodule pointer move?" (`diff --raw` for gitlink OID changes)
2. **Submodule repo** — "Which files changed inside you?" (`status -z` for working tree, `diff --numstat` between OIDs for pointer moves)

The expanded file changes replace the opaque submodule row in both staged and unstaged lists. For diffs, file contents are read from the submodule at the OID recorded by the parent gitlink. For commits, staged submodule files are committed inside the submodule first, the resulting gitlink is staged in the parent, and then the parent is committed.

## Test plan

- [x] `pnpm run format` — clean
- [x] `pnpm run lint` — clean
- [x] `pnpm run typecheck` — clean
- [x] `pnpm vitest run src/main/core/git/impl/git-service.test.ts` — 35/35 pass

### New integration tests

- **Expands dirty submodule files in full status** — edits a file inside a submodule, verifies `getFullStatus()` returns the inner file path with correct submodule metadata.
- **Does not expand uninitialized submodules as parent-repo files** — deinits a submodule, verifies its internal paths don't leak into parent status.
- **Stages and commits submodule file edits before committing the parent gitlink** — stages a submodule file, commits, verifies both repos end up clean and the parent commit references the submodule.
- **Expands committed submodule pointer changes to inner files** — commits inside the submodule directly, verifies the parent status shows expanded inner file changes (not just the submodule root).
- **Reads submodule file contents from the parent gitlink ref for diffs** — verifies `getFileAtHead()` and `getFileAtIndex()` resolve to the correct submodule OID contents.

### Manual verification

1. Created a repo with a submodule, opened it as an Emdash project, created a task.
2. Edited a file inside the submodule in the task worktree.
3. Confirmed the Changes panel shows `lib/child/a.txt` (not just `lib/child`).
4. Clicked the file and confirmed the diff shows the correct content.
5. Staged, committed — both repos returned to clean state.
6. Verified the parent commit includes the gitlink update and the submodule commit was created first.

Closes #1104

### DIFF
--- a/src/main/core/git/impl/git-service.test.ts
+++ b/src/main/core/git/impl/git-service.test.ts
@@ -1,6 +1,13 @@
+import { execFile } from 'node:child_process';
 import { createHash } from 'node:crypto';
+import { promises as fs } from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { promisify } from 'node:util';
 import { describe, expect, it } from 'vitest';
+import { LocalExecutionContext } from '@main/core/execution-context/local-execution-context';
 import type { IExecutionContext } from '@main/core/execution-context/types';
+import { LocalFileSystem } from '@main/core/fs/impl/local-fs';
 import type { FileSystemProvider } from '@main/core/fs/types';
 import { GitService } from './git-service';
 import { computeBaseRef } from './git-utils';
@@ -44,6 +51,7 @@ const BRANCH_FORMAT =
   'branch -a --format=%(refname:short)|%(upstream:short)|%(upstream:track)|%(refname)';
 
 const stubFs = {} as FileSystemProvider;
+const execFileAsync = promisify(execFile);
 
 function makeContext(exec: MockExec, root = '/repo'): IExecutionContext {
   return {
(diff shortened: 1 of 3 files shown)


arnestrickmann:
@codyjhsieh 
Thank you for the thoughtful PR and the detailed validation. We have submodule support in mind, but I’m going to close this for now since we won’t be able to get this PR reviewed and landed as-is at the moment. 

Really appreciate you taking the time to work through this.

PR #2243, opened 2026-05-27
  fix(tasks): hydrate conversations and terminals on demand

jschwxrz:
- bounds PTY respawns to unexpected exits to avoid runaway process growth
- loads only initially open conversations during task hydration
- hydrates conversation sessions only while their tabs are open, then dehydrates them when closed
- hydrates terminal sessions lazily on first frontend connection
- extracts conversation hydration reconciliation into a focused disposable state machine


jschwxrz:
@greptileai

Davidknp (approved):

PR #2250, opened 2026-05-27
fix(pty): support tmux clipboard copy

janburzinski:
# summary
- fixes that copying in tmux doesnt work 

(very bad) demo:
https://streamable.com/smc5mi


janburzinski:
@greptileai

arnestrickmann (approved):
thanks jan - user will be happy

PR #2251, opened 2026-05-27
fix(opencode): sync theme state

janburzinski:
# summary
- fixes opencode not getting the right theme 

another bad demo:
https://streamable.com/th9x1a


janburzinski:
@greptileai

PR #2252, opened 2026-05-27
feat: allow renaming projects

janburzinski:
# summary
- allows renaming projects

demo:
https://streamable.com/c7qmk3


janburzinski:
@greptileai

apps/emdash-desktop/src/renderer/features/projects/stores/project.ts
```diff
@@ -128,6 +132,13 @@ export class ProjectStore {
     this.errorCode = undefined;
   }
 
+  rename(data: LocalProject | SshProject): void {
```
Davidknp: Just a single rename method that does an rpc call and updates the data and name property on the ProjectStore should suffice, no? No need for a rename method on the project-manager it should be limited to managing the lifecycle of projects.

apps/emdash-desktop/src/main/core/projects/operations/renameProject.ts
```diff
@@ -0,0 +1,56 @@
...
+  name: string
+): Promise<Result<Project, RenameProjectError>> {
+  const trimmedName = name.trim();
+  if (!trimmedName || trimmedName.length > MAX_PROJECT_NAME_LENGTH) {
+    return err({ type: 'invalid-name' });
+  }
+
+  try {
+    const existingProject = await getProjectById(projectId);
+    if (!existingProject) return err({ type: 'project-not-found' });
```
Davidknp: No need to check this before just do the update query it will fail if the project doesn't exist also no need to return the whole project here and use result types. It is just a database update the likelihood of this failing is very low.

janburzinski:
@greptileai

janburzinski:
@Davidknp thanks for the feedback!! :D updated

arnestrickmann:
@Davidknp could you review?

Issue #1890, opened 2026-05-05
[feat]: add back "preview" button


kchung:
### Feature Summary

Before V1, there was an ability to click on an "Preview" button to open a customized URL for your task. This seemed to have been removed

### Problem or Use Case

This feature solves a common ADE workflow of having many tasks and worktrees open and being able to just open a task-specific browser window to test

### Proposed Solution

The solution would match the pre-v1 functionality:

1. Allow the user to configure what the "Preview" url would be
  a. Expose Emdash environment variables (like port, task name)
2. Add a "Preview" button in the UI

arnestrickmann:
Thanks @kchung for opening this issue - bringing back an in app browser is on our roadmap!

Ingole712521:
Can i work on this

arnestrickmann:
Thanks for updating this, @kchung

PR #2253, opened 2026-05-27
fix(ssh): rewrite dev-server preview URL to SSH host for remote projects

### LINKED ISSUE #1979: [feat]: Dev server preview URL should use SSH connection host for remote projects
### Feature Summary

When running a dev server on a remote SSH project, the dev server pill button opens the URL with `127.0.0.1` (or `localhost`), which refers to the **local** machine — not the remote host where the server is actually running. For SSH projects, the URL should use the SSH connection's configured hostname so the browser can reach the remote dev server.

### Problem or Use Case

I develop on a remote server via SSH (Emdash's SSH connection feature). When I start a dev server (e.g. Rails/Puma), Emdash detects the URL from terminal output (e.g. `http://0.0.0.0:51710`), normalises it to `http://127.0.0.1:51710`, and shows the green pill button. Clicking it opens my local browser to `127.0.0.1:51710` — which is my laptop, not the remote server. The page doesn't load.

The SSH connection already has the remote hostname configured (e.g. `my-remote-host.example`). The preview URL should use that host instead.

### Proposed Solution

In `normalizeUrl` and `parseTarget` (around line 11697 in the bundled `main/index.js`), when the workspace is an SSH project, replace `0.0.0.0` / `127.0.0.1` / `localhost` with the SSH connection's `host` field rather than hardcoding `127.0.0.1`.

Something like:

```js
function normalizeUrl(raw, sshHost) {
  if (sshHost) {
    return raw
      .replace("0.0.0.0", sshHost)
      .replace("127.0.0.1", sshHost)
      .replace("localhost", sshHost);
  }
  return raw.replace("0.0.0.0", "127.0.0.1");
}
```

The SSH host is already available via `sshConnectionManager` and the project's `connectionId` — it just needs to be threaded through to `wireTerminalDevServerWatcher` in `SshTerminalProvider`.

### Alternatives Considered

* **SSH port forwarding**: Doesn't scale with dynamic per-worktree ports (Emdash generates ports from a hash of the worktree path).
* **SOCKS proxy**: Requires browser configuration — adds friction.
* **Patching the asar locally**: Currently working around this by modifying `normalizeUrl` in the extracted asar. Effective but breaks on every update.

### Additional Context

* The `SshTerminalProvider` already passes `probe: false` to `wireTerminalDevServerWatcher`, which is correct (can't probe localhost for a remote server). The URL rewriting is the missing piece.
* Related to generalaction/emdash#1890 (configurable preview URL) but more specific — this is about the existing auto-detected dev server pill using the correct host for SSH projects.
* `URL_PATTERN` regex (`/https?:\/\/(?:localhost|127\.0\.0\.1|0\.0\.0\.0)(:\d{2,5})?/`) already captures the URL correctly from terminal output; only the normalisation step needs the SSH host awareness.

Aarekaz:
## Summary

Fixes #1979.

When a dev server in an SSH-remote terminal logs `http://0.0.0.0:3000`, the preview pill currently opens `http://127.0.0.1:3000` on the user's local machine — wrong, since the server lives on the remote host.

This PR threads the SSH connection host into the URL emitted by the dev-server watcher:

1. `SshConnectionManager` records `config.host` at connect time in a `Map`, cleared only on intentional `disconnect()` and reconnect exhaustion. Exposes `getHost(id)`.
2. `SshTerminalProvider` snapshots the host once in its constructor (callers establish the SSH connection before constructing the provider, so the map is already populated).
3. `dev-server-watcher`'s `normalizeUrl` now takes an optional `sshHost` and rewrites the URL hostname via WHATWG URL parsing (handles IPv4 + plain IPv6, falls back to the raw URL on unparseable input).

The local-terminal path is unchanged. Probing was already correctly disabled for SSH sessions, so only the URL emission needed fixing.

## Design notes

- Host map lifetime = connection-identity lifetime. It survives transient `close` events and reconnect cycles. Cleared on: `disconnect()` (all three exit branches), max reconnect attempts, and auth failure during reconnect.
- Initial-connect failure intentionally leaves one entry in the map until the next `connect(id)` attempt overwrites it or `disconnect(id)` clears it. This is the simpler invariant (per second-opinion review) — adding cleanup in `client.on('error')` would race with the reconnect path.
- IPv6 hosts are bracketed manually before assignment to `URL.hostname` (the WHATWG setter silently ignores unbracketed IPv6 literals).

## Out of scope

- Port forwarding / firewall scenarios where the remote port is not reachable from the user's laptop — that's a networking concern, not a URL concern.
- Custom port mapping (assumes same port on remote as bound locally).
- Configurable preview URL (tracked in #1890).

## Test plan

- [x] `pnpm run format`
- [x] `pnpm run lint`
- [x] `pnpm run typecheck`
- [x] `pnpm test` — 836/836 passing including:
  - 4 new manager tests (`ssh-connection-manager.test.ts`) covering: host populated on connect, host survives transient `close`, host cleared on normal disconnect, host cleared on disconnect-of-never-connected.
  - 9 new `normalizeUrl` tests (`dev-server-watcher.test.ts`) covering: no-sshHost regression, hostname rewrite for 127.0.0.1 / localhost / 0.0.0.0, IPv4 and IPv6 SSH hosts, unparseable-URL fallback.
- [ ] Manual verification on an SSH remote (would appreciate maintainer confirmation).

### DIFF
--- a/src/main/core/ssh/ssh-connection-manager.test.ts
+++ b/src/main/core/ssh/ssh-connection-manager.test.ts
@@ -1,0 +1,103 @@
+import { EventEmitter } from 'node:events';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+class MockClient extends EventEmitter {
+  static lastInstance: MockClient | null = null;
+  endCalled = false;
+
+  constructor() {
+    super();
+    MockClient.lastInstance = this;
+  }
+
+  connect(_config: unknown): void {
+    // no-op — tests drive 'ready' / 'error' / 'close' manually
+  }
+
+  end(): void {
+    this.endCalled = true;
+    // simulate ssh2 client closing after end()
+    queueMicrotask(() => this.emit('close'));
+  }
+}
+
+vi.mock('ssh2', () => ({
+  Client: MockClient,
+}));
+
+vi.mock('@main/db/client', () => ({
+  db: { select: () => ({ from: () => ({ where: () => ({ limit: () => [] }) }) }) },
+}));
+
+vi.mock('@main/lib/logger', () => ({
+  log: { info: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}));
+
+vi.mock('@main/lib/events', () => ({
+  events: { emit: vi.fn() },
+}));
+
+// Import AFTER mocks are registered.
+const { SshConnectionManager } = await import('./ssh-connection-manager');
+
+const baseConfig = { host: 'remote.example.com', username: 'tester' };
+
+describe('SshConnectionManager host lifecycle', () => {
+  beforeEach(() => {
+    MockClient.lastInstance = null;
+  });
+
+  it('records host on successful connect and exposes it via getHost', async () => {
+    const mgr = new SshConnectionManager();
+    const pending = mgr.connectFromConfig('c1', baseConfig);
+
+    // Drive a successful handshake.
+    queueMicrotask(() => MockClient.lastInstance?.emit('ready'));
+    await pending;
+
+    expect(mgr.getHost('c1')).toBe('remote.example.com');
+  });
+
+  it('keeps host populated when client emits transient close (reconnect path)', async () => {
+    const mgr = new SshConnectionManager();
+    const pending = mgr.connectFromConfig('c1', baseConfig);
+    queueMicrotask(() => MockClient.lastInstance?.emit('ready'));
+    await pending;
+
+    // The 'close' handler in createConnection schedules a reconnect. We don't
+    // care about that here — only that the host map stays intact so the
+    // SshTerminalProvider snapshot remains valid across the reconnect window.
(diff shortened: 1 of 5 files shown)


Aarekaz:
Addressed all three in 51b7d255:
- Unified `normalizeUrl` through `URL` parsing → no SSH/non-SSH trailing-slash drift.
- `config.host ?? 'localhost'` (ssh2's default) instead of `''`.
- Renamed the close-event test to reflect what it actually covers; reconnect re-entry is implicitly covered by the successful-connect test.

Tests still 837/837 green.

arnestrickmann:
Thank you for this PR and for your interest in contributing, and thank you for your feedback and engagement lately. We really appreciate it. We already implemented SSH port forwarding in another, more recent PR - would love to hear your thoughts once you've had the time to test.

arnestrickmann:
Closing this PR

PR #2262, opened 2026-05-28
feat(tasks): add mark as unread action

janburzinski:
# summary
- adds the ability to mark a task as unread again

demo:
https://streamable.com/c3ppan


janburzinski:
@greptileai

PR #2263, opened 2026-05-28
forward model override env vars from shell to agent processes

sergeym-monday:
## Summary

- Adds `ANTHROPIC_DEFAULT_HAIKU_MODEL`, `ANTHROPIC_DEFAULT_OPUS_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, and `ANTHROPIC_SMALL_FAST_MODEL` to `AGENT_ENV_VARS` so model override env vars set in the user's shell profile are forwarded to spawned agent processes
- Adds `CLAUDE_CODE_SUBAGENT_MODEL`, `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS`, and `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` for the same reason

## Why

`buildAgentEnv()` creates a minimal environment from an explicit allowlist, so any var not in `AGENT_ENV_VARS` is silently dropped — even if the Emdash main process inherited it from the user's login shell.

Tools like axcli set model override vars in `~/.zshrc` (e.g. `ANTHROPIC_DEFAULT_SONNET_MODEL=claude-sonnet-4-6[1m]`) to unlock extended capabilities such as 1M context windows. Without this fix, Claude Code sessions spawned inside Emdash fall back to their hardcoded 200k default context window while the same Claude Code version launched directly from a terminal sees 1M — because the terminal sources the shell profile but Emdash's clean env does not include those vars.

## Test plan

- [x] Run `pnpm run typecheck` — passes
- [x] Run `pnpm run lint` — passes (pre-existing unrelated warning only)
- [x] Open a Claude Code session inside Emdash with model override vars set in shell profile; confirm context window matches what the terminal shows


arnestrickmann (approved):
Thank you for your second contribution, @sergeym-monday - appreciate it!

This looks good to me. Thanks for the addition!

PR #2267, opened 2026-05-28
feat(ssh): add setup field info tooltips

janburzinski:
added these texts:
- SSH Config “Select an entry from ~/.ssh/config to prefill host, user, key, proxy, and agent forwarding settings.”
- Authentication “Choose how Emdash authenticates to the remote server. SSH config entries can preselect the best option.”
- Private Key Path “Path on this machine to the private key used for the connection, for example ~/.ssh/id_ed25519.”
- Passphrase “Only needed if the selected private key is encrypted with a passphrase.”
- ProxyCommand “Command from your SSH config used to reach this host through a proxy. It is read-only here because it comes from ~/.ssh/config.”
- ProxyJump “Optional bastion host to connect through before reaching the target server, for example user@bastion:2222.”
- ForwardAgent “Forward your local SSH agent to the remote server so nested SSH and Git commands can use your loaded local keys. Enable only for trusted hosts.”

demo:
https://streamable.com/m88nel

### DIFF
--- a/src/renderer/lib/components/add-ssh-conn-modal.tsx
+++ b/src/renderer/lib/components/add-ssh-conn-modal.tsx
@@ -496,7 +496,7 @@ export function AddSshConnModal({
             <form.Field name="authType">
               {(field) => (
                 <FieldSet>
-                  <div className="mb-3 flex w-fit items-center gap-1.5">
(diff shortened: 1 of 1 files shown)


janburzinski:
@greptileai

jschwxrz (approved):

PR #2272, opened 2026-05-28
fix(commands): ignore allowed shortcut conflicts

janburzinski:
# summary
- fixes that cmd+w would get processed twice

demo of it working now:
https://streamable.com/fdx3s7


arnestrickmann (approved):
thanks for catching this and for the fix

PR #2273, opened 2026-05-28
fix(pty): prevent duplicate clipboard image paste

janburzinski:
# summary
- fixes that cmd+v would double paste an image 
- control+v/cmd+v both still work 

demo of the fix:
https://streamable.com/fdx3s7


arnestrickmann (approved):
thanks for the fix!

correctly only pasting one image when testing. will update the user who reported it.

Image

PR #2274, opened 2026-05-28
fix: cmdk file search not working when stuck

janburzinski:
# summary
- fixes file search in cmdk when workspace no t indexed yet
- lazy workspace file indexing on search as a fallback
- max waiting 250ms for indexing to prevent lag 
- refreshes cmdk results after bg refresh finishes


janburzinski:
@greptileai

PR #2280, opened 2026-05-29
fix(tasks): prevent task rename from closing pull requests

jschwxrz:
- keep branch rename local-only and remove remote branch delete and push behavior
- add an optional rename branch checkbox to RenameTaskModal
- disable branch rename for linear-linked tasks, open pull requests and shared branches
- update renameTask to rename task metadata by default and only rename the branch when requested
- add backend guards so stale callers cannot rename protected branches


jschwxrz:
@greptileai

PR #2286, opened 2026-05-29
fix: saving markdown files while editing doesnt work via cmds

janburzinski:
# summary
- fixes that saving a markdown file works again

demo:
https://streamable.com/nu2ulm


arnestrickmann (approved):
thank you for the quick fix!

PR #2287, opened 2026-05-29
fix: amp task does not resume correctly after restart

janburzinski:
# summary
- fixes amp not resuming correctly

demo:
https://streamable.com/oaqvpp


janburzinski:
@greptileai

arnestrickmann:
Thank you @janburzinski - I think this is already updated as of today on main with the merge of a newer PR, is that correct?

PR #2290, opened 2026-05-29
fix(tasks): respect include issue context default

janburzinski:
# summary
- fixes issue context only being included automatically when active in settings


arnestrickmann (approved):
LGTM, setting is now respected in task creation.

thanks jan

PR #2291, opened 2026-05-29
fix(agent-hooks): delegate codex notifications to script

janburzinski:
# summary
- codex notifications now delegate to generated shell/powerhsll scripts
- session start/permissions/idle notifcations handled through same hook path

demo:
https://streamable.com/z88n8b


janburzinski:
@greptileai

PR #2294, opened 2026-05-29
feat(terminals): add configurable default shell

jschwxrz:
- adds an app-level default terminal shell setting for local projects
- persists each terminal shell id so restored terminals keep their selected shell
- resolves system shells at spawn time and falls back to system when a stored shell is unavailable
- keeps lifecycle script terminals on the system shell
- narrows shell picker options to common interactive shells and shows resolved system shell labels
- adds shell icons for zsh, bash, fish, powershell, pwsh and cmd


jschwxrz:
@greptileai

PR #2298, opened 2026-05-30
fix(conversations): respawn agent ptys after exit

jschwxrz:
- restore agent conversation respawn when the underlying pty exits
- replace backend ptys under a stable logical session id so the renderer stays attached
- reset the renderer pty view when the backend starts a replacement process
- guard stale pty exits so old processes cannot unregister newer replacements
- simplify conversation replacement state and remove unused runtime events
- cover local and ssh replacement behavior with focused integration-style tests


jschwxrz:
@greptileai

PR #2303, opened 2026-05-30
fix(resource-monitor): label terminal sessions

janburzinski:
# summary
- fixes that open terminals would just get displayed as uuids instead of "temrinal 1" for example

demo:
image
before: https://streamable.com/re2ub0


arnestrickmann (approved):
good addition

PR #2304, opened 2026-05-30
fix(linear): include refreshed issue context

janburzinski:
# summary
- fixes that the settings toggle for "including issue context" now works and is respecetd 

# demo
https://streamable.com/wkvlz3


janburzinski:
@greptileai

PR #2309, opened 2026-05-30
fix(resource-monitor): show real usage of processes

janburzinski:
# summary
- count full pty process tree, not just shell wrapper PIDs
- fix underreported agent/script cpu and memory


janburzinski:
@greptileai

jschwxrz (approved):

PR #2318, opened 2026-05-31
fix(pty): terminal rendering when opening settings

janburzinski:
### Description
we had the issue that the terminal would bug out compeltely and look very weird. this should hopefully address some of that 

- keep workspace mounted behind settings
- avoid xterm remount on cmd,
- fix terminal rendering after returning

https://streamable.com/zwpdk9

<details>
<summary>Checklist</summary>

</details>


janburzinski:
@greptileai

PR #2321, opened 2026-05-31
feat(cli): emdash workspace CLI (list / create / remove / send)

astaub:
> **Draft / for discussion.** This is a headless CLI we built while dogfooding emdash. Opening it to get your read on the *approach* (especially the coupling to `@main/*` internals) before polishing for merge — not asking to land as-is.

## What it is

An `emdash` command-line surface that operates on the **same SQLite DB + git worktrees as the desktop app**, by reusing the app's own main-process modules (Drizzle schema, `computeWorkspaceKey`, `WorktreeService`, `buildAgentSessionCommand`, `makeTmuxSessionName`/`makePtySessionId`) so the rows/worktrees/agent sessions it creates match UI-created ones.

```
emdash workspace list   [--project <name>] [--include-archived] [--json]
emdash workspace create --project <p> --branch <b> [--base <branch>] [--name <n>]
                        [--checkout-existing | --no-worktree] [--push-branch]
                        [--prompt "<text>" [--agent <name>] [--auto-approve]] [--json]
emdash workspace remove --project <p> (--branch <b> | --id <id>)
                        [--pre-remove <cmd>] [--skip-hook] [--force] [--json]
emdash workspace send   --project <p> (--branch <b> | --id <id>) --message "<text>" [--json]
```

The motivating use case: an external orchestrator can spin a worktree and dispatch its agent in one call — `create --branch X --prompt "<packet>"` — then `send` follow-ups, `list` as a board, and `remove` (with a capture-before-delete hook) to tear down.

## How it runs

CJS bundle (esbuild → `out/cli/index.cjs`) run under `ELECTRON_RUN_AS_NODE` so `better-sqlite3`'s native ABI matches the desktop build. `pnpm build:cli` then `pnpm cli workspace list`.

## Design

- **Core is db-injected and Electron-free** (`src/main/cli/{workspace-commands,agent-dispatch,local-worktree,args}.ts`) → unit-tested against a temp SQLite DB via `openFixture`. The entry (`index.ts`) wires the real DB + settings + agent dispatcher.
- **`create` is atomic** (git worktree first, DB rows only on success), idempotent by branch, and resolves remote base refs (`origin/main`).
- **`remove` is hook-gated**, defaults to safe branch delete (`-d`; `-D` only with `--force`), kills the agent's tmux session first, scopes `--id` to the project, and only removes worktrees inside the project's pool.
- **`create --prompt`** launches the agent in a *detached tmux session* named the way the app expects, so the app re-attaches when the task is opened (requires tmux mode on).

## Tests

`src/main/cli/*.test.ts` (node) + `*.db.test.ts` (main-db, real SQLite + git): 38 tests covering arg parsing, list, create (atomicity/idempotency/remote-base/push/prompt-dispatch), remove (teardown/idempotency/hook-abort/unmerged-branch retention/cross-project `--id` refusal/out-of-pool rm guard), send, and the pure tmux-argv builder.

> ⚠️ emdash CI (`code-consistency-check.yml`) runs only format/typecheck/lint — **not** the test suite. These were run locally (all green). Happy to add a CI step to run `vitest --project node --project main-db` if you'd want that.

## Honest risk register (full version in `src/main/cli/README.md`)

- **Couples to `@main/*` internals** (not a public API) — the CLI must move in lockstep with the app. Typecheck catches signature drift (this branch rebased cleanly onto current `main`, 86 commits of churn, no changes needed); *runtime* drift (tmux naming, command builder) would need a smoke test. The biggest open question for you: is this coupling acceptable, or should the shared bits be extracted into a package both consume?
- **`remove` is `rm`-class** — mitigated (safe branch delete, pool-scoped worktree rm, project-scoped `--id`, session kill, capture hook) but it tears down real state.
- **No app-concurrency coordination** (writes the live DB while the app may run; WAL handles locking, not logical races).
- **Deleting a task leaves orphaned `conversations`/`terminals`/`messages` rows** — this matches the app's own `deleteTask` (FK cascade is off app-wide); not changed here to avoid divergence.
- **`create --prompt` adoption needs tmux mode on**; hook env / `.emdash.json` task env aren't forwarded to CLI-launched agents; keystroke-injection providers don't get the prompt at launch (reported, non-zero exit).
- **Scope:** local projects only; no `--from-issue/--from-pr`, no prompt templates.

## Not yet verified in the live app

- A CLI-created workspace appears in the sidebar after reload (confirmed once by a tester; the app doesn't live-refresh external DB writes).
- The app re-attaches to a `create --prompt` tmux agent (vs spawning a duplicate) with tmux mode on.

### DIFF
--- a/bin/emdash-cli.mjs
+++ b/bin/emdash-cli.mjs
@@ -1,0 +1,53 @@
+#!/usr/bin/env node
+/**
+ * Launcher for the emdash CLI.
+ *
+ * Responsibilities (kept Electron-free so it can run under plain Node OR
+ * Electron-as-Node):
+ *   1. Resolve EMDASH_DB_FILE to the shared app database BEFORE the bundled CLI
+ *      loads, so `@main/db/client` never needs Electron's `app.getPath`.
+ *   2. Dynamically import the built bundle (out/main/cli.js), which parses argv
+ *      and runs the requested command.
+ *
+ * Intended to be run via Electron-as-Node so the better-sqlite3 native ABI
+ * matches the desktop build:
+ *   ELECTRON_RUN_AS_NODE=1 electron bin/emdash-cli.mjs workspace list
+ */
+
+import os from 'node:os';
+import path from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+const CURRENT_DB_FILENAME = 'emdash4.db';
+
+function defaultUserDataPath() {
+  const home = process.env.HOME ?? os.homedir();
+  if (process.platform === 'darwin') {
+    return path.join(home, 'Library', 'Application Support', 'emdash');
+  }
+  if (process.platform === 'win32') {
+    const appData = process.env.APPDATA ?? path.join(home, 'AppData', 'Roaming');
+    return path.join(appData, 'emdash');
+  }
+  const xdgConfig = process.env.XDG_CONFIG_HOME ?? path.join(home, '.config');
+  return path.join(xdgConfig, 'emdash');
+}
+
+if (!process.env.EMDASH_DB_FILE || !process.env.EMDASH_DB_FILE.trim()) {
+  process.env.EMDASH_DB_FILE = path.join(defaultUserDataPath(), CURRENT_DB_FILENAME);
+}
+
+const here = path.dirname(fileURLToPath(import.meta.url));
+const bundlePath = path.resolve(here, '..', 'out', 'cli', 'index.cjs');
+
+try {
+  await import(pathToFileURL(bundlePath).href);
+} catch (error) {
+  if (error && (error.code === 'ERR_MODULE_NOT_FOUND' || error.code === 'MODULE_NOT_FOUND')) {
+    process.stderr.write(
+      `emdash CLI bundle not found at ${bundlePath}.\nRun \`pnpm build:cli\` first.\n`
+    );
+    process.exit(1);
+  }
+  throw error;
+}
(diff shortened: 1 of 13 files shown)


bherila:
I hit a packaged macOS canary failure that looks directly relevant to this CLI branch: bherila/emdash#1.

The packaged app contains both expected CLI files in `app.asar`:

- `/bin/emdash-cli.mjs`
- `/out/cli/index.cjs`

but this packaged invocation fails:

```bash
APP="$HOME/Applications/Emdash Canary.app"
BIN="$APP/Contents/MacOS/Emdash Canary"
CLI="$APP/Contents/Resources/app.asar/bin/emdash-cli.mjs"

ELECTRON_RUN_AS_NODE=1 "$BIN" "$CLI" --help
```

The launcher reports:

```text
emdash CLI bundle not found at .../app.asar/out/cli/index.cjs.
Run `pnpm build:cli` first.
```

but the deeper error is:

```text
Error: Cannot find module 'electron'
Require stack:
- .../out/cli/index.cjs
```

So I think there are two separate issues worth covering before this lands:

1. The CLI bundle still has a packaged runtime dependency on `electron`.
2. `bin/emdash-cli.mjs` appears to catch nested `MODULE_NOT_FOUND` from inside the bundle and re-label it as “CLI bundle not found,” even when the bundle exists.

Would you prefer this packaged-app case be handled inside this draft PR, or should I open a focused follow-up PR that:

- keeps the CLI import graph Electron-free in packaged Electron-as-Node mode;
- changes launcher error handling to distinguish a genuinely missing bundle from nested module-resolution failures;
- adds a packaged CLI smoke test or at least a `build:cli && --help` bundle smoke test?

jschwxrz:
hi, will close the pr as it is no longer up to date

PR #2354, opened 2026-06-02
fix(worktrees): repair stale worktree directory reuse

jschwxrz:
- removes invalid stale worktree directories before creating or checking out a worktree at the same path
- surfaces stale directory cleanup failures as worktree setup failures instead of silently continuing
- keeps worktree deletion tied to the updated workspace model
- adds retry-aware local recursive removal for transient filesystem cleanup failures


jschwxrz:
@greptile review

PR #2424, opened 2026-06-08
feat(plane): add issue provider integration

### LINKED ISSUE #2414: [feat]: Configurable Plane URL for self-hosted Plane instances
### Feature Summary

Allow self-hosted [Plane](https://plane.so) users to point emdash at their own Plane instance by configuring a custom base URL, instead of being locked to the hosted `app.plane.so` cloud.

### Problem or Use Case

Plane is an open-source project management tool, and a large share of its users self-host it on their own domain (e.g. `https://plane.mycompany.com`). Today emdash's issue providers cover Linear, GitHub, Jira, GitLab, Forgejo, Asana, Monday, Trello, Featurebase and Plain — but there's no way to connect to a self-hosted Plane instance with a custom URL.

For teams that have standardized on self-hosted Plane, this means they can't pull their Plane issues/work items into emdash at all, even though the rest of their stack is supported.

### Proposed Solution

When connecting Plane in emdash settings, expose a **Plane instance URL** field (defaulting to the Plane cloud URL) alongside the API token, so users can enter their self-hosted base URL.

This mirrors the pattern emdash already uses for other self-hostable providers like Forgejo and GitLab, where the instance URL is normalized and validated (see `src/main/core/issues/helpers/hosted-instance.ts` → `normalizeHostedInstanceUrl`). Reusing that helper would keep validation/normalization consistent (enforce http/https, strip trailing slashes, reject query/hash, etc.).

Expected UX:
- A "Plane instance URL" input in the Plane connection form, pre-filled with the cloud default.
- API requests target `<instanceUrl>/api/...` so both cloud and self-hosted work from the same code path.
- Clear error if the URL is unreachable or invalid.

### Alternatives Considered

- Hardcoding only the Plane cloud URL — excludes the (large) self-hosted user base.
- Using a generic MCP/webhook workaround — loses the first-class issue-provider integration that the other trackers get.

### Additional Context

Plane's self-hosted deployments are first-party and common, so a configurable base URL is essential for parity with how emdash already handles self-hosted Forgejo and GitLab. Happy to help test against a self-hosted Plane instance.

arnestrickmann:
## Summary

- add Plane as a first-class issue provider, separate from Plain
- support Plane Cloud and self-hosted Plane instances through a configurable API base URL
- add Plane credential validation, list/search/context support, renderer setup UI, provider metadata, and local Plane logo asset

## Self-hosted support

- defaults the API base URL to `https://api.plane.so` for Plane Cloud
- lets users enter a self-hosted Plane base URL such as `https://plane.example.com`
- normalizes and validates the configured URL with the existing hosted-instance helper
- stores API keys in encrypted app secrets and non-secret config in KV
- builds all API calls from the configured base URL and current `/api/v1/.../work-items/` endpoints

## Validation

- `corepack pnpm run format:check`
- `corepack pnpm run typecheck`
- `corepack pnpm run lint` (passes with existing unrelated `use-automations.ts` warning)
- `corepack pnpm exec vitest run --project node src/main/core/plane/plane-client.test.ts src/main/core/plane/plane-connection-service.test.ts src/main/core/plane/plane-issue-provider.test.ts`

Closes #2414


arnestrickmann:
@greptile

arnestrickmann:
@greptile

PR #2431, opened 2026-06-09
feat(diff): rendered markdown preview toggle in the diff viewer

george-vii:
## What
Adds an Eye/Pencil toggle to markdown files opened from the Changes/diff sidebar —
the same toggle (and position) the file editor already uses for markdown — letting
you flip between the source diff and the rendered markdown.

- Pencil (default) → Monaco source diff (unchanged)
- Eye → the modified ("after") content rendered via `MarkdownRenderer`
- Split mode → original rendered (left) ∥ modified rendered (right), reusing the
  diff toolbar's existing unified/split toggle

Deleted markdown files stay diff-only (no "after" content to render).

## Why
The rendered markdown view was only reachable from the file explorer, not when
reviewing changes — so reviewing a `.md` diff meant reading raw source.

## How
- Reuses the shared `PreviewSourceToggle`, `MarkdownRenderer`, and the `ShowHide`
  mount-preservation pattern (keeps Monaco's diff models registered so the preview
  can read their content).
- Extracts a `computeDiffUris()` helper as the single source of truth for the
  original/modified model URIs (shared by the Monaco diff editor and the preview).
- Adds optional `previewLabel`/`sourceLabel` props to `PreviewSourceToggle`
  (defaults preserved).

## Screenshots

**Pencil (default) — source diff:**
image

**Eye — rendered preview:**
image

**Eye + Split — original ∥ modified, rendered:**
image

## Testing
- New unit test for the diff tab's preview-toggle state.
- `pnpm run typecheck`, `lint`, `format:check`, and `test` all pass.
- Manually verified: unified + split, staged and unstaged changes.

### DIFF
--- a/apps/emdash-desktop/src/renderer/features/tasks/diff-view/main-panel/diff-file-renderer.tsx
+++ b/apps/emdash-desktop/src/renderer/features/tasks/diff-view/main-panel/diff-file-renderer.tsx
@@ -13,14 +13,22 @@ import {
 } from '@renderer/features/tasks/task-view-context';
 import { getFileKind } from '@renderer/lib/editor/fileKind';
 import { PreviewSourceToggle } from '@renderer/lib/editor/preview-source-toggle';
+import { useDelayedBoolean } from '@renderer/lib/hooks/use-delay-boolean';
 import { rpc } from '@renderer/lib/ipc';
 import { modelRegistry } from '@renderer/lib/monaco/monaco-model-registry';
 import { buildMonacoModelPath } from '@renderer/lib/monaco/monacoModelPath';
 import { StickyDiffEditor } from '@renderer/lib/monaco/sticky-diff-editor';
 import { MarkdownRenderer } from '@renderer/lib/ui/markdown-renderer';
 import { ShowHide } from '@renderer/lib/ui/show-hide';
+import { Spinner } from '@renderer/lib/ui/spinner';
 import { getLanguageFromPath } from '@renderer/utils/languageUtils';
-import { gitRefToString, HEAD_REF, STAGED_REF, type GitObjectRef } from '@shared/core/git/git';
+import {
+  gitRefToString,
+  HEAD_REF,
+  STAGED_REF,
+  type GitObjectRef,
+  type GitRef,
+} from '@shared/core/git/git';
 import { getDraftCommentTargetKey, type DraftCommentTarget } from '@shared/lineComments';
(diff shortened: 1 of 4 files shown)


arnestrickmann:
Thanks for contributing this, the direction fits nicely with the existing markdown/file preview work.

One correctness concern before merge: the rendered markdown preview needs to stay tied to the same source as the diff. Right now the preview reads markdown from the Monaco diff models, but linked images are loaded from the current workspace filesystem. For staged/git/PR diffs that can show images from the wrong state, or miss images that exist in the index/ref being previewed. It would be better to resolve images from the same side/source being rendered: filesystem for working tree, index for staged, and getImageAtRef for commit/PR refs.

Also, please make sure the preview re-renders when the underlying diff models are invalidated or refreshed. The current observable dependencies look like they may miss disk model content changes, so the source diff could update while the rendered preview remains stale.

The UI placement and reuse of existing renderer/toggle pieces looks good; I think these two data-source/reactivity fixes would make it much safer.

george-vii:
Hey @arnestrickmann, no worries at all — and thanks for the thorough review! Both points are addressed:

**1. Per-side image source.** Linked images now resolve from the same source as the side being rendered, mirroring the original/modified ref selection used for the diff text:
- working tree → `fs.readImage`
- index (staged side) → `git.getImageAtIndex`
- commit/PR refs → `git.getImageAtRef`

So staged/git/PR previews read images from the matching ref instead of the working tree. One caveat worth flagging: on SSH workspaces the git-side reads return `unavailable`, so a referenced image currently renders as a broken-image placeholder rather than the explicit "unavailable" message `ImageDiffView` shows — happy to add a matching placeholder if you'd prefer.

**2. Reactivity / staleness.** The preview now reads content from the diff models and stays in sync via `onDidChangeContent`, so it tracks in-place model refreshes (index/disk/ref reloads) instead of going stale. The previous approach leaned on `bufferVersions`, which the registry only bumps for the editable buffer model — never git/index models — which was the staleness you caught. I left a comment on the component noting why it intentionally diverges from the file-tab renderer here, to guard against a future refactor reverting it.

Also folded in a small delayed loading spinner (per the Greptile note) so the preview shows a spinner instead of a blank pane while models are still fetching.

arnestrickmann:
Hi, thanks again for opening this PR and for your interest in contributing. We are going to close this PR for now because we decided to track this internally, but I will make sure to notify you once we ship this. 

Thanks again!

PR #2433, opened 2026-06-09
fix(editor): rendered preview stays stale after discarding changes

george-vii:
## Bug
After discarding a change to a file open in the editor's **rendered preview** (markdown/html/svg), the preview keeps showing the discarded content even though the working tree is reverted and the Changes panel is clean. It only refreshes on a tab remount (close + reopen).

## Repro
1. Open a Markdown file in the rendered preview (eye view).
2. Edit it so it appears under Changes.
3. Discard the change while the preview is open.
4. Changes panel goes clean, but the preview still renders the old line.

## Root cause
Disk-driven buffer reloads (`applyDiskUpdate`, `reloadFromDisk`) update the Monaco buffer model inside a `reloadingFromDisk` guard. The buffer's `onDidChangeContent` listener intentionally bails while that guard is set (so a disk reload isn't treated as a user edit) — but that also skips bumping `bufferVersions`. The rendered preview's only reactive dependency is `bufferVersions`, so it never re-renders. The Monaco *source* view updates fine (bound to the model directly), which is why only preview mode is affected and a remount "fixes" it.

## Fix
Bump `bufferVersions` explicitly in both reload paths after applying the reloaded content, decoupled from the dirty/autosave logic the listener guards. Covers discard, watcher-surfaced external edits, branch switches, and the conflict-dialog "Accept Incoming" path.

## Testing
- New regression test: a disk reload updates a clean open buffer and bumps `bufferVersions`.
- typecheck, lint, format, and the node test project pass.
- Verified in-app: discarding a change to an open Markdown preview now updates immediately.

### DIFF
--- a/apps/emdash-desktop/src/renderer/lib/monaco/monaco-model-registry.test.ts
+++ b/apps/emdash-desktop/src/renderer/lib/monaco/monaco-model-registry.test.ts
@@ -5,6 +5,7 @@ import { MonacoModelRegistry } from './monaco-model-registry';
 const rpcState = vi.hoisted(() => ({
   indexContent: 'base' as string | null,
   refContent: 'base' as string | null,
+  diskContent: 'base' as string,
 }));
 
 vi.mock('@renderer/lib/ipc', () => ({
@@ -13,7 +14,7 @@ vi.mock('@renderer/lib/ipc', () => ({
       fs: {
         readFile: vi.fn(async () => ({
           success: true,
-          data: { content: 'base', truncated: false, totalSize: 4 },
+          data: { content: rpcState.diskContent, truncated: false, totalSize: 4 },
         })),
       },
       git: {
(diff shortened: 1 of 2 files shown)


george-vii:
Addressed the review feedback:

- **`reloadFromDisk` untested** — added a regression test for the conflict-dialog path: a dirty buffer plus an external disk change records a conflict, and `reloadFromDisk` ("Accept Incoming") replaces the buffer content, clears dirty/conflict state, and bumps `bufferVersions`. Verified the test fails if the `reloadFromDisk` bump is removed.
- **Unconditional bump in `reloadFromDisk`** — left as-is. This path is only reachable from the conflict dialog, and `applyDiskUpdate` only records a conflict when buffer and disk content diverge, so a same-content bump is a rare edge (the user manually syncing the buffer before clicking Accept). The only `bufferVersions` consumers are pure render subscriptions, so the cost would be a single extra re-render.
- **Two `runInAction` blocks in `applyDiskUpdate`** — left as-is. The version bump and the `dirtyUris` cleanup sit in different conditional branches, and no observer reads both observables together, so there's no observable intermediate state. Happy to merge them if you'd prefer it stylistically.

arnestrickmann:
Hi @george-vii, thank you for opening the PR and for your interest in contributing! 
We are working on improving this internally, so closing this PR - appreciate it and thank you again!

PR #2473, opened 2026-06-10
feat(diff-view): unified changes view (split/unified toggle)

snirye:
## Summary
Adds a Unified view to the Changes panel that lists every file differing between the working tree and the PR base (or default branch when no PR exists), under a single read-only list. The existing 3-section Split view is preserved and the user toggles between them via a new icon in the panel header. Persisted per-task via app settings.

- `getMergeBase` RPC + new `unified` diff group plumbed through tab manager / lifecycle / file renderer (orig = merge-base commit, modified = working-tree buffer)
- `UnifiedChangesStore` derives base ref (PR base ?? default branch), fetches `merge-base(base, HEAD)`, and produces one row per file via existing `getChangedFiles`
- Refresh piggy-backs on the same git/fs/ref events as the split-view status pipeline (no new poll loops)
- Diff is read-only — toolbar labels the source as `All changes (read-only)`

## Test plan
- [ ] Open a task with an open PR, modify a file with both committed-unpushed and uncommitted edits → flip to Unified → file appears once with combined diff
- [ ] Open a task with no PR, ensure Unified compares against the project's default branch
- [ ] Toggle between Split and Unified — split-mode selections preserved across mode flips
- [ ] Switch list/tree mode while in Unified — persists independently from split sections
- [ ] Empty states render: no base configured, no merge-base (orphan branch), no diff vs base

Spec: `docs/superpowers/specs/2026-06-10-unified-changes-view-design.md`
Plan: `docs/superpowers/plans/2026-06-10-unified-changes-view.md`

### DIFF
--- a/apps/emdash-desktop/src/main/core/git/controller.ts
+++ b/apps/emdash-desktop/src/main/core/git/controller.ts
@@ -89,6 +89,18 @@ export const gitController = createRPCController({
     }
   },
 
+  getMergeBase: async (projectId: string, workspaceId: string, base: GitObjectRef) => {
+    try {
+      const env = resolveWorkspace(projectId, workspaceId);
+      if (!env) return err({ type: 'not_found' as const });
+      const sha = await env.git.getMergeBase(base);
+      return ok({ sha });
+    } catch (e) {
+      log.error('gitCtrl.getMergeBase failed', { projectId, workspaceId, base, error: e });
+      return err({ type: 'git_error' as const, message: String(e) });
+    }
+  },
+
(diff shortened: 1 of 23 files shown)


arnestrickmann:
@snirye Thanks a lot for opening this PR and for the thought and effort you put into it, we really appreciate your contribution to Emdash.

The rework of the right sidebar is definitely on our list, but we can’t communicate a concrete ETA for it yet.

That said, your ideas and impulses are very valuable to us. We’ll definitely take a close look at them as we continue working on the design of the sidebar, and especially also at your thoughts around the Git flow. Thanks again for taking the time to put this together.

PR #2510, opened 2026-06-13
fix: detect remote agents using the execution context's platform, not the client's

### LINKED ISSUE #2474: [bug]: Remote SSH agent detection runs Windows `where` on Linux hosts (client platform used instead of remote)
### Bug Description

When Emdash runs on **Windows** and connects to a **remote SSH project on a Linux server**, installed agents (Claude Code, opencode, ...) are never detected, even though they are on the remote `PATH` and work fine in a plain SSH terminal. Clicking Install in the agents menu ends with **"The agent was not detected after installation."**

Root cause: the command used to resolve agent binaries is chosen from the **local** platform instead of the execution context's platform:

```ts
// apps/emdash-desktop/src/main/core/dependencies/probe.ts
const RESOLVE_CMD = process.platform === 'win32' ? 'where' : 'which';
```

For remote projects, `DependencyManager` is created with `SshExecutionContext` (`getDependencyManager(connectionId)` in `dependency-manager.ts`), so `resolveCommandPath()` executes `RESOLVE_CMD` **on the remote host**. With a Windows client this means running `where claude` on the Linux server — `where` does not exist there, the probe fails, and every agent is reported as missing. The post-install probe fails the same way, producing `not-detected-after-install`.

macOS/Linux clients are unaffected (`RESOLVE_CMD` = `which`), which is probably why this hasn't been caught. Possibly related to generalaction/emdash#2140 (WSL detection on Windows).

### Steps to Reproduce

1. Run Emdash on Windows.
2. Add a remote SSH project pointing to a Linux server that has `claude` (or `opencode`) installed and on `PATH` for non-interactive shells.
3. Open the agent selector — no agents are detected.
4. Click Install → installation runs, then fails with "The agent was not detected after installation."

* **Actual:** `where claude` is executed on the Linux host, fails with `where: command not found`, agents are reported as not installed.
* **Expected:** the resolve command should match the **remote** platform (`which`/`command -v` on POSIX hosts), so agents installed on the remote are detected regardless of the client OS.

### Suggested Fix

Pick the resolve command per execution context rather than from `process.platform`, e.g. expose the target platform on `IExecutionContext` (SSH contexts are POSIX → `command -v`; local context keeps the current behavior).

### Workaround

On the Linux server, create an executable shim `~/.local/bin/where`:

```sh
#!/bin/sh
exec which "$@"
```

After that, detection and the version probe work immediately (verified with Emdash 1.1.32, Windows client → Ubuntu 24.04 host, Claude Code 2.1.170).

serhiizghama:
## Problem

Closes #2474.

When Emdash runs on **Windows** and connects to a **remote SSH project on a Linux server**, installed agents (Claude Code, opencode, …) are never detected even though they are on the remote `PATH`. Install ends with *"The agent was not detected after installation."*

Root cause: the binary-resolve command was chosen from the **local client** platform:

```ts
// probe.ts
const RESOLVE_CMD = process.platform === 'win32' ? 'where' : 'which';
```

But `resolveCommandPath()` runs that command through `ctx.exec()`, which for remote projects is an `SshExecutionContext` executing on the **remote host**. A Windows client therefore ran `where claude` on the Linux server — `where` doesn't exist there, so every probe failed and all agents were reported missing. The post-install probe failed the same way (`not-detected-after-install`). macOS/Linux clients were unaffected because `which` exists on the remote, which is why this went unnoticed.

## Solution

Pick the resolve command from the platform the context **executes on**, not the local client:

- Add `readonly isWindows` to `IExecutionContext`.
- `LocalExecutionContext` → `process.platform === 'win32'` (preserves existing behavior).
- `SshExecutionContext` → `false` (SSH remotes are POSIX).
- `probe.ts` resolves with `ctx.isWindows ? 'where' : 'which'`.

Local macOS/Linux and local Windows behavior is unchanged; the only behavioral change is that a remote POSIX host is now probed with `which` regardless of the client OS.

## Testing

- New `probe.test.ts` covers POSIX vs Windows resolution and the regression directly: a POSIX SSH context resolves with `which` even when the client is Windows.
- Updated existing `IExecutionContext` test mocks for the new required field.
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run format:check` all pass; affected suites (probe, dependency-manager, execution-context, terminals, agent-hooks, git) green.

### DIFF
--- a/apps/emdash-desktop/src/main/core/agent-hooks/claude-trust-service.test.ts
+++ b/apps/emdash-desktop/src/main/core/agent-hooks/claude-trust-service.test.ts
@@ -281,6 +281,7 @@ describe('ClaudeTrustService', () => {
     const ctx: IExecutionContext = {
       root: undefined,
       supportsLocalSpawn: false,
+      isWindows: false,
       exec: vi.fn().mockImplementation(async (command: string, args: string[] = []) => {
         if (command === 'sh') {
           return { stdout: '/home/remote-user', stderr: '' };
--- a/apps/emdash-desktop/src/main/core/agent-hooks/cursor-trust-service.test.ts
+++ b/apps/emdash-desktop/src/main/core/agent-hooks/cursor-trust-service.test.ts
@@ -67,6 +67,7 @@ function makeCtx(): IExecutionContext {
   return {
     root: undefined,
     supportsLocalSpawn: false,
+    isWindows: false,
     exec: vi.fn().mockImplementation(async (command: string) => {
       if (command === 'sh') return { stdout: '/home/remote-user', stderr: '' };
       return { stdout: '', stderr: '' };
--- a/apps/emdash-desktop/src/main/core/agent-hooks/hook-config.test.ts
+++ b/apps/emdash-desktop/src/main/core/agent-hooks/hook-config.test.ts
@@ -13,6 +13,7 @@ vi.mock('@main/core/dependencies/probe', () => ({
 function makeExecutionContext(): IExecutionContext {
   return {
     supportsLocalSpawn: false,
+    isWindows: false,
     exec: vi.fn(async () => ({ stdout: '', stderr: '' })),
     execStreaming: vi.fn(async () => {}),
     dispose: vi.fn(),
--- a/apps/emdash-desktop/src/main/core/agent-hooks/workspace-trust-service.test.ts
+++ b/apps/emdash-desktop/src/main/core/agent-hooks/workspace-trust-service.test.ts
@@ -19,6 +19,7 @@ function makeCtx(): IExecutionContext {
   return {
     root: undefined,
     supportsLocalSpawn: false,
+    isWindows: false,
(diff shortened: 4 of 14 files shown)


arnestrickmann:
Thank you for digging into this and for the clear write-up. 

Since this was opened, `main` has had a pretty large refactor in this area: dependency probing moved into the shared dependency runtime, so this PR now conflicts on the old `probe.ts` / `hook-config.test.ts` paths. I don’t think we should resolve that by bringing the old files back.

Would you be willing to rebase onto latest `generalaction/main` and reapply the same idea in the new structure? The important invariant is exactly what you identified: binary resolution should use the platform of the execution host, not the desktop client’s `process.platform`.

Thank you 🙏

arnestrickmann:
Quick follow up: 
@serhiizghama, we've pushed a fix for this. Thank you for initially flagging this issue and then immediately taking a spin at fixing it. Appreciate it! Closing this PR.

PR #2514, opened 2026-06-13
fix(tasks): refresh open windows when a task is archived headlessly

dfox288:
### Description

Archiving a task only refreshed the sidebar when the archive was initiated from the renderer. The store updated itself optimistically inside its own `archiveTask` flow, so any archive initiated from the **main process** — e.g. the automations runtime tearing a task down — left the open window showing a stale, still-active task until a hard reload.

Fix: `taskService.archiveTask` now emits a `task:archived` event after the archive operation completes, and the task manager store subscribes and mirrors the same archive mutation it already performs for GUI-initiated archives (set `archivedAt`, release the conversation/terminal registries, transition the task store to dry-unprovisioned). The handler is guarded idempotent (`archivedAt` already set ⇒ no-op), so renderer-initiated archives that already updated optimistically are not double-processed.

Implementation notes:
- New `taskArchivedChannel` (`task:archived`) in `shared/core/tasks/taskEvents.ts`, carrying `{ taskId, projectId }`.
- Emitted once, unconditionally, after the archive op in `taskService.archiveTask`.
- Renderer handler filters by `projectId` and is a no-op for unknown/unregistered/already-archived tasks.

### Related issues

None — found while building on top of the archive flow.

### Testing

- `pnpm run typecheck` — clean (on top of current `main`)
- `pnpm run format:check` — clean
- `pnpm run lint` — clean on touched files
- `pnpm exec vitest run src/main/core/tasks/task-service.test.ts` — new test passes (asserts the op runs and the event is emitted, in that order)
- Manual: archived a task from the main process while a window was open; sidebar updated live without a reload. Confirmed a normal GUI archive still works and is not double-applied.

The renderer handler mirrors the existing optimistic mutation and was validated manually; there is no pre-existing unit-test harness for the task-manager MobX store, so I did not add one for that half.

### DIFF
--- a/apps/emdash-desktop/src/main/core/tasks/task-service.test.ts
+++ b/apps/emdash-desktop/src/main/core/tasks/task-service.test.ts
@@ -1,0 +1,82 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { events } from '@main/lib/events';
+import { taskArchivedChannel } from '@shared/core/tasks/taskEvents';
+import { archiveTask as archiveTaskOp } from './operations/archiveTask';
+
+// task-service pulls in the full main-process graph at import time; mock the
+// heavy singletons / db so the module loads in a plain node test, and stub
+// every task operation so we can drive the service in isolation.
+vi.mock('@main/lib/events', () => ({
+  events: { emit: vi.fn(), on: vi.fn(() => vi.fn()) },
+}));
+vi.mock('@main/lib/logger', () => ({ log: { error: vi.fn(), warn: vi.fn(), info: vi.fn() } }));
+vi.mock('@main/db/client', () => ({ db: {} }));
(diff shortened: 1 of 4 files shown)


dfox288:
Thanks for the review. Both points addressed in the latest push:

- **Failure-case test**: added a test asserting that when the archive operation throws, `events.emit` is never called (so open windows never hide a task that's still alive in the DB).
- **MobX intermediate state**: collapsed the two `runInAction` blocks in the headless handler into a single action (with `_releaseTaskRegistries` inside it), so reactions never observe `archivedAt` set before the store has transitioned. The GUI path stays split only because it must await the RPC between steps.

arnestrickmann:
Thanks for the PR, and sorry for the delay here, @dfox288 

This is not blocked on anything from your side right now. Thanks also for addressing the review feedback.

We’ll take a final look from our side and follow up shortly.

jschwxrz:
closing this as outdated

PR #2540, opened 2026-06-16
fix(browser): make tabs not always reload in the in app browser

janburzinski:
### Description
- in app browser tabs not reloading on refocus

https://streamable.com/4hdjuw

<details>
<summary>Checklist</summary>

</details>


janburzinski:
@greptileai

arnestrickmann:
LGTM — thank you for the fix. Merging this.

One thing I noticed while testing: when I switch away from a task that has a browser tab open, then come back and re-enter that tab, the browser reloads.

I think we should prevent this self-reload behavior entirely if possible, since it can interrupt the user’s state. @janburzinski - I would propose to do this in a separate PR.

arnestrickmann (approved):

Issue #2458, opened 2026-06-09
[feat]: Add automatic port forwarding for SSH-connected projects


HugoMarchant:
### Feature Summary

Add automatic port forwarding for emdash projects accessed over SSH connections.

When a dev server or agent-started process listens on a remote port, emdash should detect it, create a local port forward automatically, and surface a clickable local URL in the UI, similar to VS Code Remote SSH.

### Problem or Use Case

I often work on projects over SSH where dev servers run on the remote machine, usually binding to localhost, 127.0.0.1, or 0.0.0.0.

Currently, to preview those apps locally, I need to manually identify the remote port and create an SSH tunnel, for example:

ssh -L 5173:localhost:5173 user@remote-host

This adds friction when using emdash remotely, especially when agents start servers automatically or when a project chooses a dynamic port. It would be helpful if emdash could detect and expose remote dev server ports automatically so the local preview workflow feels seamless.

### Proposed Solution

When an emdash project is accessed over SSH, emdash should detect ports opened by processes in that project/session and automatically create local port forwards.

The ideal UX would be:

I open an emdash project over SSH.

I or an agent runs a dev server, such as:

pnpm dev
The app starts on a remote port, such as 5173.
emdash detects the port and creates a local forward.

emdash shows a clickable local URL in the UI, for example:

Port 5173 forwarded to http://localhost:5173

The port forwarding UI could show:

remote port
local forwarded URL
forwarding status
associated project/session/process, if available
actions to open, copy URL, stop forwarding, or restart forwarding

It should also handle common edge cases:

If the local port is already in use, emdash should choose another available local port and clearly show the mapping.
Forwarded ports should be cleaned up when the SSH session/project closes.

Port detection could come from a few sources:

parsing terminal/PTY output for URLs like http://localhost:3000, http://127.0.0.1:5173, or http://0.0.0.0:8000
checking listening ports on the remote machine for project-owned processes

### Alternatives Considered

I currently work around this by manually creating SSH tunnels with ssh -L, but this is easy to forget and does not work well when the port is chosen dynamically or when an agent starts the server.

Another workaround is to bind the remote dev server to a public interface, but that is less safe and less convenient than local forwarding.

### Additional Context

VS Code Remote SSH provides a similar workflow by detecting remote ports, forwarding them locally, and exposing them in the UI. Having this in emdash would make remote development and agent-driven preview workflows much smoother.

[...]

PR #2544, opened 2026-06-16
feat: add ssh port forwarding for remote previews

jschwxrz:
- add automatic ssh port forwarding when terminal output detects local preview urls
- replace host rewriting with local loopback urls backed by ssh tunnels
- add preview server state for direct and forwarded previews with status and lifecycle cleanup
- surface preview server controls for open, copy, stop, restart and manual forwarding
- remove legacy host preview events and dev server store/pills

#2458


arnestrickmann:
Very nice, works great.

One thing I noticed:

When I click a URL for a started dev server in the terminal drawer and open it in the Emdash browser, the URL includes a trailing ) at the end. Because of that, the dev server doesn’t open correctly until I manually remove the ) from the URL.

jschwxrz:
@greptileai

PR #2590, opened 2026-06-19
fix(port-forwards): try IPv6 loopback when the IPv4 target refuses

### LINKED ISSUE #2589: SSH port forward only targets 127.0.0.1 — misses dev servers bound to localhost/::1
Following up on the SSH port forwarding from #2544 (great feature, using it daily). Hit a sharp edge worth fixing.

### Symptom
A dev server started without an explicit host is detected by Emdash, but the auto-forward never connects. Passing `--host 0.0.0.0` worked around it, but that exposes the server on every interface (including the Tailscale/LAN IP), which we don't want.

### Root cause
Vite and friends default to `localhost` when no host is passed. On Node >= 17 `localhost` resolves to IPv6 `::1` first, so the server binds only to `[::1]:5173`. The tunnel in `port-forward-tunnel.ts` hardcodes the remote destination to IPv4:

```ts
const REMOTE_TARGET_HOST = '127.0.0.1';
// ...
client.forwardOut(LOCAL_BIND_HOST, 0, REMOTE_TARGET_HOST, options.remotePort, ...)
```

So Emdash dials `127.0.0.1:5173` on the remote where nothing is listening, and the forward fails. `--host 0.0.0.0` only worked because `0.0.0.0` also covers the IPv4 loopback. Note that the detected host (`terminal-url-detector.ts` already parses `localhost` vs `127.0.0.1`) is discarded for SSH, since `forwardOut` always targets `127.0.0.1`.

### Workaround (our side)
`vite --host 127.0.0.1` binds the IPv4 loopback only. Verified: `127.0.0.1:5173 -> 200`, Tailscale IP not reachable (nothing over-exposed), and it lands exactly where Emdash forwards. Downside: it has to be set per project.

### Suggested fix
Forward to the detected hostname rather than a hardcoded IPv4 literal: pass `localhost` to `forwardOut` and let the remote sshd resolve the family, exactly like `ssh -L 5173:localhost:5173` does (that's why the manual tunnel works regardless of IPv4/IPv6). For full robustness, a dual-stack attempt (try `::1`, fall back to `127.0.0.1`, or vice versa) would cover servers bound to only one family.

Happy to send a PR for this if it sounds right.

fiorelorenzo:
### Description

The SSH port-forward tunnel dialed a hardcoded `127.0.0.1` on the remote (`REMOTE_TARGET_HOST` in `port-forward-tunnel.ts`). A dev server started on the default `localhost` host resolves to IPv6 `::1` first on Node >= 17, so it commonly binds only to `[::1]`. The forward then connects to `127.0.0.1`, where nothing is listening, and the preview never loads, even though the URL is detected correctly.

This change makes the remote target dual-stack: per connection, try `127.0.0.1` first and fall back to `::1` if the remote refuses, forwarding through whichever loopback family the server actually listens on. IPv4-only servers keep working exactly as before (first attempt succeeds); `localhost`/IPv6-only servers now work without forcing `--host`.

The fallback is per connection and order-preserving, so there is no stale cached target and no behavior change for servers reachable on `127.0.0.1`.

### Related issues

Fixes #2589

### Testing

From the repo root with the pinned toolchain (Node 24.14.0, pnpm 10.28.2):

- `pnpm run format:check` - pass
- `pnpm run lint` - pass
- `pnpm run typecheck` - pass (after `pnpm run build` to emit workspace package types)
- `pnpm --filter ./apps/emdash-desktop run test` - pass (269 files, 1937 tests), including a new `port-forward-tunnel` test that asserts the IPv4 target is tried first and the connection falls back to `::1` when IPv4 refuses

Note: `pnpm run test` also has an unrelated pre-existing failure in `packages/core` (`git-repository` default-branch test) on machines where git's `init.defaultBranch` is not set to `main`; it is independent of this change.

N/A (main-process networking change).

### DIFF
--- a/apps/emdash-desktop/src/main/core/port-forwards/port-forward-tunnel.test.ts
+++ b/apps/emdash-desktop/src/main/core/port-forwards/port-forward-tunnel.test.ts
@@ -71,6 +71,40 @@ function makeRejectingProxy(error: Error) {
   };
 }
 
+function makeFamilyAwareProxy(reachableHost: string) {
+  const calls: Array<{ remoteHost: string; remotePort: number }> = [];
+
+  return {
+    calls,
+    proxy: {
+      get isConnected() {
+        return true;
+      },
+      get client() {
+        return {
+          forwardOut(
+            _sourceHost: string,
+            _sourcePort: number,
+            remoteHost: string,
+            remotePort: number,
+            callback: (error: Error | undefined, channel: ClientChannel) => void
+          ) {
+            calls.push({ remoteHost, remotePort });
+            if (remoteHost === reachableHost) {
+              callback(undefined, new EchoChannel() as unknown as ClientChannel);
+              return;
+            }
+            callback(
+              new Error('(SSH) Channel open failure: Connection refused'),
+              undefined as unknown as ClientChannel
+            );
+          },
+        } as SshClientProxy['client'];
+      },
+    } satisfies Pick<SshClientProxy, 'client' | 'isConnected'>,
+  };
+}
+
 function listen(server: net.Server): Promise<number> {
   return new Promise((resolve, reject) => {
     server.once('error', reject);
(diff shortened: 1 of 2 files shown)


fiorelorenzo:
Thanks for the review. Addressed both points in ea65cac:

1. **Fallback now only fires on a connect failure.** ssh2 attaches the RFC 4254 channel-open reason code to the `forwardOut` error, so the retry is gated on `reason === SSH_OPEN_CONNECT_FAILED` (2). Administratively-prohibited, resource-shortage and dropped-session errors are surfaced immediately instead of being masked behind a pointless second dial.
2. **First error is preserved.** When every loopback family fails, `onConnectionError` now receives the first (primary IPv4) error rather than the last one.

Added tests for both: a non-connect failure is surfaced without any `::1` fallback dial, and the first error wins when both families refuse. `format`, `lint`, `typecheck` and the desktop test suite pass locally (port-forward-tunnel: 6 tests).

jschwxrz:
nice, thank you!

jschwxrz (approved):

Issue #2580, opened 2026-06-18
[bug]: SSH remote projects accumulate orphaned detached tmux agent sessions (never reaped on conversation close)


fiorelorenzo:
### Bug Description

On SSH-connected remote projects, each conversation's agent runs in a **detached** tmux session on the remote host (`emdash-<base64(projectId:taskId:conversationId)>` running `claude --session-id … --dangerously-skip-permissions`). Closing a conversation tab only **detaches** the PTY (intentional since [#2281](<https://github.com/generalaction/emdash/issues/2281>), "preserve tmux sessions on close"); it does not stop the underlying session. Since nothing ever reaps them, these detached sessions and their `claude` processes accumulate **indefinitely** on the remote box, surviving SSH disconnects, host sleep, and Emdash restarts.

After a few days of normal open/close churn within a single task, I found **12 live** `claude` **processes** from one project still running on the remote host (only 2 attached; the other 10 idle 22–51h).

Secondary, more painful impact on **headless Linux** remotes: Claude Code there cannot use a system keyring (related: generalaction/emdash#1875), so all these long-lived `claude` processes share a single `~/.claude/.credentials.json`. Concurrent OAuth refresh-token rotation across them invalidates each other's refresh tokens, producing **frequent unexpected re-login prompts**. So the leak is not just resource buildup, it actively breaks auth on remote Linux.

### Steps to Reproduce

1. Connect Emdash to a remote project over SSH (remote Linux host with tmux).
2. In a task, start an agent (Claude) conversation. On the remote, `tmux ls` now shows an `emdash-…` session.
3. **Close** the conversation tab (do not use Stop or Delete).
4. Open a new conversation; repeat the open/close cycle a few times, optionally across days and Emdash restarts.
5. On the remote host run `tmux ls` and `pgrep -af 'claude --session-id'`.
6. Observe detached `emdash-*` sessions and their `claude` processes pile up and are never cleaned up.

### Actual vs Expected Behavior

**Actual:** Detached remote tmux sessions and `claude` processes accumulate unbounded. They are only removed by an explicit Stop/Delete, or by `destroyAll` (which iterates the current process's in-memory `knownSessionIds`). There is no reconciliation against `tmux list-sessions`, so sessions orphaned by closed tabs or by previous app runs are never discovered or reaped. On headless Linux this also causes auth churn / repeated logouts via the shared credentials file.

[...]

PR #2591, opened 2026-06-19
fix(pty): reap setsid()-detached descendants on kill

### LINKED ISSUE #2110: [bug]: PTY kill() doesn't reach setsid()-detached descendants
#689 sends `SIGHUP` to the PTY foreground process group, which reaches the interactive shell but not descendants that have called `setsid()`. That leaves `ts-checker-rspack-plugin` workers (from rspack/rsbuild dev servers), the `watchman` daemon, and `sudo`'d children running after task deletion. I've seen them survive for days across emdash worktree sessions.

fiorelorenzo:
## Problem

`LocalPtySession.kill()` only signals the PTY's **foreground process group** (`process.kill(-pid, …)`). Descendants that call `setsid()` form their own session/process group and escape that signal, so they survive task deletion. In practice this leaves the **watchman** daemon, **ts-checker-rspack-plugin** workers (rsbuild/rspack dev servers), and `sudo`'d children running — often for days, frequently holding ports — across worktree sessions.

Fixes #2110.

## Fix

Snapshot the descendant pid tree **before** signalling — once the shell dies the survivors are reparented to `init` and the parent links that identify them are gone — then SIGTERM the foreground group *and* every descendant individually, escalating to SIGKILL after the existing 2s grace.

- New `process-tree.ts`: `parsePidPpidPairs()` + cycle-safe `collectDescendantPids()` (pure), and `collectLocalDescendantPids()` which snapshots via `ps -A -o pid=,ppid=` (supported by both Linux/procps and macOS/BSD `ps`). Best-effort: returns `[]` if `ps` is unavailable, so the kill path never throws.
- `local-pty.ts`: snapshot + `signalProcessTree()` for the SIGTERM and SIGKILL passes.
- **Windows is unchanged** — node-pty's job object already tears down the tree there.

The helper is transport-agnostic and is reused by the companion remote-tmux reaping fix (#2580).

## Testing

`pnpm run format && pnpm run lint && pnpm run typecheck` clean; full `node` Vitest project green. New unit tests cover the parser, the BFS (multi-root, cycle-safe), and the kill path (descendant SIGTERM/SIGKILL, single snapshot reused, Windows no-op).

### DIFF
--- a/apps/emdash-desktop/src/main/core/pty/local-pty.test.ts
+++ b/apps/emdash-desktop/src/main/core/pty/local-pty.test.ts
@@ -1,9 +1,19 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { LocalPtySession } from './local-pty';
+import { collectLocalDescendantPidsAsync } from './process-tree';
 
 vi.mock('node-pty', () => ({
   spawn: vi.fn(),
 }));
+
+vi.mock('./process-tree', () => ({
+  collectLocalDescendantPidsAsync: vi.fn(() => Promise.resolve([])),
+}));
+
+/** Let the async descendant snapshot (a resolved promise + its .then) settle. */
+async function flushSnapshot(): Promise<void> {
+  for (let i = 0; i < 5; i++) await Promise.resolve();
+}
 
 describe('LocalPtySession', () => {
   type MockPtyProcess = ConstructorParameters<typeof LocalPtySession>[1];
@@ -19,8 +29,20 @@ describe('LocalPtySession', () => {
     });
   }
 
+  /** Register an onExit handler and return a trigger for the captured callback. */
+  function captureExit(): () => void {
+    let exitHandler: ((info: { exitCode: number; signal: number }) => void) | undefined;
+    vi.mocked(mockProc.onExit).mockImplementation((handler) => {
+      exitHandler = handler;
+      return { dispose: vi.fn() };
+    });
+    pty.onExit(() => {});
+    return () => exitHandler?.({ exitCode: 0, signal: 0 });
+  }
+
   beforeEach(() => {
     vi.clearAllMocks();
+    vi.mocked(collectLocalDescendantPidsAsync).mockResolvedValue([]);
     vi.useFakeTimers();
 
     mockProc = {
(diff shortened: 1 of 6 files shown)


fiorelorenzo:
Thanks @greptile-apps — both findings were valid and are addressed in the latest push.

**P1 — `onExit` cancelled the SIGKILL before it reached detached descendants.** Correct, and it defeated the PR's purpose: the shell exits within ms of SIGTERM, `onExit` cleared the timer, and daemons that ignore SIGTERM (watchman, ts-checker workers) never got SIGKILL. Fixed by splitting the escalation into two timers:
- the **foreground-group** SIGKILL stays cancellable on `onExit` (the group dies with the shell, and a late SIGKILL could hit a reused pid);
- the **descendant** SIGKILL runs on an independent timer that `onExit` does **not** cancel, since setsid()-detached descendants outlive the shell.

Added a regression test (`kill() still SIGKILLs detached descendants even when the shell exits first`) that combines a quick shell exit with a live descendant and asserts the group SIGKILL is suppressed while the descendant one still fires.

**P2 — `execFileSync` blocked the Electron main-process event loop.** Switched the snapshot to async `execFile` (`collectLocalDescendantPidsAsync`). It still runs *before* any signal — so the descendant tree is captured while the shell is alive and its parent links are intact — but it no longer blocks the event loop. `kill()` dispatches the snapshot and signals on resolution; `this.proc.kill()` moved into that path too.

fiorelorenzo:
Thanks for the re-review and the catch on `descendantKillTimer`.

Agreed it's benign, and I've left it as-is for now: `LocalPtySession` has no `dispose()` in the `Pty` interface, and `onExit` — the only teardown hook — must *not* clear `descendantKillTimer` (that's the P1 fix). The timer self-clears once it fires, so the worst case is the session being retained for ≤2s after the process is gone (bounded, not a leak), and `signalPids` swallows `ESRCH` on the already-dead pids.

If a `dispose()` is ever added to the session lifecycle, that's the natural place to `clearTimeout` both timers — happy to wire it up then.

jschwxrz:
thank you for the pr! i pushed two follow-up commits to reduce pid or pgid reuse risk and clean up the code organization.

a known limitation of this pr is that it only covers local node pty sessions and not the ssh path. that is fine for now because the ssh path is going to change soon anyways. we are planning to run a daemon process on remote machines which should give us a better ownership boundary there and let us apply the same model remotely.

longer term we should probably look at kernel owned process containment as well. linux has cgroups and windows has job objects, which would let us kill owned processes more deterministically. macos does not have an equivalent so this approach is still the best fallback available there.

jschwxrz (approved):

PR #2592, opened 2026-06-19
fix(ssh): reap orphaned remote tmux sessions and their dev servers

### LINKED ISSUE #2580: [bug]: SSH remote projects accumulate orphaned detached tmux agent sessions (never reaped on conversation close)

fiorelorenzo:
## Problem

SSH remote agents run in detached tmux sessions on the host (`emdash-<base64url(sessionId)>`). Closing a conversation tab only **detaches** (preserve-on-close, #2281) — intentional, for resumability — but nothing ever reconciles tmux against live state, so:

- sessions for **deleted** conversations/terminals, and **every** session left behind by an app restart (in-memory tracking lost), accumulate on the host indefinitely; and
- `tmux kill-session` only SIGHUPs the pane processes, so dev servers (vite/metro/expo, watchman) that double-fork / `setsid()` survive as **port-holding orphans**.

Fixes #2580.

## Fix

- **Reconcile on SSH project mount** (`RemoteTmuxReaperService`, wired to `projectManager` `projectOpened`): list the host's `emdash-*` sessions and reap the orphans — those whose leaf entity (conversation/terminal) is gone from the DB. Safety constraints so a tmux server shared by several projects/hosts is never mis-reaped:
  - scoped to the **mounted project only** (the `projectId` is decoded from the session name);
  - **lifecycle-script** sessions (setup/run/teardown — they create tmux sessions but have no DB row) are always preserved;
  - **detached-but-still-tracked** sessions are kept, so resumability is unaffected.
  This also handles the app-restart case the issue calls out.
- **`killTmuxSessionTree`**: snapshot the pane process trees (`list-panes` + remote `ps`) **before** `kill-session`, then SIGTERM/SIGKILL the detached descendants. Wired into the explicit SSH Stop/Destroy paths and the reconciler. Reuses the `process-tree` helper from #2110.
- DB-dependent reconciliation is split from the DB-free kill helpers so the SSH providers don't transitively load the database.

## Stacked on #2591

This branch is stacked on #2591 (the `process-tree` helper). Until that merges, the diff here includes its commit — **review only the second commit** (`fix(ssh): …`). Once #2591 lands I'll rebase this onto `main`.

## Testing

`pnpm run format && pnpm run lint && pnpm run typecheck` clean; full `node` Vitest project green (255 files / 1874 tests). New unit tests cover session-name decode round-trip, `list`/tree-kill ordering, and orphan selection (live / lifecycle / foreign-project / unparseable all preserved).

### DIFF
--- a/apps/emdash-desktop/src/main/app/shutdown.ts
+++ b/apps/emdash-desktop/src/main/app/shutdown.ts
@@ -1,6 +1,7 @@
 import { app } from 'electron';
 import { agentHookService } from '@main/core/agent-hooks/agent-hook-service';
 import { automationsService } from '@main/core/automations/automations-service';
+import { remoteTmuxReaperService } from '@main/core/pty/remote-tmux-reaper-service';
 import { prSyncScheduler } from '@main/core/pull-requests/pr-sync-scheduler';
 import { stopResourceSampler } from '@main/core/resource-monitor/resource-sampler';
 import { runtimeManager } from '@main/core/runtime/runtime-manager';
@@ -42,6 +43,7 @@ export async function runQuitCleanup(): Promise<void> {
   stopResourceSampler();
   updateService.dispose();
   prSyncScheduler.dispose();
+  remoteTmuxReaperService.dispose();
 
   // critical phase
   const criticalSteps: Array<[string, () => Promise<void>]> = [
--- a/apps/emdash-desktop/src/main/core/conversations/impl/ssh-conversation.ts
+++ b/apps/emdash-desktop/src/main/core/conversations/impl/ssh-conversation.ts
@@ -11,7 +11,8 @@ import { ptySessionRegistry } from '@main/core/pty/pty-session-registry';
 import { resolveSshCommand } from '@main/core/pty/spawn-utils';
 import { openSsh2Pty } from '@main/core/pty/ssh2-pty';
 import { getTerminalColorEnv } from '@main/core/pty/terminal-color-scheme';
-import { killTmuxSession, makeTmuxSessionName } from '@main/core/pty/tmux-session-name';
+import { killTmuxSessionTree } from '@main/core/pty/tmux-reaper';
+import { makeTmuxSessionName } from '@main/core/pty/tmux-session-name';
 import { providerOverrideSettings } from '@main/core/settings/provider-settings-service';
 import type { SshClientProxy } from '@main/core/ssh/lifecycle/ssh-client-proxy';
 import { events } from '@main/lib/events';
@@ -331,7 +332,7 @@ export class SshConversationProvider implements ConversationProvider {
       }
     }
     if (this.tmux) {
-      await killTmuxSession(this.ctx, makeTmuxSessionName(sessionId));
+      await killTmuxSessionTree(this.ctx, makeTmuxSessionName(sessionId));
     }
     this.supervisor.forget(sessionId);
   }
@@ -340,7 +341,9 @@ export class SshConversationProvider implements ConversationProvider {
     const sessionIds = Array.from(this.knownSessionIds);
     await this.detachAll();
     if (this.tmux) {
-      await Promise.all(sessionIds.map((id) => killTmuxSession(this.ctx, makeTmuxSessionName(id))));
+      await Promise.all(
+        sessionIds.map((id) => killTmuxSessionTree(this.ctx, makeTmuxSessionName(id)))
+      );
     }
     for (const sessionId of sessionIds) {
       this.supervisor.forget(sessionId);
--- a/apps/emdash-desktop/src/main/core/pty/remote-tmux-reaper-service.ts
+++ b/apps/emdash-desktop/src/main/core/pty/remote-tmux-reaper-service.ts
@@ -1,0 +1,42 @@
+import type { IDisposable, IInitializable } from '@emdash/shared';
+import { projectManager } from '@main/core/projects/project-manager';
+import type { ProjectProvider } from '@main/core/projects/project-provider';
+import { log } from '@main/lib/logger';
+import { reconcileProjectTmuxSessions } from './tmux-reconcile';
+
+/**
+ * Reaps orphaned `emdash-*` tmux sessions on a remote host when an SSH project
+ * is mounted. These accumulate when conversations/terminals are deleted while
+ * detached, or when the app restarts and loses the in-memory session tracking
+ * that the explicit Stop/Delete paths rely on (issue #2580).
+ *
+ * Runs once per SSH project open, in the background, and never blocks the mount.
+ * Preserve-on-close resumability is untouched: only sessions that no longer map
+ * to any DB entity are removed.
+ */
+export class RemoteTmuxReaperService implements IInitializable, IDisposable {
+  private _unsubscribe: (() => void) | null = null;
+
+  initialize(): void {
+    this._unsubscribe = projectManager.on('projectOpened', (_projectId, provider) => {
+      this.onProjectMounted(provider);
+    });
+  }
+
+  dispose(): void {
+    this._unsubscribe?.();
+    this._unsubscribe = null;
+  }
+
+  private onProjectMounted(provider: ProjectProvider): void {
+    if (provider.type !== 'ssh') return;
+    void reconcileProjectTmuxSessions(provider.ctx, provider.projectId).catch((err) => {
+      log.warn('RemoteTmuxReaperService: reconcile failed', {
+        projectId: provider.projectId,
+        error: String(err),
+      });
+    });
+  }
+}
+
+export const remoteTmuxReaperService = new RemoteTmuxReaperService();
--- a/apps/emdash-desktop/src/main/core/pty/tmux-reaper.test.ts
+++ b/apps/emdash-desktop/src/main/core/pty/tmux-reaper.test.ts
@@ -1,0 +1,121 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { ExecResult, IExecutionContext } from '@main/core/execution-context/types';
+import { makePtySessionId } from '@shared/core/pty/ptySessionId';
+import { killTmuxSessionTree, listEmdashTmuxSessions } from './tmux-reaper';
+import { makeTmuxSessionName } from './tmux-session-name';
+
+type ExecCall = { command: string; args: string[] };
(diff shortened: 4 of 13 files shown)


fiorelorenzo:
Thanks @greptile-apps — all three are fair, addressed in the latest push.

**Pane pids in `reapPids` (recycled-pid risk).** Good catch. `collectSessionDescendantPids` now returns only the escaped descendants, never the pane pids: `kill-session` already signals the panes, so by reap time they are likely dead and their pids may be recycled — whereas the setsid() descendants we are actually after survive `kill-session` and are still alive. When `ps` is unavailable it now returns `[]` (no way to identify the escapees; the panes are kill-session's job) rather than falling back to pane pids. Added a test for the ps-unavailable path.

**`kill -TERM; kill -KILL` with no gap.** Right — back-to-back it was just SIGKILL with extra noise. Dropped it to `kill -KILL`. Graceful shutdown isn't a goal here (the OS frees the ports on death either way), so this matches the intent without the misleading SIGTERM.

**DB query before project filtering.** Fixed — the reconciler now resolves this project's candidate sessions (decode + projectId + lifecycle filter) first and returns early when there are none, so a tmux server shared by several projects no longer triggers a DB round-trip per mount when nothing on the host belongs to the project. Added a test asserting no DB lookup when every listed session is foreign.

Format, lint, typecheck and the full node test project (1876 tests) are green.

arnestrickmann:
Thanks for the detailed implementation and write-up here.

Did you get a chance to test it locally? I’m especially curious about the close-tab case: start an SSH tmux conversation, close the conversation tab without deleting it, restart/reconnect Emdash, and then check whether the remote `emdash-*` tmux session is actually reaped.

It would also be helpful to know exactly how you tested the dev-server cleanup path, e.g. whether you verified that a detached Vite/Metro/Expo process started inside the tmux session no longer survives after Stop/Delete or reconciliation.

Thank you 🙏

fiorelorenzo:
Tested it locally end-to-end — built this branch as a dev Electron app on macOS
(isolated DB), pointed at a real headless-Linux SSH remote with tmux enabled.

**Close-tab + reconcile (your first question).** I started an SSH tmux
conversation — a remote `emdash-<proj:task:conv>` session appears on the host —
then **closed the tab without deleting it**. The remote session is **preserved**
(detached but alive), i.e. preserve-on-close is intact, as intended. I then
planted an orphan `emdash-<sameProject:task:ghostConv>` whose conversation id is
**not** in the DB (the state a deleted/abandoned conversation leaves behind) and
**restarted the app**. On project mount the reconciler:
- **kept** the real conversation's session (still in the DB → resumable), and
- **reaped** the orphan (no DB row).

So to answer your exact sequence directly: close-without-delete + restart
**keeps** the session (it's resumable) — reaping only fires for sessions whose
leaf entity is gone from the DB. Scoping held too: ~21 pre-existing `emdash-*`
sessions from other projects on the same host were left untouched.

**Dev-server cleanup (your second question).** The orphan I planted ran a
`setsid()`'d, port-holding process that ignores SIGHUP *and* SIGTERM — a stubborn
vite/metro/watchman-style child: a ppid-descendant of the pane but in its own
process group, exactly the case that escapes `kill-session`'s SIGHUP. After the
reconciler reaped the session via `killTmuxSessionTree`, the **port was freed** —
the detached dev server was actually killed. I also reproduced the same reaping
standalone at the command level: `tmux list-panes` → `ps` snapshot *before*
`kill-session` → `kill -KILL` the snapshotted descendants. Plain `kill-session`
alone leaves the port held; `killTmuxSessionTree` frees it. SIGTERM wouldn't have
done it (the process ignored it), which is why `reapPids` uses `kill -KILL`.

(For context on the motivation: on that same host I had ~21 leaked `emdash-*`
sessions accumulated from normal use across 3 projects — exactly the unbounded
accumulation this reconciles.)

jschwxrz:
thank you for the pr! it is a pragmatic fix for the ssh tmux leak.

however, I do not really think it is the final model we should go for. the issue is that remote cleanup crosses an unreliable boundary. a task/project is deleted while the ssh host is unreachable or while the project is not mounted. tmux sessions are one visible symptom, but the same leak can apply to worktrees, port forwards and other remote resources.

we should probably go for an some other model (outbox, tombstone or similar). the goal should be to make resource cleanup first class and not add reconciliation for every resource type that lives on the remote.

wdyt?

fiorelorenzo:
Yeah, agreed — this is really a special case of a bigger problem. Reconciling tmux against the DB doesn't generalize to worktrees, port-forwards or anything else that lives remotely, and the actual gap is the one you pointed at: the delete intent and the remote cleanup are split by an unreliable boundary (host down, project not mounted, or the app closed when you hit delete).

I'd be happy to help build the proper version. The way I'd picture an outbox/tombstone model: when you delete a conversation/task/project, you write a durable "cleanup pending" row for each remote resource it owned, in the same transaction as the delete — so the intent can't get lost even if the host is offline. Something like `(connection_id, resource_type, resource_key, attempts, not_before)`, unique per resource so re-enqueueing is a no-op. Then a generic drainer picks those up whenever the host is reachable (on connect/reconnect + a timer) and runs a handler per `resource_type` — `tmux-session`, `worktree`, `port-forward`, etc. — idempotent, with backoff, deleting the row on success. If the host is reachable at delete time you just run it inline and skip the row (fast path).

That flips it around: cleanup becomes first-class and keyed by resource type, and reconciliation drops to a "just in case" safety net instead of being the mechanism. It also lines up with the remote daemon you mentioned on #2591 — the daemon could be what drains the outbox, or own the resources directly; either way the outbox is the durable intent layer. And the descendant-killing bit from this PR (`killTmuxSessionTree`) just becomes the tmux handler, so it's not throwaway.

A few things I'm genuinely unsure about and would defer to you on: which delete boundaries enqueue what (conversation vs task vs project), what to do with tombstones when a host is gone for good (connection deleted), and whether the remote daemon replaces the drainer or just consumes the outbox.

On this PR itself I don't have strong feelings — keep it as a stop-gap for the loudest symptom (the tmux pile-up; I've got ~20 leaked sessions on one box right now) while the real thing lands, or close it and move the reaping into the drainer. Whatever fits your plan. Happy to take a stab at prototyping the outbox + a tmux handler if that's useful.

jschwxrz:
I think this is something we will probably draft and implement internally, as part of the remote daemon process, but thank you for the offer! But that might be some time, so until then, I propose that we merge this PR as a temporary fix for the issue. I have adjusted some things in #2591 that also affect this PR. If you have the time, could you rebase and adjust the tmux fix to the changes in #2591? Otherwise I can take care of it. Thank you!

fiorelorenzo:
Done. Rebased on main and adapted to the #2591 changes. Turned out small: the helpers I use (`parsePidPpidPairs` / `collectDescendantPids`) came through your refactor untouched, so the only real adjustment was moving the reaper's `dispose` into the new `registerQuitHandler`. All green locally, so it should be good to merge whenever works for you. Thanks!

jschwxrz (approved):

PR #2606, opened 2026-06-19
fix: probe remote agents before status reads

jschwxrz:
- restores remote agent probing before remote agent metadata and installation status reads
- keeps getDependencyManager focused on manager access without hidden agent probing
- dedupes concurrent ssh dependency manager creation per connection
- dedupes agent probes by HostDependencyManager identity
- clears cached dependency managers on ssh disconnect, ssh deletion and byoi teardown
- upserts agent installation status events into an empty cache so probe updates can populate the ui


arnestrickmann:
🔥

PR #2613, opened 2026-06-21
Fix stale diff models after branch state changes

### LINKED ISSUE #2576: [bug]: Diff panel shows changes from previous branch after switching branches
### Bug Description

When switching branches within the same task worktree, Emdash's Changed diff can include changes from the previously checked-out branch.

I expect the Changed diff to match `git diff` for the currently checked-out branch. Instead, it appears to combine the current changes with changes made and committed on the previous branch.

This affects workflows using stacked branches/PRs in the same worktree.

### Steps to Reproduce

1. Start a task, allowing Emdash to create its task branch and worktree.
2. Modify a file on the task branch.
3. Click that file in the Git panel to review its diff.
   - The displayed diff is correct.
4. Commit the change.
5. Check out a new branch in the same worktree, for example to create a stacked branch/PR.
6. Modify the same file on the new branch.
7. Run `git diff -- path/to/file` in the terminal and note the current branch's changes.
8. Click the changed file in Emdash's Git panel.

**Actual result**

The Emdash diff includes changes from both the previous branch and the current branch, so it shows more changes than
terminal `git diff`.

**Expected result**

The Emdash Changed diff should display exactly the current index-to-working-tree diff, equivalent to:

```sh
git diff -- path/to/file
```

Previously viewed or committed changes from another branch should not appear.

### emdash Version

v1.1.34

### Operating System

macOS 26

aerial-systems:
Fixes #2576.

## Summary

- refresh `disk://` and `git://` Monaco models when they are re-opened from the pending-eviction cache (`refs === 0`)
- this avoids reusing stale HEAD/staged/disk contents when the same file path is opened again shortly after branch/worktree state changed
- add regression coverage for pending-eviction git and disk/buffer model reuse

## Why

Diff tabs/models are retained for the 60s eviction window after close. If a file is closed and then re-opened on another branch before eviction, `registerDisk` / `registerGit` previously cancelled the eviction and returned the cached model without re-fetching. That can leave the diff panel showing stale content from the prior branch for the same path.

## Tests

```bash
pnpm exec oxfmt src/renderer/lib/monaco/monaco-model-registry.ts src/renderer/lib/monaco/monaco-model-registry.test.ts
pnpm exec oxlint src/renderer/lib/monaco/monaco-model-registry.ts src/renderer/lib/monaco/monaco-model-registry.test.ts
pnpm exec vitest run --project node src/renderer/lib/monaco/monaco-model-registry.test.ts
```

Result: `4 passed`.

### DIFF
--- a/apps/emdash-desktop/src/renderer/lib/monaco/monaco-model-registry.test.ts
+++ b/apps/emdash-desktop/src/renderer/lib/monaco/monaco-model-registry.test.ts
@@ -5,6 +5,7 @@ import { MonacoModelRegistry } from './monaco-model-registry';
 const rpcState = vi.hoisted(() => ({
   indexContent: 'base' as string | null,
   refContent: 'base' as string | null,
+  diskContent: 'base' as string,
 }));
 
 vi.mock('@renderer/lib/ipc', () => ({
@@ -23,7 +24,11 @@ vi.mock('@renderer/lib/ipc', () => ({
       fs: {
         readFile: vi.fn(async () => ({
           success: true,
-          data: { content: 'base', truncated: false, totalSize: 4 },
+          data: {
+            content: rpcState.diskContent,
+            truncated: false,
+            totalSize: rpcState.diskContent.length,
+          },
         })),
       },
       editor: {
@@ -105,6 +110,7 @@ describe('MonacoModelRegistry', () => {
   beforeEach(() => {
     rpcState.indexContent = 'base';
     rpcState.refContent = 'base';
+    rpcState.diskContent = 'base';
   });
 
   it('clears a staged git model when the index no longer contains the file', async () => {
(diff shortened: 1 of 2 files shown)


aerial-systems:
Pushed follow-up `f77cb36` addressing the stale-cache failure paths that Copilot flagged:

- `invalidateModel()` now clears stale `git://` content and marks status `error` if the git RPC fails.
- `invalidateModel()` now clears stale `disk://` content and marks status `error` if disk read fails or returns null content.
- Disk refresh also handles `truncated`/too-large reads by clearing the stale model and setting `too-large` status/size.
- Added regression tests for pending-eviction git refresh failure and disk refresh failure.

Verification rerun:

```bash
pnpm exec oxfmt src/renderer/lib/monaco/monaco-model-registry.ts src/renderer/lib/monaco/monaco-model-registry.test.ts
pnpm exec oxlint src/renderer/lib/monaco/monaco-model-registry.ts src/renderer/lib/monaco/monaco-model-registry.test.ts
pnpm exec vitest run --project node src/renderer/lib/monaco/monaco-model-registry.test.ts
```

Result: `6 passed`.

aerial-systems:
Pushed a second follow-up, `e87b735`, for the Greptile review notes:

- Pending-eviction `disk://` and `git://` cache hits now set `modelStatus` to `loading` before the async refresh, so observers no longer see stale `ready` content during the fetch.
- Disk refreshes that fail or become `too-large` now clear the stale disk mirror without routing a synthetic empty string through `applyDiskUpdate()`.
- Clean pending buffer models are explicitly evicted on unavailable/too-large disk refresh, so a later `registerBuffer()` follows the same seed-from-disk path as a first registration instead of being silently overwritten.
- Added regression coverage for the loading transition and the too-large pending-buffer path.

Verification rerun:

```bash
pnpm exec oxfmt src/renderer/lib/monaco/monaco-model-registry.ts src/renderer/lib/monaco/monaco-model-registry.test.ts
pnpm exec oxlint src/renderer/lib/monaco/monaco-model-registry.ts src/renderer/lib/monaco/monaco-model-registry.test.ts
pnpm exec vitest run --project node src/renderer/lib/monaco/monaco-model-registry.test.ts
```

Result: `1 passed`, `8 tests passed`.

arnestrickmann:
Thanks for opening this PR, @aerial-systems. We’re taking a look now and will test the changes on our side. We’ll get back to you in this thread once we’ve had a chance to verify everything.

jschwxrz:
closing this as outdated

PR #2636, opened 2026-06-23
Fix git inspection errors when adding projects

### LINKED ISSUE #2635: Cannot add project: valid git repo on external volume (/Volumes/*) reported as "not a git repository"
## Summary

emdash refuses to add a project with **"Cannot add project — myapp is not a git repository"**, even though the folder is a fully valid git repository. The repo lives on a secondary/external volume (`/Volumes/Data/...`) rather than under the home directory.

## Steps to reproduce

1. Have a valid git repo located on a non-home volume, e.g. `/Volumes/Data/dev/myapp`
2. In emdash, try to add that folder as a project
3. Toast appears: **"Cannot add project — myapp is not a git repository."**

## Expected

The project is added — it is a git repository.

## Actual

emdash reports it is not a git repository and refuses to add it.

## Proof it IS a git repo

```
$ pwd
/Volumes/Data/dev/myapp

$ git rev-parse --is-inside-work-tree
true

$ git rev-parse --show-toplevel
/Volumes/Data/dev/myapp

$ ls -ld .git
drwxr-xr-x@ 13 user staff 416 .git        # real .git directory, not a worktree file

$ git status
On branch main
Your branch is up to date with 'origin/main'.
...

$ which git && git --version
/usr/bin/git
git version 2.50.1 (Apple Git-155)
```

## Environment

- macOS 26.4.1 (build 25E253)
- git 2.50.1 (Apple Git-155), at `/usr/bin/git`
- Repo path is on an **external/secondary APFS volume** mounted at `/Volumes/Data` (not under `~`)

## Suspected cause

The git-repo detection appears to fail specifically for paths under `/Volumes/*`. Likely one of:

1. **Sandbox / file-access permission** — the app lacks read access to the external volume (macOS TCC "Files and Folders" / Full Disk Access not granted for that volume), so its `.git` check returns false instead of surfacing a permission error.
2. **Path handling bug** — the detection logic doesn't handle multi-segment paths on non-home volumes, or spawns `git` with a working dir / env that resolves differently for `/Volumes/*`.

A direct `git -C <path> rev-parse --is-inside-work-tree` from a terminal returns `true` at the same path, so the underlying git invocation works — the failure is in how emdash performs the check.

## Suggested fix

- Detect repos via `git -C <path> rev-parse --is-inside-work-tree` (or check for `.git`) and, when it fails, distinguish "not a repo" from "couldn't access path / permission denied" so the toast is accurate.
- Verify the app has the entitlements / TCC permission to read `/Volumes/*`.

Image

rabanspiegel:
Fixes #2635.

This changes project git inspection to use git -C and shows an inspection error when the path cannot be read instead of saying the folder is not a git repository.

Tests:
- pnpm --filter @emdash/core exec vitest run src/git/git-runtime.test.ts src/exec/bound-exec.test.ts
- pnpm exec vitest run --project node src/main/core/projects/operations/createProject.test.ts src/renderer/features/projects/stores/project-manager.test.ts
- pnpm --filter @emdash/core run typecheck
- pnpm --filter @emdash/emdash-desktop run typecheck
- pnpm --filter @emdash/core run lint
- pnpm --filter @emdash/emdash-desktop run lint

### DIFF
--- a/apps/emdash-desktop/src/main/core/projects/operations/create-local-project.ts
+++ b/apps/emdash-desktop/src/main/core/projects/operations/create-local-project.ts
@@ -8,7 +8,7 @@ import { db } from '@main/db/client';
 import { projects } from '@main/db/schema';
 import { log } from '@main/lib/logger';
 import type { CreateProjectResult, ProjectPathStatus } from '@shared/projects';
-import { checkIsValidDirectory } from '../path-utils';
+import { getDirectoryStatus } from '../path-utils';
 import { ensureProjectRepository } from './create-project-utils';
 import { ensureRepositoryWorkspace } from './ensure-repository-workspace';
 
@@ -22,8 +22,15 @@ export type CreateLocalProjectParams = {
 export async function createLocalProject(
   params: CreateLocalProjectParams
 ): Promise<CreateProjectResult> {
-  const isValidDirectory = checkIsValidDirectory(params.path);
-  if (!isValidDirectory) {
+  const directoryStatus = getDirectoryStatus(params.path);
+  if (directoryStatus.kind === 'inspect-failed') {
+    return err({
+      type: 'inspect-failed',
+      path: params.path,
+      message: directoryStatus.message,
+    });
+  }
+  if (directoryStatus.kind !== 'directory') {
     return err({
       type: 'invalid-directory',
       path: params.path,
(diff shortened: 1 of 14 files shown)


jschwxrz:
solid for local projects, but ssh projects still go through the legacy git-service and that path still classifies git inspection failures as "not-repository". so would have to adjust that as well for parity

rabanspiegel:
will merge as is! We'll achieve SSH parity once we switch remotes to the new git service 🫡

PR #2644, opened 2026-06-23
fix(linux): find capitalized legacy config directory

### LINKED ISSUE #1789: [bug]: Import option missing on Linux due to case-sensitive config directory (~/.config/Emdash vs ~/.config/emdash)
## Summary
On Linux (case-sensitive file systems), the onboarding **Import** option in Emdash v1 does not appear if legacy config files are stored in a capitalized directory (`~/.config/Emdash`) instead of the expected lowercase path (`~/.config/emdash`).

## Environment
- OS: Kubuntu (Ubuntu-based, case-sensitive filesystem)
- Install method: `.deb`
- Previous version: Emdash 0.4x
- Upgraded to: Emdash v1 (observed on current `.deb` release)

## Root Cause
Linux file systems (e.g., ext4) are case-sensitive.  
The legacy config directory created by pre-v1 `.deb` installs is:

    ~/.config/Emdash

However, v1 appears to only check:

    ~/.config/emdash

Because of this mismatch, existing data is not detected and the Import option is never shown.

This issue would not surface on macOS (default case-insensitive filesystem), which may explain why it was missed.

## Workaround
Manually rename the directory:

    mv ~/.config/Emdash ~/.config/emdash

After renaming, the Import option appears as expected.

## Suggested Fix
- Ensure all detection logic (not just migration) checks for both:
  - `~/.config/emdash`
  - `~/.config/Emdash`
- Alternatively, normalize or migrate directories during startup
- Consider a case-insensitive check on Linux for legacy paths

## Impact
Affects Linux users upgrading from pre-v1 versions installed via `.deb`, preventing discovery of existing data and blocking the import flow.

### Steps to Reproduce

1. Install Emdash 0.4x via `.deb`
2. Confirm config directory exists at:

        ~/.config/Emdash

3. Upgrade to Emdash v1 using `.deb`
4. Launch the app and go through onboarding

## Expected Behavior
The onboarding panel should detect existing config data and display the **Import** option.

## Actual Behavior
The **Import** option does not appear.

### emdash Version

1.1.3

### Operating System

Kubuntu 25.10

### Additional Context

It looks like this may already be addressed in `main`:

    // src/main/db/path.ts
    const legacyDirs = ['Electron', 'emdash', 'Emdash'];

If so, the issue may be:
- Not yet included in the released `.deb`
- Or the onboarding Import detection logic is using a different path check than the migration logic

rasitakyol:
### Description

Fixes Linux legacy import detection when pre-v1 data lives in the old capitalized config directory (`~/.config/Emdash`) while v1 checks the lowercase userData directory (`~/.config/emdash`).

This adds a Linux-only fallback candidate for legacy import paths and uses the resolved existing legacy/beta database paths in import status, preview, and import execution. When the legacy v0 database is found in the fallback directory, settings and conversation backfill also read from that same legacy userData directory.

### Related issues

Fixes #1789

### Testing

- `pnpm exec vitest run --project main-db src/main/db/legacy-port/legacy-source/path.test.ts`
- `pnpm exec vitest run --project main-db src/main/db/legacy-port`
- `pnpm run lint`
- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm run build`

Note: local pnpm prints an engine warning because the shell resolves Node v22.22.3 while the repo wants Node >=24, but the commands above completed successfully.

N/A

### DIFF
--- a/apps/emdash-desktop/src/main/db/legacy-port/controller.ts
+++ b/apps/emdash-desktop/src/main/db/legacy-port/controller.ts
@@ -1,9 +1,7 @@
-import { join } from 'node:path';
 import { count } from 'drizzle-orm';
 import { drizzle } from 'drizzle-orm/better-sqlite3';
 import { app } from 'electron';
 import { db } from '@main/db/client';
-import { PREVIOUS_DB_FILENAME } from '@main/db/default-path';
 import * as schema from '@main/db/schema';
 import { projects, tasks } from '@main/db/schema';
 import { log } from '@main/lib/logger';
@@ -13,7 +11,8 @@ import { openLegacyReadOnly } from './legacy-source/open-readonly';
 import {
   hasBetaDatabaseFile,
   hasLegacyDatabaseFile,
-  resolveLegacyDatabasePath,
+  resolveExistingBetaDatabasePath,
+  resolveExistingLegacyDatabasePath,
 } from './legacy-source/path';
 import { createDefaultLegacyPortStateStore, runLegacyPort } from './service';
 import { createLegacyPortPreview } from './source-analysis';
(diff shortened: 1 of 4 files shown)


jschwxrz:
thank you for the fix!

jschwxrz (approved):

PR #2650, opened 2026-06-23
docs(agents): update provider integration paths

rasitakyol:
## Summary
- update stale provider integration doc paths after the shared/dependency/agent-hook module moves
- keep the provider workflow doc pointing at files that exist in the current tree

## Test Plan
- [x] verified the documented paths exist in the current checkout
- [x] `pnpm --filter @emdash/emdash-desktop run format:check`

### DIFF
--- a/agents/integrations/providers.md
+++ b/agents/integrations/providers.md
@@ -2,8 +2,8 @@
 
 ## Source Of Truth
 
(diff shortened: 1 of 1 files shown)


arnestrickmann (approved):
good catch - thanks for updating @rasitakyol

PR #2657, opened 2026-06-24
fix: stale agent hook declarations

janburzinski:
### Description
- refreshed agent hook declarations
- added missing lifecycle/session/error hooks
- fixed stale kilo stop support
- added mistral tool-use hook

<details>
<summary>Checklist</summary>

</details>


apps/emdash-desktop/src/shared/core/agents/agent-provider-registry.ts
Davidknp, line 1: We should be able to get rid of this after the plugin refactor now, right?

janburzinski: valid!! was just a little unsure. removed it :D

janburzinski:
@greptileai

PR #2676, opened 2026-06-25
fix(ssh): apply write backpressure on the SSH PTY channel

### LINKED ISSUE #1994: [bug]: PTY panel crash during tmux mouse selection
### Bug Description

PTY panelem crashes (becomes unresponsive, must restart emdash) when performing multi-line mouse selection inside a tmux session with mouse mode enabled on a remote SSH connection. Likely related to high-frequency bidirectional IPC traffic (mouse input and terminal output) with no backpressure mechanism, though exact crash mechanism is unconfirmed.

### Steps to Reproduce

1. Create a project connected to a remote SSH host
2. Open a task/terminal session (PTY panel)
3. SSH into the remote and start tmux (which enables mouse mode by default via `set -g mouse on`)
4. Open a file or long output in tmux (enough content to select across multiple lines)
5. Click and drag to select multiple lines of text
6. Continue holding and dragging; within a few seconds the panel freezes and becomes unresponsive

Escape, q, ctrl+Q, etc. does not work. All other tmux sessions on the same host also freezes with a \[server unexpectedly\] message. To recover, emdash has to be restarted. The sub-tasks open under a project task (different terminal windows for a given task) do not resume by themselves, so they have to be closed, new sub-tasks (agents) have to be created, and resumed with the agent's own commands (i.e., after restarting emdash, I need to close the existing agent windows, create new agents, and then use agent's /resume command). **This is very counter-productive, one mis-click somewhere requires me to close emdash entirely and create new agents.**

**Actual**: Panel freezes during multi-line mouse drag. Eventually the panel crashes or becomes permanently unresponsive, requiring an emdash restart. No error is surfaced to the user.

**Expected**: Mouse selection should work normally inside tmux, or at minimum degrade gracefully (slowdown, dropped frames) without crashing the panel.

### emdash Version

v1.1.16

### Operating System

macOS connecting via SSH tunnel to RHEL running tmux

## Root Cause Analysis (Hypothesized, Not Profiled)

> **Note**: This analysis is based on code inspection only. The exact bottleneck (Canvas rendering, IPC serialization, GC pressure, or an Electron limitation) has not been confirmed via profiling. The crash mechanism (OOM, unresponsive timeout, canvas context loss) is also unconfirmed.

**What we know from code inspection:**

1. **Bidirectional traffic with tmux mouse mode** (verified in code):
   * When tmux mouse mode is active (`\e[?1000h` / SGR `\e[?1006h`), xterm.js reports mouse events back to the PTY via `terminal.onData()` in `use-pty.ts`
   * Each mouse move generates both an input message (mouse position → PTY) AND an output message (tmux visual update → renderer)
2. **No backpressure exists** (verified in code):
   * `webContents.send()` in `events.ts` fires unconditionally
   * No mechanism for renderer to signal "slow down" to main process
   * Main process will continue queuing IPC messages regardless of renderer state
3. **Unbounded flush buffer** (verified in code):
   * `buffer += data` in `pty-session-registry.ts` has no size cap
   * Ring buffer is capped at 64KB, but active flush buffer is not

**Hypothesized crash sequence** (one plausible path, not confirmed):

1. Mouse drag at high rate (\~100+ events/sec, exact rate unknown)
2. Renderer becomes CPU-bound from some combination of: DOM events, xterm parsing, Canvas repaints, IPC handling
3. Main process continues sending via IPC (no backpressure)
4. Renderer eventually crashes or becomes permanently unresponsive

**Possible bottlenecks** (not determined which, if any, is primary):

* CanvasAddon (Canvas2D) repaint frequency under rapid cursor movement
* IPC message serialization/deserialization overhead at high frequency
* V8 GC pressure from string allocations in both directions
* Electron renderer unresponsive detection
* Something else entirely (xterm.js internal state, Electron IPC limits)

**Contributing factors** (observed in code, severity unknown):

* Auto-copy selection debounce in `use-pty.ts` adds work during drag
* CanvasAddon (Canvas2D) may be less performant than WebGL for rapid repaints

## Relevant Code Locations

| File | What to look for |
| -- | -- |
| `src/main/core/pty/pty-session-registry.ts` | `buffer += data` unbounded accumulation, 16ms flush timer |
| `src/main/lib/events.ts` | `webContents.send()` with no backpressure |
| `src/renderer/lib/pty/pty.ts` | `terminal.write(data)` no error handling in `connect()` |
| `src/renderer/lib/pty/use-pty.ts` | `terminal.onData` sends mouse input back to PTY; `onSelectionChange` debounce |

## Suggested Fixes (Pending Root Cause Confirmation)

These are reasonable directions based on code inspection, but the right fix depends on profiling to identify the actual bottleneck:

1. **IPC backpressure**: Implement flow control so main process pauses PTY forwarding when renderer signals overload (e.g., high-water mark on pending writes)
2. **Throttle xterm.write()**: Batch incoming data per `requestAnimationFrame` instead of writing synchronously on every IPC message
3. **Cap flush buffer**: Add max size (e.g., 16KB) to the accumulation buffer in pty-session-registry, drop oldest data on overflow
4. **Consider WebGL addon**: `@xterm/addon-webgl` handles rapid repaints better than CanvasAddon for GPU-capable systems
5. **Defensive try-catch**: Wrap `terminal.write()` in try-catch to prevent renderer crash if xterm ever does throw (defensive, not primary fix)

**To confirm root cause**, a developer could:

* Run Electron DevTools Performance profiler during reproduction
* Check `chrome://tracing` for IPC queue depth
* Add temporary logging to measure events/sec during mouse drag
* Test with WebGL addon to see if crash still occurs

fiorelorenzo:
## Problem

`Ssh2PtySession.write()` ignored the boolean returned by ssh2's `channel.write()`, so it kept writing even when the channel's send buffer was over its high-water mark. A tmux mouse drag (mouse mode is on by default) floods SGR mouse reports faster than the remote can drain; with no backpressure this froze the PTY panel and the remote tmux server, requiring an app restart.

Fixes #1994.

## Fix

Respect the stream contract in `Ssh2PtySession`:

- when `channel.write()` returns `false` (buffer over high-water mark), defer further input and flush it on the next `drain` event, in order;
- the normal path (`write()` returns `true`) is unchanged — no queueing, no added latency;
- `kill()` drops any deferred input and detaches the `drain` listener so nothing is written to a closed channel.

## Testing

`format` / `lint` / `typecheck` clean; unit tests cover the deferral/flush ordering, repeated drains while still full, and the kill-while-draining path (plus the existing wrap-semantics test).

The fix targets the missing-backpressure root cause called out in the issue. I couldn't reproduce the exact freeze locally — it needs a GUI mouse drag over SSH, and triggering it risks hanging a shared tmux server — so confirmation from the reporter would be welcome.

### DIFF
--- a/apps/emdash-desktop/src/main/core/pty/ssh2-pty.test.ts
+++ b/apps/emdash-desktop/src/main/core/pty/ssh2-pty.test.ts
@@ -6,10 +6,12 @@ class FakeClientChannel extends EventEmitter {
   writes: string[] = [];
   windows: Array<{ rows: number; cols: number; height: number; width: number }> = [];
   closed = false;
+  /** When false, `write()` returns false to simulate a full send buffer. */
+  acceptWrites = true;
 
   write(data: string): boolean {
     this.writes.push(data);
-    return true;
+    return this.acceptWrites;
   }
 
   setWindow(rows: number, cols: number, height: number, width: number): void {
@@ -41,5 +43,55 @@ describe('Ssh2PtySession', () => {
     expect(dataHandler).toHaveBeenCalledWith('remote output');
     expect(channel.closed).toBe(true);
     expect(exitHandler).toHaveBeenCalledWith({ exitCode: 0, signal: undefined });
+  });
+
+  it('defers writes while the channel buffer is full and flushes them on drain, in order', () => {
+    const channel = new FakeClientChannel();
+    const session = new Ssh2PtySession('s', channel as never);
+
+    channel.acceptWrites = false; // buffer over high-water mark
+    session.write('a'); // buffered by ssh2, returns false -> start draining
+    session.write('b'); // deferred
+    session.write('c'); // deferred
+    expect(channel.writes).toEqual(['a']);
+
+    channel.acceptWrites = true;
+    channel.emit('drain');
+    expect(channel.writes).toEqual(['a', 'b', 'c']);
+    expect(channel.listenerCount('drain')).toBe(0);
+  });
+
+  it('keeps deferring across multiple drains while the channel stays full', () => {
+    const channel = new FakeClientChannel();
+    const session = new Ssh2PtySession('s', channel as never);
+
+    channel.acceptWrites = false;
+    session.write('a'); // draining
+    session.write('b'); // deferred
+    channel.emit('drain'); // still full: 'b' is buffered but write returns false again
+    expect(channel.writes).toEqual(['a', 'b']);
+
+    channel.acceptWrites = true;
+    session.write('c'); // still deferred (draining re-armed)
+    channel.emit('drain');
+    expect(channel.writes).toEqual(['a', 'b', 'c']);
+  });
+
+  it('drops deferred writes and removes the drain listener on kill()', () => {
+    const channel = new FakeClientChannel();
+    const session = new Ssh2PtySession('s', channel as never);
+
+    channel.acceptWrites = false;
+    session.write('a'); // draining
+    session.write('b'); // deferred
+    session.kill();
+
+    expect(channel.closed).toBe(true);
+    expect(channel.listenerCount('drain')).toBe(0);
+
+    channel.acceptWrites = true;
+    channel.emit('drain'); // no-op, listener removed
+    session.write('c'); // ignored, session closed
+    expect(channel.writes).toEqual(['a']);
   });
 });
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Hi @fiorelorenzo thanks again for your PRs! 
We'll have a look and get back to you here shortly as always.

jschwxrz (approved):

PR #2679, opened 2026-06-25
fix(diff): refresh the Changed diff after a branch switch in the same worktree

### LINKED ISSUE #2576: [bug]: Diff panel shows changes from previous branch after switching branches

fiorelorenzo:
## What

The Changed and Staged diffs use the literal ref `HEAD` (`commitRef('HEAD')`) for their original side. After switching branches inside a task worktree, that original kept showing the previous branch's content, so the Changed diff showed stale changes. Fixes #2576.

## Root cause

There were two separate problems and both had to be fixed.

1. Renderer (first commit). A `head` worktree update only invalidated `HEAD_REF` git models, not the `commitRef('HEAD')` ones the diff actually registers. `HEAD_REF` and `commitRef('HEAD')` compare equal only within their own kind, so the original model was never re-fetched after a branch switch.

2. Main process (second commit). `GitWorktree.getFileAtRef` delegated straight to `repository.readBlobAtRef`, but `GitRepository` is shared per common dir and its `exec` is bound to whichever worktree opened it first (usually the project's main checkout). A symbolic ref like `HEAD` therefore resolved against that first worktree, not the caller, so a task worktree on its own branch read the wrong branch's blob. Even with the renderer invalidation in place the re-fetch kept returning the wrong worktree's HEAD, which is why fixing only the renderer was not enough.

## The fix

Resolve the ref to an immutable commit oid with the worktree-bound `exec` before reading. The oid is content addressed so the shared object store reads it correctly regardless of cwd. Already-immutable oids and resolution failures pass through untouched, and the existing `git show` fallback still handles odd refs.

## Testing

- New unit test in `git-worktree.test.ts` reproduces the exact scenario: a linked worktree on its own branch sharing a repository bound to the main checkout. It fails without the main-process fix and passes with it.
- New unit test for the renderer invalidation helper covers both the `head` and `status` paths.
- Verified against the real built package and a real worktree: with both fixes `getFileAtRef('HEAD')` returns the calling worktree's HEAD content and matches the ground truth `git show HEAD:file`.

### DIFF
--- a/apps/emdash-desktop/src/renderer/lib/monaco/git-model-invalidation.test.ts
+++ b/apps/emdash-desktop/src/renderer/lib/monaco/git-model-invalidation.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from 'vitest';
+import { HEAD_REF, STAGED_REF, type GitRef } from '@shared/core/git/types';
+import { commitRef, refsEqual } from '@shared/core/git/utils';
+import { invalidateWorktreeGitModels } from './git-model-invalidation';
+import type { MonacoModelRegistry } from './monaco-model-registry';
+
+/** Minimal registry whose findGitUris matches by ref like the real one (via refsEqual). */
+function makeRegistry(entries: Array<{ ref: GitRef; uri: string }>) {
+  const invalidated: string[] = [];
+  const registry = {
+    findGitUris: ({ ref }: { ref?: GitRef }) =>
+      entries.filter((e) => ref !== undefined && refsEqual(e.ref, ref)).map((e) => e.uri),
+    invalidateModel: (uri: string) => {
+      invalidated.push(uri);
+      return Promise.resolve();
+    },
+  } as unknown as MonacoModelRegistry;
+  return { registry, invalidated };
+}
+
+describe('invalidateWorktreeGitModels', () => {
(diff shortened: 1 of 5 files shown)


jschwxrz:
closing this as outdated

PR #2681, opened 2026-06-25
fix(pty): write SSH image attachments into .emdash, not the worktree root

### LINKED ISSUE #2680: [bug]: SSH image attachments are written to the worktree root and never cleaned up
### Bug Description

When attaching an image (paste or drag and drop) to an agent running on a **remote SSH** project, the image is uploaded into the **root of the worktree** and is never removed. Every attached image leaves an untracked file behind, so the repo working tree gets progressively dirtier and `git status` fills up with junk like:

```
<uuid>-emdash-drop-<uuid>-image.png
```

This is not specific to Claude. The upload happens in the shared PTY layer (`pty.uploadFiles`), so it affects every agent used over SSH.

The local (non SSH) path does not have this problem: there the image is persisted to the OS temp dir, tracked, and cleaned up on quit and by an age based sweep (`cleanupExpiredDroppedBlobs`). The remote branch has no equivalent: it writes inside the working tree and never cleans up.

### Steps to Reproduce

1. Open a remote SSH project and start an agent (Claude, Codex, etc.).
2. Paste or drop an image into the agent terminal.
3. Run `git status` in that worktree.

**Actual:** a file named `<uuid>-<original-name>` appears in the worktree root as an untracked file and stays there forever. After a few images the repo is full of orphaned `*-image.png` files.

**Expected:** the attached image should not pollute the tracked working tree, and any temporary copy needed to hand the image to the agent should be cleaned up (mirroring the local behavior).

### Technical Notes

`apps/emdash-desktop/src/main/core/pty/controller.ts` (`uploadFiles`):

```ts
const remoteName = `${randomUUID()}-${basename(localPath)}`;
await workspace.fs.copyLocalFile!(localPath, remoteName);
return `${workspace.path}/${remoteName}`;
```

`copyLocalFile` writes `remoteName` relative to `workspace.path` (the worktree root), and nothing ever deletes it. Compare with the local persistence in `persist-dropped-blob.ts`, which writes to `app.getPath('temp')` and has `cleanupExpiredDroppedBlobs` / `will-quit` cleanup.

### Possible Fix

Upload remote attachments outside the tracked working tree (a remote temp or cache dir, or a git ignored location) and/or track and remove them once the agent has consumed them, mirroring the local design.

fiorelorenzo:
## What

On remote SSH projects, attaching an image (paste or drag and drop) to an agent left an untracked `<uuid>-<name>` file in the worktree root every time, and it was never cleaned up. After a few attachments `git status` is full of orphaned `*-image.png` files. Fixes #2680.

This is not Claude specific: the upload runs in the shared PTY layer (`pty.uploadFiles`), so it affects every agent used over SSH.

## Root cause

`apps/emdash-desktop/src/main/core/pty/controller.ts` (`uploadFiles`) copied the local file straight into the workspace root:

```ts
const remoteName = `${randomUUID()}-${basename(localPath)}`;
await workspace.fs.copyLocalFile!(localPath, remoteName);
return `${workspace.path}/${remoteName}`;
```

`copyLocalFile` writes relative to `workspace.path` (the worktree root) and nothing ever removed the file. The local (non SSH) path does not have this problem: there images go to the OS temp dir with `will-quit` and age based cleanup (`persist-dropped-blob.ts`).

## The fix

Upload into the git-ignored `.emdash/uploads` directory instead of the worktree root. `.emdash` is already emdash's per-project runtime state dir (it holds the SSH worktree pool via `getDefaultSshWorktreeDirectory` and attachments via `saveAttachment`), so attachments no longer show up in `git status`. The directory is created with `mkdir({ recursive: true })` before the transfer.

## Testing

- New unit test in `controller.test.ts` asserts the upload creates `.emdash/uploads` and that both the destination relative path and the returned remote path live under `.emdash/uploads`, never directly in the worktree root.

## Note

This stops the pollution of the tracked tree. Bounding disk usage of `.emdash/uploads` over time (a TTL sweep like the local `cleanupExpiredDroppedBlobs`) would be a reasonable follow up.

### DIFF
--- a/apps/emdash-desktop/src/main/core/pty/controller.test.ts
+++ b/apps/emdash-desktop/src/main/core/pty/controller.test.ts
@@ -1,5 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { conversationEvents } from '@main/core/conversations/conversation-events';
+import { taskSessionManager } from '../tasks/task-session-manager';
+import { workspaceRegistry } from '../workspaces/workspace-registry';
 import { ptySessionRegistry } from './pty-session-registry';
 
 vi.mock('./persist-dropped-blob', () => ({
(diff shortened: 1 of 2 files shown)


fiorelorenzo:
Good catch on the mock leak. Pushed a follow-up that resets `getTask`/`getWorkspaceId`/`get` on the shared mock singletons in `afterEach`, so a later test still starts from the empty mocks.

jschwxrz (approved):

jschwxrz:
thank you!

PR #2682, opened 2026-06-25
fix(projects): keep emdash's .emdash runtime dir out of git status

fiorelorenzo:
## What

emdash keeps per-project runtime state under `.emdash/` inside the repo: the SSH worktree pool (`getDefaultSshWorktreeDirectory` -> `.emdash/worktrees`), saved attachments (`saveAttachment` -> `.emdash/attachments`), and uploaded images (#2681 -> `.emdash/uploads`). Nothing in the app ensured `.emdash/` was git-ignored, so it only stayed out of `git status` if the user happened to have `.emdash` in a global gitignore. Without that, `.emdash/` shows up as untracked and clutters the working tree.

This adds `.emdash/` to the repo's `.git/info/exclude` when a project is opened.

## Why info/exclude

`.git/info/exclude` is a local, uncommitted ignore file, so this never touches the user's tracked `.gitignore`. It lives in the git common dir, so a single entry on the main checkout also covers every linked task worktree (where the per-worktree `.emdash/uploads` is written).

## Details

- New helper `ensureEmdashGitExcluded(fs)` in `src/main/core/projects/ensure-emdash-excluded.ts`, called fire-and-forget from `buildProvider` so it runs once per project open for both local and SSH providers.
- Uses the `FileSystemProvider` abstraction (`stat` / `exists` / `read` / `write`) so it works over both local and SSH without shelling out.
- Best effort and idempotent: skips repos whose `.git` is a file (linked worktree / submodule, where the exclude is out of the fs root) and skips when `.emdash` is already ignored. Failures are logged, never thrown.

## Testing

- Unit test covering: skip on non-directory `.git`, create the entry when `info/exclude` is missing, append while preserving existing content, and no-op when `.emdash` (with or without trailing slash) is already present.
- Manually verified the underlying behavior: with `core.excludesfile=/dev/null`, a fresh repo shows `.emdash/` as untracked; after appending `.emdash/` to `.git/info/exclude` it disappears from `git status`.

## Relationship

Relates to #2680 and complements #2681. #2681 moves SSH image uploads into `.emdash/uploads`; this PR makes sure `.emdash/` (uploads, worktree pool, and attachments) is actually ignored even for users without a global `.emdash` gitignore.

### DIFF
--- a/apps/emdash-desktop/src/main/core/projects/create-project-provider.ts
+++ b/apps/emdash-desktop/src/main/core/projects/create-project-provider.ts
@@ -19,6 +19,7 @@ import { log } from '@main/lib/logger';
 import { gitRepoUpdateChannel } from '@shared/core/git/events';
 import { safePathSegment } from '@shared/path-name';
 import type { LocalProject, SshProject } from '@shared/projects';
+import { ensureEmdashGitExcludedSafe } from './ensure-emdash-excluded';
 import { ProjectProvider, type ProjectProviderTransport } from './project-provider';
 import type { ProjectSettingsProvider } from './settings/provider';
 import { LocalProjectSettingsProvider } from './settings/providers/local-project-settings-provider';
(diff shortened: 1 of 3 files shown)


fiorelorenzo:
Good hardening note. Pushed a follow-up that bails out when the `.git/info/exclude` read comes back truncated, so a partial view can never overwrite the tail of the file. Added a test for it too.

jschwxrz (approved):

PR #2686, opened 2026-06-25
feat(storage): add task worktree storage management

rabanspiegel:
## Summary

Adds a Storage settings page for reviewing task worktree disk usage and deleting stale tasks from one place.

This version:
- shows task worktree size grouped by project
- shows selected size before deletion
- deletes selected tasks and their owned worktrees
- keeps branch deletion out of this flow

## Why

Worktrees can accumulate and take meaningful disk space, especially across old tasks. Users need a direct way to see which tasks are taking space and remove them without manually inspecting ~/emdash/worktrees.

This also gives us a clean foundation for future storage work without putting disk-scanning logic directly in the renderer or duplicating task deletion behavior.

## Implementation

Adds a small @emdash/core/storage package for filesystem measurement. This keeps the disk-size logic independent from Electron/main-process UI concerns and makes it easier to reuse later in a workspace-server direction.

Adds a typed main-process storage RPC namespace:
- storage.listTaskStorageUsage
- storage.deleteTasks

The Storage service queries tasks/workspaces, measures local worktree usage, and returns Storage-specific view data to the renderer.

Deletion intentionally reuses the canonical task deletion path:

```ts
taskService.deleteTask(projectId, taskId, {
  deleteWorktree: true,
  deleteBranch: false,
});
```

So Storage does not directly delete task DB rows, view state, telemetry, or task events. That stays owned by the task lifecycle code.

The task deletion path now also has a local owned-path fallback: if the mounted project provider cannot remove the worktree, deletion can still remove the DB-owned local worktree path, while refusing to delete the project root.

## UI

Adds Settings -> Storage with:
- total measured task worktree size
- selected size
- project-grouped task rows
- refresh
- delete selected tasks

Regular sidebar/task-list deletes stay quiet on success, but now show an error toast if deletion fails. Storage keeps delete feedback because bulk deletion can partially fail.

## Non-goals

This does not add:
- generated/cache cleanup
- branch deletion
- pnpm store pruning
- low-disk prompts
- background cleanup policies
- telemetry for Storage page usage

## Testing

- pnpm run typecheck
- pnpm run lint
- focused task deletion lifecycle tests
- core storage measurement tests


packages/core/src/files/fs/measure-usage.ts
jschwxrz, line 1: storage cleanup is a feature of the desktop app, not a core runtime primitive. the only reusable primitive here is filesystem measurement. that capability should become a generic size/stat method on the filesystem abstraction coming in PR #2668

the storage domain should then be removed

apps/emdash-desktop/src/main/core/tasks/operations/task-lifecycle-utils.ts
jschwxrz, line 1: this file bypasses our worktree/runtime abstractions by calling node:fs etc. directly. until now, it should have used the execution context and as soon as pr #2668 drops, the new file system abstraction in core

apps/emdash-desktop/src/main/core/tasks/operations/task-lifecycle-utils.ts
```diff
@@ -1,12 +1,122 @@
...
+  workspace: LocalWorkspaceCleanupTarget,
+  projectPath: string
+): Promise<boolean> {
+  if (!workspace.path || !isLocalWorkspace(workspace)) return false;
+
+  const workspacePath = path.resolve(workspace.path);
+  const projectRootPath = path.resolve(projectPath);
+  if (workspacePath === projectRootPath) {
+    if (workspace.kind === 'worktree') {
+      throw new Error(`Refusing to remove project root path "${workspace.path}"`);
```
jschwxrz: the repo convention is returning "Result" for expected error and only throw for unexpected errors. this is an expected error.

there are multiple occurrences for this in the current pr. using result makes failure modes more explicit, typed and forces callers to handle them properly. it also avoids parsing thrown error.message (see the renderer storage components)

apps/emdash-desktop/src/main/core/storage/storage-service.ts
jschwxrz, line 1: I would avoid the storage-service here. it is not a class and should not be a class because it is not stateful. therefore the name "service" is a bit misleading. I would separate the responsibilities from the service and decompose into smaller operation/helper files.

rabanspiegel:
@greptileai

PR #2713, opened 2026-06-29
fix: add windows ssh agent pipe fallback for ssh conns

janburzinski:
## Summary
- add a shared ssh agent socket resolver that fallsback to the Windows openssh named pipe when `SSH_AUTH_SOCK` is unavailable
- usethat fallback for agent auth and agent forwarding in SSH connection setup


janburzinski:
@greptileai

janburzinski:
@greptileai

PR #2719, opened 2026-06-30
fix: windows ssh include glob expansion

janburzinski:
## Summary
- expand `~\` alongside `~/` when resolving ssh include paths
- enable windows glob path handling for quoted and unquoted `Include` patterns


janburzinski:
@greptileai

PR #2723, opened 2026-06-30
feat(settings): restore agent and model defaults

arnestrickmann:
## Summary
- Restore global default agent and default model settings in the Agents settings tab.
- Add separate automation defaults for agent and model without changing the automation creation flow.
- Apply saved defaults when creating tasks, conversations, and automation runs, while preserving per-flow overrides.
- Keep the model selector slot stable for agents without selectable models and explain the disabled state in a tooltip.

## Why
Automations and repeated task creation need predictable defaults for both the harness and the underlying model. This brings those settings back in the central Agents settings page and keeps the hover-card shortcut aligned with the same app setting.

Linear: https://linear.app/general-action/issue/ENG-1687/bring-back-default-agent-and-default-underlying-model-settings

## Screenshot
CleanShot 2026-06-30 at 12 17 59

## Validation
- corepack pnpm run typecheck
- corepack pnpm exec vitest run --project node src/main/core/automations/actions/taskCreate.test.ts src/renderer/tests/conversation-provider-selection.test.ts
- focused oxlint and oxfmt checks

### DIFF
--- a/apps/emdash-desktop/src/main/core/automations/actions/taskCreate.test.ts
+++ b/apps/emdash-desktop/src/main/core/automations/actions/taskCreate.test.ts
@@ -326,6 +326,83 @@ describe('executeTaskCreate', () => {
     );
   });
 
+  it('uses automation default agent and model when the automation has no provider or model', async () => {
+    vi.mocked(appSettingsService.get).mockImplementation(async (key) => {
+      if (key === 'defaultAutomationAgent') return 'codex' as never;
+      if (key === 'defaultAutomationModel') return 'gpt-5-codex' as never;
+      return null as never;
+    });
+
+    await executeTaskCreate(
+      {
+        ...automation,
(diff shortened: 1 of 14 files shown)


arnestrickmann:
Addressed the review feedback in the amended commit 42cfa4915:

- Automation forms now opt out of applying the runtime automation default model into local form state. Opening an existing automation with no explicit model keeps conversationConfig.model unset, so future runs continue to use the current defaultAutomationModel fallback instead of silently snapshotting it on form open.
- Existing explicit automation models still seed into the form through the existing seedModel path and can be changed by the user.
- The disabled no-model placeholder no longer presents itself as an activatable ARIA button.
- I intentionally kept defaultAutomationModel independent from the global defaultModel; null for the automation default means use the agent default model, not inherit task/conversation defaults. Added a backend comment to make that explicit.

arnestrickmann:
@greptile

arnestrickmann:
@greptile

PR #2747, opened 2026-07-02
fix(projects): scope project path uniqueness per SSH connection (#2731)

### LINKED ISSUE #2731: [bug]: Remote project path uniqueness is global, not per-SSH-connection
### Bug Description

`projects.path` has a global UNIQUE index, not scoped per SSH connection. Since different remote hosts commonly mount projects at the same conventional path, registering a project on a second host fails if any other host already has a project registered at that same path.

### Steps to Reproduce

1. Add SSH connection `host-a`, register a project at path `/home/user/project`.
2. Add SSH connection `host-b`.
3. Register a project at the same path `/home/user/project` on `host-b`.
4. Observe the error.

- Actual: `Error invoking remote method 'projects.createProject': SqliteError: UNIQUE constraint failed: projects.path`
- Expected: Paths should only need to be unique per connection (`(ssh_connection_id, path)`), since paths are independent across separate remote hosts.

### emdash Version

1.1.35

### Operating System

macOS (Darwin 24.3.0)

### Additional Context

```sql
CREATE UNIQUE INDEX idx_projects_path ON projects (path);
```

Suggested fix: composite unique index on `(ssh_connection_id, path)` instead of `path` alone, with a migration.

fiorelorenzo:
## What

`projects.path` has a global UNIQUE index, so registering a project at a path that another host already uses fails, even though the two live on different machines. Since remote hosts commonly mount projects at the same conventional path, adding the same path on a second SSH connection blows up with `UNIQUE constraint failed: projects.path`.

Closes #2731.

## Fix

Path uniqueness should be per machine, so the global index becomes two partial unique indexes (migration `0019`):

- **local projects** (`ssh_connection_id IS NULL`): unique by `path`
- **SSH projects** (`ssh_connection_id IS NOT NULL`): unique by `(ssh_connection_id, path)`

Two partial indexes rather than a single `(ssh_connection_id, path)` one, because SQLite treats `NULL`s as distinct: a plain composite index would silently drop path uniqueness for local projects (any number of local projects could share a path). The two-index split is the same partial-index pattern already used for `idx_workspaces_key` in this schema.

```sql
DROP INDEX IF EXISTS `idx_projects_path`;
CREATE UNIQUE INDEX `idx_projects_local_path` ON `projects` (`path`) WHERE "ssh_connection_id" is null;
CREATE UNIQUE INDEX `idx_projects_ssh_connection_path` ON `projects` (`ssh_connection_id`,`path`) WHERE "ssh_connection_id" is not null;
```

The migration only loosens the constraint (the old global unique already forbade every case the new indexes forbid), so no existing database can violate it and there is nothing to backfill.

### Follow-on

Allowing a local and a remote project to share a path string (they are different machines) means a path-only lookup can now match across kinds, so `getLocalProjectByPath` is scoped to local rows (`ssh_connection_id IS NULL`). This keeps `inspectProjectPath`'s local existence check from returning an SSH project. `getSshProjectByPath` was already scoped by `connectionId`, so it needed no change.

## Tests

`0019_project_path_unique_per_connection.test.ts` covers:

- the old `idx_projects_path` is dropped
- both partial unique indexes exist with the right `WHERE` clause
- the same path on two different SSH connections is now allowed (the bug)
- the same path on the same connection is still rejected
- two local projects at the same path are still rejected
- a local and an SSH project may share a path string

```
pnpm run test:migrations   -> 7 files, 24 tests passed
pnpm exec vitest run src/main/core/projects/   -> 11 files, 124 tests passed
```

format and lint are clean. Followed the migration authoring flow: `db:generate`, committed the `pre-0019` fixture, regenerated fixtures with `db:fixtures`.


fiorelorenzo:
@greptileai please re-review. I pushed the fix for your P2 finding in 35b4d470e: `getLocalProjectByPath` now also matches `eq(projects.workspaceProvider, 'local')` alongside `isNull(sshConnectionId)`, so an orphaned SSH project (connection deleted via `onDelete: 'set null'`, leaving `ssh_connection_id` null while `workspace_provider` stays 'ssh') is no longer returned as a LocalProject. I kept `isNull(sshConnectionId)` so the query still uses the `idx_projects_local_path` partial index. CI is green.

fiorelorenzo:
Rebased on main and force-pushed to clear the conflict.

While this was in review, main landed its own migration 0019 (`0019_eager_meteorite`, the `provider_accounts` table), which collided with this PR's `0019_project_path_unique_per_connection`. I renumbered mine to **0020** so it runs after `eager_meteorite`, regenerated `drizzle/meta` (journal + snapshot) with `drizzle-kit generate`, regenerated the `baseline`/`empty` fixtures, and renamed the migration test to `0020`. The two migrations touch different tables (`provider_accounts` vs the `projects` indexes), so there was no semantic conflict, only the numbering.

The regenerated `0020` SQL is byte-identical to the old `0019` one (drop the global path index, add the two partial unique indexes). Migration tests pass (27), format and lint are clean.

jschwxrz:
closing this as resolved

PR #2749, opened 2026-07-02
feat(terminals): support dragging terminals into task tabs

arnestrickmann:
## Summary

Adds terminal tabs as a first-class task tab kind by letting users drag an existing terminal from the bottom terminal pane into the main task view. Dropping a terminal opens it as a large main-pane terminal tab while reusing the existing terminal session, so terminal creation and deletion remain owned by the terminal drawer.

The implementation preserves the existing drawer/full terminal workflow: the New Terminal command still creates terminals in the drawer, closing a main-pane terminal tab only closes that view, and deleting a terminal from the drawer stale-closes any matching main-pane terminal tab.

## Reviewer notes

- Verify that terminal tab state follows the existing provider-based task tab model: descriptors persist/restore through pane snapshots, terminal tabs are single-mounted by terminal ID, and split-pane/tab movement uses the existing pane layout APIs.
- Stress-test PTY ownership by dragging Terminal 1 into the main pane, creating Terminal 2 in the drawer, then opening both as main tabs; each should keep separate output/history and the drawer should not mount a PTY that is already open in a main tab.
- Confirm existing terminal workflows still behave the same: New Terminal opens in the drawer, drawer deletion owns terminal cleanup, and closing a main terminal tab does not delete the terminal session.

## Validation

- corepack pnpm --dir apps/emdash-desktop exec vitest run src/renderer/features/tabs/pane-store.test.ts src/renderer/features/tasks/commands.test.ts
- corepack pnpm --dir apps/emdash-desktop run typecheck
- corepack pnpm --dir apps/emdash-desktop run lint


arnestrickmann:
@greptile

arnestrickmann:
Addressed the terminal-tab review points in the latest push:

- Deferred stale terminal tab close calls that can fire during provider initialization, so restored/deleted terminals close after initialization finishes.
- Replaced the casted fallback resource with a real missing-terminal resource that safely implements the terminal tab resource surface while the tab is being closed.
- Added generic single-mount behavior for explicit pane targets: dragging/opening an already-mounted terminal into another pane moves the existing tab there instead of duplicating it or silently focusing the old pane. Covered by pane-layout-store tests.

Davidknp (approved):

PR #2780, opened 2026-07-07
fix: prevent triage icon clipping

janburzinski:
## summary
- Fix triage status icon clipping by expanding its SVG viewBox

## screenshot
Screenshot 2026-07-07 at 14 40 56


arnestrickmann (approved):
good catch, thank you

PR #2781, opened 2026-07-07
fix(tasks): preserve create task tab state

janburzinski:
Previously, switching tabs in the create task modal reset draft content; now tab contents stay mounted so edits are preserved.

## summary
- preserve create task modal tab contents when switching tabs
- add coverage for preserving state across task config tab switches

## screenrecording
https://cap.link/j6d15rqppn7td23


janburzinski:
@greptileai

arnestrickmann:
Thanks!

arnestrickmann:
Have you also tested if the content persists inside the input field after switching?

janburzinski:
yes, input state perists when switching tabs and back (if you meant that :D)

PR #2794, opened 2026-07-08
fix(tasks): keep task modal open after app blur

janburzinski:
## summary
- keep the new task modal from dismissing after the app window blurs
- scope the blur/outside-press guard to the task modal

## demo
- https://cap.link/tgrsmhwwcmyvmj2

## testing
- not run; changes were already committed


arnestrickmann (approved):
thanks! tested and confirmed the behavior - lgtm - will update the user who flagged this.

PR #2804, opened 2026-07-09
chore: upgrade TypeScript to 7

janburzinski:
## Summary

- upgrade workspace TypeScript consumers from `^6.0.2` to `^7.0.2`
- patch `storybook-solidjs-vite@10.4.0` so Solid component metadata generation uses the TS6 legacy JS compiler API fallback when TS7 no longer exposes `JsxEmit`
- declare `@typescript/typescript6` as a `storybook-solidjs-vite@10.4.0` dependency via pnpm `packageExtensions`, so the patched package resolves its fallback explicitly instead of relying on root hoisting
- update `pnpm-lock.yaml` and register the patch/package extension in `pnpm-workspace.yaml`

## Benchmark overview

Benchmarked with `tsc -p <tsconfig> --noEmit --pretty false`, 3 alternating runs each, comparing this branch against the TS6 baseline checkout.

| Project | TS6 mean | TS7 mean | Speedup |
| --- | ---: | ---: | ---: |
| `packages/chat-ui` | 20.147s | 4.250s | 4.74× faster |
| `packages/ui` | 32.776s | 7.362s | 4.45× faster |
| `apps/emdash-desktop` | 98.059s | 23.463s | 4.18× faster |
| Average project mean | 50.327s | 11.692s | 4.30× faster |

## Checks run

- [x] `pnpm install --frozen-lockfile --ignore-scripts`
- [x] explicit `@typescript/typescript6` resolution from `storybook-solidjs-vite`
- [x] `pnpm run typecheck`
- [x] `pnpm run build`
- [x] `pnpm run lint`
- [x] `pnpm run format:check`
- [x] `git diff --check`
- [x] `pnpm --filter @emdash/chat-ui run build-storybook`
- [x] `pnpm --filter @emdash/ui run build:storybook`
- [x] focused reruns for flaky/default-timeout failures against TS6 baseline

### DIFF
--- a/apps/emdash-desktop/package.json
+++ b/apps/emdash-desktop/package.json
@@ -151,7 +151,7 @@
     "tailwind-merge": "^2.6.0",
     "tailwindcss": "^4.2.1",
     "tw-animate-css": "^1.4.0",
-    "typescript": "^6.0.2",
+    "typescript": "^7.0.2",
     "vite": "^6.4.1",
     "vitest": "^4.1.0",
     "vitest-browser-react": "^2.1.0"
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     ]
   },
   "devDependencies": {
-    "@typescript/typescript6": "^6.0.2",
     "nx": "^23.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     ]
   },
   "devDependencies": {
+    "@typescript/typescript6": "^6.0.2",
     "nx": "^23.0.0"
   }
 }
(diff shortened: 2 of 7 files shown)


janburzinski:
@greptileai

Davidknp:
If this needs a patch I don't think we should do this switch. We can already utilise the performance gains by using tsgo whilst still staying on ts6, no?

janburzinski:
yea, youre right. we could maybe just use tsgo for other things too then, maybe? and just scrap ts7 for now

Davidknp:
The typecheck command is already using `tsgo` in most places. Most IDEs also have the option to use a tsgo backed LSP as well. And for builds vite and tsdown handle it internally.

PR #2809, opened 2026-07-09
feat(tasks): add archive shortcut

janburzinski:
### Description

Adds a task-scoped `Archive Task` command with a configurable keyboard shortcut.

- Adds `Archive Task` to keyboard shortcut settings with default `Mod+Shift+Backspace`
- Exposes `task.archive` in the task command provider / command palette
- Archives the current task and returns to the project view
- Prevents the shortcut from firing while Monaco is focused to avoid editor conflicts

### Related issues

ENG-1793

### Screenshot/Recording

Screen recording: https://cap.link/by67ep2778855qr

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/apps/emdash-desktop/src/renderer/features/command-palette/command-icons.ts
+++ b/apps/emdash-desktop/src/renderer/features/command-palette/command-icons.ts
@@ -1,4 +1,5 @@
 import {
+  Archive,
   ArrowDownToLine,
   ArrowLeft,
   ArrowRight,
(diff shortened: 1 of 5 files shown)


janburzinski:
@greptileai

janburzinski:
for some reason the gh action is not starting

arnestrickmann:
Quick heads up: the shortcut didn't work when you were clicked into the terminal, because the terminal eats the Backspace key before the app sees it. I pushed a fix that lets the shortcut through.

arnestrickmann:
@janburzinski 
quick thing: the shortcut doesn't work when you're clicked into the terminal. the terminal swallows backspace before the app sees it, and in a task view you're basically always in the terminal. shortcuts like cmd+k work fine because letters pass through. can you change the default binding to something without backspace, e.g. cmd+shift+e? then it just works everywhere and we don't need extra code for it 🙏

arnestrickmann (approved):

PR #2810, opened 2026-07-09
ci(release): add arm64 deb and rpm builds to release workflows

fatih-acar:
### Description

Adds arm64 `.deb` and `.rpm` packages to the production and canary release workflows.

A new `release-linux-arm` job is added to both `release-prod.yml` and `release-canary.yml`. It mirrors the existing `release-linux` job but:

- Runs on a **native arm64 runner** (`ubuntu-22.04-arm`) rather than cross-compiling — the native modules (`better-sqlite3`, `node-pty`, `@parcel/watcher`) are rebuilt from source and cross-compiling them on x64 is unreliable.
- Keeps `container: ubuntu:22.04` for GLIBC parity (matches the `verify-linux.ts` ≤ 2.35 gate used by the x64 job).
- Builds the deb and rpm targets: `--platform linux --arch arm64 --targets deb,rpm`.

It also makes the existing `arch` workflow input actually control the Linux builds (previously the Linux job was hardcoded to x64 and ignored the input):

- `release-linux` (x64) runs unless `arch == 'arm64'`
- `release-linux-arm` runs unless `arch == 'x64'`
- `both` (the default) and tag pushes build both.

`finalize-release` now depends on `prepare-release` plus the build jobs and uses `if: ${{ !cancelled() && !failure() }}`, so a build job skipped by arch filtering no longer blocks finalization, while a failed `prepare-release` (or any failed build) still suppresses it.

No electron-builder config changes were needed: `build.ts` passes an explicit `createTarget(...)` map to electron-builder, which overrides the `linux.target` arch arrays in the config, so the release path is driven entirely by the `--arch`/`--targets` flags. Touching those arrays would have been dead code for releases and would have broken local `pnpm run package:linux` on x64 dev machines.

### Related issues

_None._

### Testing

- Validated both workflow YAML files parse and confirmed the job graph, `needs`, and `if` conditions.
- No unit/type/lint surface (workflow YAML only); `actionlint` was not available in the environment.
- Not yet exercised end-to-end — dispatching the release workflow requires release credentials and the org's arm64 hosted runners to be enabled.

### Notes for reviewers

- The arm64 deb/rpm upload to the GitHub draft (via electron-builder `publish: always`) and to R2. They are downloadable packages only — no auto-update manifest, which is appropriate for deb/rpm.
- **Infra prerequisite:** `ubuntu-22.04-arm` hosted runners must be available to the org (free for public repos; requires enablement for private repos).

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -43,6 +43,7 @@ jobs:
         run: node --experimental-strip-types scripts/release/prepare-release.ts --channel canary
 
   release-linux:
+    if: github.event.inputs.arch != 'arm64'
     needs: [prepare-release]
     runs-on: ubuntu-latest
     container: ubuntu:22.04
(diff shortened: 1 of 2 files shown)


arnestrickmann:
Thanks for opening this PR, @fatih-acar - this addition is on our list atm. Will get back to you with feedback timely here!

fatih-acar:
hello @arnestrickmann @jschwxrz any update regarding this one? that would allow me to avoid rebuilding from source everytime :smile: 
thanks again for the awesome work!

jschwxrz:
looking into this again 😅

jschwxrz:
hi @fatih-acar, arm64 builds are now included in the releases!

PR #2825, opened 2026-07-10
fix: parcel app crashes

Davidknp:
The error occurred due to a race condition in the`@parcel/watcher` INotify (macOS) backend which can occur when using multiple FileWatcher instances on the same path. 

The issue is fixed by using single `FSWatchService` instance per host which makes the issue much less likely to occur. Additionally the watcher is now instantiated in a subprocess with the effect that parcel crashes will no longer crash the whole app. 

As part of this the FSWatchService is now utilising the wire package and the wire package has been extended to support a fire and forget `eventStream` endpoint. Additionally utils for worker process management have been added and bundling has been improved using a` worker-manifest`.


Davidknp:
@greptileai

Davidknp:
@greptileai

PR #2827, opened 2026-07-10
fix(windows): sync title bar with app theme

janburzinski:
### Description

Synchronizes Electron's native theme with the selected Emdash theme so the Windows title bar no longer stays light while the app is in dark mode.

The native theme is applied during startup and reconciled immediately when the theme setting changes. System theme mode continues to follow the operating-system preference.

### Related issues
ENG-1815

<details>
<summary>Checklist</summary>

- [x] Docs are not needed for this behavior fix
- [x] No automated test was added because this behavior is provided by Electron's native Windows window decoration

</details>

### DIFF
--- a/apps/emdash-desktop/src/main/app/window.ts
+++ b/apps/emdash-desktop/src/main/app/window.ts
@@ -1,5 +1,5 @@
 import { join } from 'node:path';
-import { BrowserWindow } from 'electron';
+import { BrowserWindow, nativeTheme } from 'electron';
 import devIcon from '@/assets/images/emdash/emdash-dev.png?asset';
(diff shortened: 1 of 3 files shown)


arnestrickmann:
@greptile

arnestrickmann (commented):
Could we guard this with `process.platform === 'win32'` to avoid unintentionally changing native theme rendering on macOS and Linux?

janburzinski:
ofc :) @arnestrickmann

arnestrickmann (approved):

arnestrickmann:
Thanks, @janburzinski - also for addressing my feedback!

PR #2831, opened 2026-07-11
fix(tasks): clarify archive and delete effects

janburzinski:
### Description

Clarifies the archive and delete copy, including what is preserved. No lifecycle behavior changes.

### Related issues

ENG-1802


arnestrickmann (approved):
thanks for making these more detailed - I'll let the user know that we extended the descriptions

arnestrickmann:
I think it would be better if we hide the text behind info icons that the user can hover next to each of the options - then, the tooltips will include the content: 

CleanShot 2026-07-21 at 16 17 44@2x

this would make it a lot cleaner - otherwise, looks good. happy to merge after you've made this change.

PR #2842, opened 2026-07-12
feat(mcp): add integrations.sh server search

janburzinski:
### Description

Adds live MCP server search backed by [integrations.sh](https://integrations.sh) to the MCP view. When the search box has 2+ characters, the main process queries the integrations.sh domain index, loads the matching domains' MCP surfaces, and shows installable entries in a new "integrations.sh" section — deduped against the bundled catalog and already-installed servers (by key and by server URL).

Implementation notes:

- `McpService.searchIntegrationsSh` filters/sorts the domain index by popularity, fetches up to 12 domain manifests in parallel, and caches both the domain index and per-query results for 5 minutes (with stale-cache fallback on failure). A single failing domain fetch degrades to just that domain's missing entries instead of dropping the whole result set.
- Remote payloads are validated before use: domains/surfaces are shape-checked, install configs are restricted to http(s) URLs or stdio commands, and `docs` URLs are protocol-validated with a fallback to the generated integrations.sh page.
- Renderer: debounced (300ms) React Query search in `useMcps`, new section in `McpView`, remote `iconUrl` support in `McpServerIcon`/`McpCard`, and a new `integrations_sh` source for the `mcp_server_added` telemetry event.
- `getProviders` now ensures agent dependencies are probed before mapping providers, and manual refresh re-reads the shell env.

https://cap.link/3zyk292bmnt6r25

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/apps/emdash-desktop/src/main/core/mcp/controller.ts
+++ b/apps/emdash-desktop/src/main/core/mcp/controller.ts
@@ -25,6 +25,16 @@ function mapProviders(): McpProvidersResponse[] {
 }
 
 export const mcpController = createRPCController({
+  searchIntegrationsSh: async ({ query }: { query: string }) => {
+    try {
+      const data = await mcpService.searchIntegrationsSh(query);
+      return { success: true, data };
+    } catch (error) {
+      log.error('Failed to search integrations.sh:', error);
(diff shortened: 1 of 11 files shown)


janburzinski:
@greptileai

PR #2843, opened 2026-07-13
docs(providers): fix catalog totals

janburzinski:
## Summary

- correct the provider catalog total from 35 to 34
- remove the unregistered Gemini provider from the provider and ACP lists
- correct the ACP-capable provider total from 22 to 21

## Why

The documentation had drifted from the live plugin registry after the Gemini plugin was removed.


arnestrickmann (approved):

arnestrickmann:
Thanks!

PR #2845, opened 2026-07-13
fix(antigravity): use prompt flag for initial messages

janburzinski:
Uses `-p` when launching Antigravity with an initial prompt, because `-i` requires a TTY.


janburzinski:
@greptileai

arnestrickmann:
feedback in discord ->

PR #2849, opened 2026-07-13
fix(chat): prevent provider name clipping

janburzinski:
## Summary

Fixes ENG-1832 by keeping the provider/model trigger from shrinking and giving its clipped text label enough line height for descenders.

## Root cause

The model selector could shrink under toolbar pressure, and its text label used a unit line height inside an overflow-clipped span, which cut off descenders such as `g`.

## screenshot
after:
Screenshot 2026-07-13 at 14 36 27


arnestrickmann (approved):
thanks!

PR #2860, opened 2026-07-14
fix(ui): unify tooltip horizontal padding

janburzinski:
### Description

Keep tooltip horizontal padding consistent when the content includes keyboard shortcut badges. This removes the conditional right-padding override that made shortcut tooltips asymmetric.

### Related issues

ENG-1844

### Screenshot

Screenshot 2026-07-14 at 12 15 22
Screenshot 2026-07-14 at 12 15 12


arnestrickmann (approved):

arnestrickmann:
thanks!

PR #2873, opened 2026-07-15
feat(chat): add fast mode control

janburzinski:
## Summary

- expose provider fast-mode config through the ACP session model and runtime
- add a Fast mode selector to the chat composer model menu
- cover config derivation and provider config updates with focused tests

## Why

Codex and Claude expose fast mode as provider-owned ACP config options, but Emdash previously ignored those categories, so users could not view or change the setting from chat.

## Demo

https://cap.link/mam1dvkjp6yz93v


janburzinski:
@greptileai

PR #2874, opened 2026-07-15
fix(telemetry): register task lifecycle tracking

janburzinski:
### Description

Register the existing task telemetry module from the Electron main-process entry point. Without this side-effect import, its lifecycle listeners were never installed, so current releases stopped sending `task_created`, initial `issue_linked_to_task`, and `task_provisioned` events.

Add focused regression coverage for listener registration and all three event paths.

### Related issues

[ENG-1876](https://linear.app/general-action/issue/ENG-1876/investigate-posthog-issue-tracker-usage-drop-off-and-create-task-modal)

<details>
<summary>Checklist</summary>

- [x] Docs are not required for this internal telemetry registration fix

</details>


arnestrickmann (approved):

arnestrickmann:
Thanks for bringing it back!

PR #2875, opened 2026-07-15
feat(settings): add hook injection opt-out

### LINKED ISSUE #1944: [feat]: Add setting to disable automatic hook config injection (.claude/settings.local.json, .codex/config.toml)
### Description

Every time emdash spawns an agent session, `HookConfigWriter.prepareHookConfig()` writes
notification hook configs into the project/worktree cwd:

- `.claude/settings.local.json` — adds `Notification` and `Stop` hook entries pointing at
  emdash's local POST endpoint (carries `EMDASH_HOOK_PORT` / `EMDASH_HOOK_TOKEN` markers).
- `.codex/config.toml` — sets `notify = [...]` to the bundled notify script.
- `.opencode/plugins/emdash-notifications.js`, `.pi/extensions/emdash-hook.ts` for those
  providers.

There is currently **no setting that disables this injection**. The two related toggles
do not gate it:

| Setting | Effect |
|---|---|
| `notifications.enabled` | Only suppresses `Notification.show()` (OS toast). The hook file is still written. |
| `localProject.writeAgentConfigToGitIgnore` | Only controls whether the path is added to `.gitignore`. The hook file is still written. |

### Why this is a problem

- It silently mutates user-owned config files (`.claude/settings.local.json` is meant for
  user-specific permissions, not tool-managed hooks).
- It runs on every fresh `LocalConversationProvider` instance — disabling per-task caching
  via `preparedHookProviders` is per-instance and resets across tasks/sessions.
- Users who don't want their projects coupled to a running emdash daemon (e.g. they share
  the worktree with other agents, run agents directly, or audit their repo for unexpected
  files) have no opt-out short of patching the asar or making the files immutable
  (`chmod 444` / `chflags uchg`), both of which produce silent write failures emdash logs
  as warnings.

### Reproduction

1. emdash 1.1.10, macOS.
2. Open any project, start a Claude or Codex task.
3. `cat .claude/settings.local.json` — the `hooks` block now contains an entry referencing
   `EMDASH_HOOK_PORT`. Same for `.codex/config.toml` (`notify = ...`).
4. Toggle off `notifications.enabled` and `writeAgentConfigToGitIgnore` from Settings.
5. Start a new task — the hook entries are re-written.

### Source pointers (1.1.10, packed `out/main/index.js`)

- `class HookConfigWriter` — `writeClaudeHooks` / `writeCodexNotify` / `writeForProvider` /
  `writeAll`.
- Caller: `LocalConversationProvider.prepareHookConfig` invoked from `startSession`. The
  guard only checks `writeAgentConfigToGitIgnore` for `.gitignore` writes, never gates the
  config file write itself.

### Proposed fix

Add a new setting (e.g. `localProject.injectAgentNotificationHooks: boolean`, default
`true` to preserve current behavior) and short-circuit `prepareHookConfig` when it's
`false`. Optionally per-provider granularity (`{ claude: false, codex: true }`) for users
who want notifications for one provider but not another.

A "clean up existing emdash hook entries from this project" action in Settings would also
be helpful, since users who toggle this off will have stale entries from previous runs.

### emdash version
1.1.10

### OS
macOS (Darwin 25.3.0)

rasitakyol:
## Description

Adds a default-on repository setting that lets users opt out of Emdash-managed agent notification hooks for local tasks.

When disabled, Emdash skips persistent config/plugin and `.gitignore` writes, does not modify Kimi's inline `--config`, and omits `EMDASH_HOOK_*` plus `EMDASH_PTY_ID` from the spawned agent environment. Existing managed entries are left untouched, so toggling the setting does not destructively rewrite user config.

The injection preference is kept separate from installation success: default-on sessions retain the previous environment behavior even if refreshing a hook file fails.

## Related issues

Closes #1944

## Testing

- Desktop Node suite: 1,981/1,981 passed.
- Local conversation regression suite: 33/33 passed.
- Kimi command tests: 2/2 passed.
- Desktop, core, and plugins typecheck, lint, and format checks passed.
- `git diff --check` passed.
- A concurrent full-app run had one unrelated `IssueSelector` 5-second timeout; that test passed immediately when rerun alone.

## Screenshot/Recording

Not included; the change adds one repository settings row and is covered by build/type checks.

## Checklist

- [x] I have read the contributing guidelines
- [x] I have added tests that prove the fix is effective
- [x] I have run the relevant checks locally
- [x] I have searched for existing pull requests and found no duplicate

### DIFF
--- a/apps/emdash-desktop/src/main/core/agent-hooks/hook-config-service.ts
+++ b/apps/emdash-desktop/src/main/core/agent-hooks/hook-config-service.ts
@@ -39,8 +39,9 @@ async function ensureGitIgnoreEntries(taskPath: string, entries: string[]): Prom
  * conversation spawn. Writes are idempotent (small-file merges), so re-writing
  * before every spawn removes the "config got cleaned mid-task" failure mode.
  *
- * Returns true if hooks are available for this provider (i.e. hook env vars
- * should be injected into the PTY spawn).
+ * Keeps the user's injection preference separate from installation success.
+ * A failed refresh must not change the default-on spawn environment, while an
+ * explicit opt-out must disable config writes and runtime hook injection.
  */
 export async function ensureHooksInstalled({
   providerId,
@@ -48,9 +49,14 @@ export async function ensureHooksInstalled({
 }: {
   providerId: string;
   taskPath: string;
-}): Promise<boolean> {
+}): Promise<{ hooksAvailable: boolean; injectionEnabled: boolean }> {
+  let injectionEnabled = true;
+
   try {
     const localProjectSettings = await appSettingsService.get('localProject');
+    injectionEnabled = localProjectSettings.injectAgentNotificationHooks ?? true;
+    if (!injectionEnabled) return { hooksAvailable: false, injectionEnabled };
+
     const writeGitIgnoreEntries = localProjectSettings.writeAgentConfigToGitIgnore ?? true;
 
     const plugin = getPlugin(providerId);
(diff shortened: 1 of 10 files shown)


arnestrickmann:
Thanks for the PR, @rasitakyol - we'll review and follow up with feedback here timely.

PR #2880, opened 2026-07-15
fix(sidebar): keep divider visible with open tabs

janburzinski:
## Summary

- keep the left sidebar divider visible when tab content is open
- render the divider on the persistent sidebar surface
- keep the adjacent resize handle transparent while preserving its hover state

## Root cause

The divider was rendered by the resize handle between panels, so mounted tab content could
visually cover it. Moving the border onto the sidebar keeps the boundary stable in both empty and
open-tab states.

Fixes ENG-1879.

## Screenshot
Screenshot 2026-07-15 at 14 57 46
Screenshot 2026-07-15 at 14 57 42


arnestrickmann (approved):
thank you for the fix! I was confused by the two screenshots in the PR description as I it seems like these are not the same as in the ticket in Linear and I read them as "before" & "after".

Merging

PR #2895, opened 2026-07-16
fix(chat): avoid clipping agent mode label descenders

janburzinski:
### Description

Increase the permission-mode trigger's line height so labels with descenders, such as `Agent`, are not clipped at the bottom. This mirrors the existing model-label treatment and keeps the change scoped to the affected trigger.

Agent mode label with fully visible descender

<details>
<summary>Checklist</summary>

- [x] Docs are not affected
- [x] A focused visual regression was manually verified; no logic behavior changed
- [x] I used [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages and the PR title

</details>


arnestrickmann:
Thanks!

arnestrickmann (approved):

PR #2897, opened 2026-07-16
fix(watch): prevent watcher resubscribe crashes

jschwxrz:
- treats dropped fsevents notifications as a resync request instead of replacing a healthy subscription
- serializes genuine watcher restarts and ignores stale callbacks during recovery
- coalesces repeated TreeResource resync requests into one active refresh and one trailing refresh
- updates the resync contract to cover recoverable event gaps


jschwxrz:
@greptileai

arnestrickmann:
Thanks @jschwxrz

PR #2906, opened 2026-07-17
fix(tasks): align issue provider placeholder inline

janburzinski:
## Summary

- render the create-task issue placeholder as one inline flex row
- keep the provider icon, provider name, and surrounding copy vertically aligned
- prevent the placeholder copy from wrapping across lines

## Screenshot
before:
before-github-inline

after:
after-github-inline


arnestrickmann (approved):
thanks!

PR #2910, opened 2026-07-18
fix(pull-requests): order CI check runs chronologically

janburzinski:
### Description

Order CI check runs by their latest timestamp instead of grouping failed runs above successful ones. The checks panel now displays newest runs first, falls back from `startedAt` to `completedAt`, and keeps checks without a valid timestamp at the end without mutating the source list.

### Related issues

[ENG-1903](https://linear.app/general-action/issue/ENG-1903/order-ci-check-runs-chronologically)

<details>
<summary>Checklist</summary>

</details>


arnestrickmann (approved):
thanks for the fix - will update the user

PR #2915, opened 2026-07-18
fix(chat-ui): remove duplicate search label

janburzinski:
### Description

Search tool rows currently render the static \`Search\` label beside provider queries that may already begin with \`Search\`, producing duplicated text such as \`Search Search for …\`.

Strip one case-insensitive leading \`Search \` prefix when building the compact search summary. Queries without that prefix remain unchanged. Regression tests cover both cases.

before:
ea74b0991379a6ae280eae96370c63f0859a8c304b042fd71133f24f4ee357ee

after:
Screenshot 2026-07-18 at 17 53 24

<details>
<summary>Checklist</summary>

- [x] Docs were not needed for this behavior-only fix

</details>


janburzinski:
@greptileai

arnestrickmann (approved):
thanks for the catch!

PR #2918, opened 2026-07-19
fix(skills): use Skills.sh capitalization

janburzinski:
Use the Skills.sh capitalization consistently across the skills UI.


arnestrickmann (approved):
thanks, good catch

PR #2919, opened 2026-07-19
fix(view-state): persist task layout across reloads

janburzinski:
### Description

Fixes view-state restoration races that could reopen a previously closed agent tab, drop the currently open PI/OpenCode tab, or restore an old sidebar task order after Reload.

- Persist task-tab and sidebar snapshots immediately after user changes.
- Flush active and suspended snapshots before a normal Electron Reload, and abort Reload when persistence fails.
- Restore dedicated task-tab state even when the legacy aggregate snapshot is absent.
- Retain snapshots across suspend/reprovision while explicitly evicting them after permanent task or project deletion.
- Send IPC-cloneable sidebar ordering snapshots by copying nested MobX arrays.

### Related issues

Linear: ENG-1838

https://cap.link/1g5jg3jpzq13yqz

<details>
<summary>Checklist</summary>

- [x] Docs were not needed for this behavior-only fix
- [x] I used [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages and the PR title

</details>

### DIFF
--- a/apps/emdash-desktop/src/main/app/menu.ts
+++ b/apps/emdash-desktop/src/main/app/menu.ts
@@ -7,6 +7,7 @@ import {
   menuGiveFeedbackChannel,
   menuOpenSettingsChannel,
   menuQuitRequestedChannel,
+  menuReloadRequestedChannel,
   menuRedoChannel,
   menuUndoChannel,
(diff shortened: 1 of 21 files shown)


janburzinski:
@greptileai

PR #2932, opened 2026-07-21
fix(conversations): keep create modal open after app blur

janburzinski:
Keeps the Create Conversation modal open when returning to Emdash by applying the same window-blur guard used by New Task.

[Screen recording](https://cap.link/hq4mr8h6vep4gr8)


arnestrickmann (approved):
thanks!

PR #2935, opened 2026-07-21
fix(chat): preserve transcript text selection

janburzinski:
## Summary

- keep transcript content in visual DOM order so native text selection follows what users see
- preserve selected text nodes while assistant responses continue streaming
- keep streamed Markdown reactive when prose becomes bold, inline code, a mention, or a link
- prevent a selection drag inside a user message from triggering the card's expand behavior

## Root cause

The projected chat layout positioned transcript fragments independently and rebuilt block, line, fragment, and code-line wrappers on streaming updates. Native browser selection is anchored to concrete text nodes, so replacing those nodes cleared the selection after mouse release. The delegated user-card click handler could also interpret the click emitted after a selection drag as an expand action.

## Screen recording

[Selection behavior before the fix](https://cap.link/9w99jfryra0hnxg)

### DIFF
--- a/packages/chat-ui/src/ChatRoot.tsx
+++ b/packages/chat-ui/src/ChatRoot.tsx
@@ -1553,6 +1553,8 @@ export function ChatRoot(props: ChatRootProps) {
 
     const onClick = (e: Event) => {
       const t = e.target as HTMLElement;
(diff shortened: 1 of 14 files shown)


janburzinski:
@greptileai

PR #2943, opened 2026-07-22
fix(conversations): persist model and permission preferences

janburzinski:
### Description

Persist the selected model and permission mode instead of resetting them whenever the create-conversation or create-task UI remounts.

- share conversation preferences between both creation flows
- remember model selections per provider so provider-specific model IDs do not leak across agents
- keep the existing task-level auto-approve default as the fallback until the user chooses a preference

### Related issues

[ENG-1905](https://linear.app/general-action/issue/ENG-1905/keep-model-and-permission-mode-pocket-state)

### Screenshot/Recording

https://cap.link/st1e4rkardss2bg

<details>
<summary>Checklist</summary>

- [x] I updated docs when behavior or setup changed, or no docs were needed

</details>


janburzinski:
@greptileai

PR #2951, opened 2026-07-25
fix(ssh): report agent status for remote PTY conversations

### LINKED ISSUE #2770: [bug]: agent status indicators never appear for SSH (remote) PTY conversations
### Bug Description

On SSH (remote) projects, the per-task agent status indicator in the sidebar and task list never appears. That is the small colored dot (or the working spinner) rendered by `AgentStatusIndicator`: working, awaiting-input, error, completed. For local projects the same indicator works. The task rows over SSH just stay blank even while the agent is actually running, waiting for input, or done.

I traced this in the code and it is not a rendering glitch, the status genuinely never leaves `idle`/`null` for remote PTY conversations. This is separate from the git +/- diff stats, which do work over SSH.

### Steps to Reproduce

1. Add a project that runs on a remote machine over SSH.
2. Create a task and start a conversation with a PTY/TUI provider (Claude Code in my case).
3. Send a prompt so the agent starts working, then let it finish or stop for input.
4. Look at the task row in the left sidebar / project task list.

Actual: no agent status dot or spinner ever shows for the remote task, in any phase (working, awaiting-input, completed, error). The row looks idle the whole time.

Expected: the same indicator behavior as a local task, a spinner while the agent works and a dot for awaiting-input / completed / error.

### emdash Version

1.1.36 (main at `b259382fd`)

### Operating System

UI on macOS, agent executing on a remote Linux host (Debian 13) over SSH.

### Additional Context

Root cause (file references against `main`):

The renderer indicator reads `taskAgentStatus` (`apps/emdash-desktop/src/renderer/features/tasks/stores/task-selectors.ts:59`), which folds the per-conversation `status` in `ConversationManagerStore.taskStatus` (`apps/emdash-desktop/src/renderer/features/tasks/conversations/conversation-manager.ts:159`). That `status` is driven by the `conversation:agent-status-changed` IPC event.

On the main side, those status transitions are produced in `agent-hooks/agent-hook-service.ts` from `AgentEvent`s (`deriveAgentStatus` at line 31, emit at line 198). For the default PTY conversation path the real start/stop/notification/error events come from the agent CLI POSTing to a local hook server. The hook server binds to loopback only (`agent-hooks/hook-server.ts:66`, `127.0.0.1`), and the hook script the agent runs targets a hardcoded loopback URL (`packages/core/src/agents/plugins/helpers/hooks.ts:33`, `http://127.0.0.1:$EMDASH_HOOK_PORT/hook`).

The gap is entirely in the SSH conversation provider:

* `LocalConversationProvider` installs the hook config and injects the hook env: `ensureHooksInstalled(...)` (`impl/local-conversation.ts:126`), `agentHookService.getPort()/getToken()` (lines 189-190), and `buildAgentEnv({ hook: { port, ptyId, token }, ... })` (lines 198-199, which sets `EMDASH_HOOK_PORT/EMDASH_PTY_ID/EMDASH_HOOK_NONCE`).
* `SshConversationProvider` does none of this. It builds the remote env as `{ ...providerEnv, ...colorEnv, ...taskEnvVars }` (`impl/ssh-conversation.ts:152` and `173-176`), with no `ensureHooksInstalled` and no `EMDASH_HOOK_*` vars. There is also no reverse SSH port-forward for the hook port, so even if those vars were exported the remote `curl http://127.0.0.1:<port>` would hit the remote box's own loopback, not the host running the hook server.

The synthetic-start fallback does not cover this either. `conversation:input-submitted` only emits a synthetic `start` when the provider has no native start hook (`agent-hook-service.ts:141`), and Claude does declare native hooks (`packages/plugins/src/agents/impl/claude/index.ts:55`, `supportedEvents: ['start','notification','stop']`), so it is gated off. On top of that the initial prompt is injected via direct `pty.write` in `impl/keystroke-injection.ts`, which bypasses `ptyController.sendInput`, so no `input-submitted` fires for the first run.

The ACP conversation path is not affected: it derives agent events in-process from the ACP state machine (`acp/production-acp-session-manager.ts:36` -> `emitAgentEvent`), so it is transport agnostic and reports status fine over SSH. The problem is specific to remote PTY/TUI conversations.

Possible fix directions:

* Wire the SSH provider like the local one and add a reverse SSH tunnel that maps a remote loopback port to the local `127.0.0.1` hook server, then inject `EMDASH_HOOK_PORT` pointing at the forwarded port (the hook payload already targets `127.0.0.1`, so nothing else in the hook layer needs to change). One tunnel per SSH connection is enough since `EMDASH_PTY_ID` disambiguates conversations.
* Or install hooks remotely and route callbacks back some other way.
* Or move remote tasks onto the ACP path where the provider supports it.

Related: generalaction/emdash#1132 (SSH/local feature parity, closed).

fiorelorenzo:
### Description

On SSH projects the agent status indicator never leaves idle, because the whole hook path is local-only. I traced this in #2770 and this is the fix for it.

Three things were missing on the remote path:

1. `ensureHooksInstalled()` always wrote through the local plugin fs (`createLocalPluginFs`), and it was only called from `LocalConversationProvider`. A remote agent would have needed its hook config on the machine running the desktop app. It now takes a `host`, and for ssh it writes the workspace config through `createRemotePluginFs`, with global-scope hooks (Claude, Codex) resolved against the remote `$HOME` via `resolveRemoteHome`.
2. The hook payload posts to `http://127.0.0.1:$EMDASH_HOOK_PORT/hook` (`packages/core/src/agents/plugins/helpers/hooks.ts`), which on a remote box is that box's own loopback. `RemoteHookTunnelManager` binds a remote loopback port with `forwardIn` and pipes accepted channels back to the hook server over the existing ssh2 client, so that URL resolves from the remote host without touching the hook layer. One tunnel per SSH connection is enough since `EMDASH_PTY_ID` disambiguates conversations, and the cache is revalidated against the live client so a reconnect rebinds.
3. `SshConversationProvider` now injects `EMDASH_HOOK_PORT` / `EMDASH_PTY_ID` / `EMDASH_HOOK_NONCE` / `EMDASH_HOOK_TOKEN` into the remote command env, pointing at the forwarded port.

If the remote refuses port forwarding (`AllowTcpForwarding no`) the vars stay unset, the manager logs a warning and caches the refusal for that connection, so behavior degrades to exactly what it is today instead of failing the spawn.

I pulled the hook var names into `buildHookEnv()` so the local PTY env and the SSH command env cannot drift apart.

Side effect worth calling out: providers whose hooks are a workspace plugin file (oh-my-pi, pi, amp, opencode, kilocode, mimocode) pass the plugin path on every spawn via `defaultArgs`, but over SSH that file was never written. Every remote oh-my-pi session started with `Failed to load extension ... Cannot find module '<workspace>/.omp/extensions/emdash-hook.ts'`. That file is now written on the remote host before the spawn.

Known limitation, not addressed here: with tmux-backed sessions a session that outlives its SSH connection keeps the old forwarded port, so status stops updating until the agent is respawned.

### Related issues

Fixes #2770

### Testing

- `pnpm run format:check`, `pnpm run lint`, `pnpm run typecheck`, all green repo-wide.
- `pnpm exec vitest run --project node` for the app: 2026 passed. The 2 failures on my box are environmental and also fail on a clean checkout of `main` here (`local-pty.integration.test.ts` needs the node-pty prebuild, `browser-webcontents-registry.test.ts`).
- New unit tests: `remote-hook-tunnel.test.ts` (binds and pipes a channel through to a real local server, ignores connections for other bound ports, one tunnel per connection, rebinds after a reconnect, null on refusal) and `hook-config-service.test.ts` (ssh host installs through the remote fs and updates the remote `.gitignore`, reports unavailable when the remote fs cannot be opened).
- End to end against a real sshd: I ran `RemoteHookTunnelManager` against an ssh2 connection to a remote Debian host, then ran the exact hook curl remotely (`curl -X POST -H "X-Emdash-Token: ..." -H "X-Emdash-Pty-Id: claude:conv-1" -H "X-Emdash-Event-Type: stop" --data-binary '{"message":"Task completed"}' http://127.0.0.1:<forwarded-port>/hook`). curl exited 0 and the local server received the request on `/hook` with the headers and body intact. That check was a throwaway, it is not in the diff since it needs a real host.

I could not exercise the full desktop flow (my remote box is headless), so a second pair of eyes on the reconnect path would be welcome.

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/apps/emdash-desktop/src/main/core/agent-hooks/hook-config-service.test.ts
+++ b/apps/emdash-desktop/src/main/core/agent-hooks/hook-config-service.test.ts
@@ -1,0 +1,119 @@
+import path from 'node:path';
+import type { IFileSystem } from '@emdash/core/files';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type { IExecutionContext } from '@main/core/execution-context/types';
+import type { IFilesRuntime } from '@main/core/runtime/types';
+import { ensureHooksInstalled } from './hook-config-service';
+
+const PLUGIN_PATH = '.omp/extensions/emdash-hook.ts';
+const PLUGIN_CONTENT = 'export default function () {}';
+const TASK_PATH = '/remote/workspace/task';
+
+const installPlugin = vi.fn(async (fs: { write(p: string, c: string): Promise<void> }) => {
+  await fs.write(PLUGIN_PATH, PLUGIN_CONTENT);
+  return [PLUGIN_PATH];
+});
+
+vi.mock('@main/core/agents/plugin-registry', () => ({
+  getPlugin: vi.fn(() => ({
+    capabilities: { hooks: { kind: 'plugin', scope: 'workspace', supportedEvents: ['stop'] } },
+    behavior: { plugins: { installPlugin } },
+  })),
+}));
+
+vi.mock('@main/core/settings/settings-service', () => ({
+  appSettingsService: { get: vi.fn(async () => ({ writeAgentConfigToGitIgnore: true })) },
+}));
+
+/**
+ * Minimal remote host: an in-memory file map plus the handful of shell commands
+ * createRemotePluginFs shells out for (mkdir/mv/rm) and the `$HOME` probe.
+ */
+function makeRemoteHost(): {
+  files: Map<string, string>;
+  host: { kind: 'ssh'; ctx: IExecutionContext; files: IFilesRuntime };
+} {
+  const files = new Map<string, string>();
+
+  const ctx = {
+    supportsLocalSpawn: false,
+    async exec(command: string, args: string[] = []) {
+      if (command === 'sh') return { stdout: '/home/remote', stderr: '' };
+      if (command === 'mv') {
+        const [from, to] = args;
+        const content = files.get(from);
+        if (content === undefined) throw new Error(`no such file: ${from}`);
+        files.delete(from);
+        files.set(to, content);
+      }
+      if (command === 'rm') for (const arg of args) files.delete(arg);
+      return { stdout: '', stderr: '' };
+    },
+    async execStreaming() {},
+    dispose() {},
+  } satisfies Partial<IExecutionContext> as unknown as IExecutionContext;
+
+  const fileSystem = {
+    async readText(target: string) {
+      const content = files.get(target);
+      return content === undefined
+        ? {
+            success: false as const,
+            error: { type: 'fs-error', path: target, message: 'ENOENT', code: 'ENOENT' },
+          }
+        : { success: true as const, data: { content } };
+    },
+    async writeText(target: string, content: string) {
+      files.set(target, content);
+      return { success: true as const, data: {} };
+    },
+    async exists(target: string) {
+      return { success: true as const, data: files.has(target) };
+    },
+  } as unknown as IFileSystem;
+
+  const filesRuntime = {
+    fileSystem: () => ({ success: true as const, data: fileSystem }),
+  } as unknown as IFilesRuntime;
+
+  return { files, host: { kind: 'ssh', ctx, files: filesRuntime } };
+}
+
+beforeEach(() => {
+  installPlugin.mockClear();
+});
+
+describe('ensureHooksInstalled', () => {
+  it('installs a workspace plugin on the remote host for ssh conversations', async () => {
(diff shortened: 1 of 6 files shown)


jschwxrz:
closing this as resolved

PR #2959, opened 2026-07-29
fix(markdown): increase checked checkbox contrast

janburzinski:
### Description
Improves the contrast between checked and unchecked task-list items rendered from Markdown.

Previously, Chromium rendered both states with the same neutral grey checkbox background because the inputs were natively disabled. Checked items were distinguishable only by the small checkmark.

This change:
- renders checked Markdown checkboxes using the existing blue info/link accent
- keeps unchecked checkboxes neutral
- preserves the non-interactive behavior using `readOnly`, `aria-disabled`, `tabIndex={-1}`, and disabled pointer events
- adds regression coverage for the checked-state styling and accessibility attributes

### Related issues
Discord

before:
image

after:
image

<details>
<summary>Checklist</summary>

</details>


arnestrickmann (approved):
thanks

PR #2967, opened 2026-07-30
fix(ui): center switch thumb vertically

janburzinski:
### Description
Fixes the vertical alignment of the default switch thumb by changing the track height from 18.4px to 18px.

The previous fractional height caused asymmetric subpixel rounding between the 16px thumb and the switch border, making the off-state appear vertically misaligned.

before:
image

after:
image

<details>
<summary>Checklist</summary>

</details>


arnestrickmann (approved):
good catch, thank you

PR #2972, opened 2026-07-31
feat(logging): record renderer and helper process crashes

chris-yoshih:
### Description

Another suggestion — independent of #2971, either can stand alone.

`render-process-gone` and `child-process-gone` have no handlers anywhere in the
repo. When a renderer, GPU, or utility process dies, no main-process code path
runs: the window white-screens or a feature quietly stops working, and
`uncaughtException` never fires, so `registerProcessErrorLogging` doesn't see it
either. `emdash.log` just ends mid-session.

That's the worst case for the failure mode #80's memory watchdog exists for. An
OS-killed renderer is exactly what leaves no trace — the evidence is a platform
crash report (a macOS `JetsamEvent`, say), outside the app, kept only if the OS
felt like it. This puts the cause in emdash's own log next to the rest of the
session, so `reason` and `exitCode` are one grep away.

`clean-exit` is dropped rather than recorded. Electron reports it for every
helper that terminates normally at shutdown, so logging it would mean a burst of
errors on every ordinary quit. For a renderer death the report also carries the
`WebContents` URL, which separates the app window from a browser-pane webview —
a page in the in-app browser crashing says nothing about emdash.

**Shape.** `reportRenderProcessGone` / `reportChildProcessGone` are pure
functions returning a message plus fields, or `null` for a clean exit — no
Electron import, so the filtering and field selection are directly testable.
`registerProcessCrashLogging(app, log)` is the wiring, typed against a narrowed
`ProcessCrashEventSource` interface rather than `Electron.App` so a test can
hand it a stub emitter. Same split as `dev-worktree-profile.ts` and
`reap-dev-profiles.ts`. `index.ts` gains one import and one call, next to the
existing `registerProcessErrorLogging`.

One thing worth flagging: reading `contents.getURL()` on a `WebContents` that is
already tearing down can throw, so it's wrapped and the URL is simply omitted
when unavailable — the crash still gets logged.

### Related issues

None filed. It came up while reading #80 — the watchdog catches the run-up to a
memory kill, but if the kill lands anyway there's nothing in the log to confirm
what happened.

### Testing

- `oxfmt --check`, `oxlint`, `tsgo --noEmit` on `apps/emdash-desktop` — all clean.
- `vitest run --project node` — 283 files / 2032 tests pass, including 10 new
  ones: `oom` / `crashed` / `killed` renderer reasons, `clean-exit` suppressed
  on both events, a missing URL falling back to `undefined`, child-process type
  and `serviceName` passthrough, and the registration seam driven by a stub
  emitter (both handlers registered, both firing, clean exits producing no log
  call).
- Not run: the full workspace `pnpm run test` across every project. No manual
  crash reproduction — I didn't want to guess at forcing an OOM kill on a real
  session, so the Electron event payloads are taken from the documented shapes
  and exercised through the stub rather than from a live crash. If you'd rather
  see a real one in a log before taking this, that's fair.

n/a — log output only.

<details>
<summary>Checklist</summary>

- [ ] I updated docs when behavior or setup changed — no docs describe log contents
- [x] I used Conventional Commits for commit messages and the PR title

</details>

### DIFF
--- a/apps/emdash-desktop/src/main/app/process-crash-logging.test.ts
+++ b/apps/emdash-desktop/src/main/app/process-crash-logging.test.ts
@@ -1,0 +1,158 @@
+import { describe, expect, it, vi } from 'vitest';
+import {
+  registerProcessCrashLogging,
+  reportChildProcessGone,
+  reportRenderProcessGone,
+  type ChildProcessGoneDetails,
+  type ProcessCrashEventSource,
+  type ProcessGoneDetails,
+} from './process-crash-logging';
+
+const APP_URL = 'app://emdash/index.html';
+
+describe('reportRenderProcessGone', () => {
+  it('reports an out-of-memory kill with the window it took down', () => {
+    const report = reportRenderProcessGone({ reason: 'oom', exitCode: 0 }, APP_URL);
+    expect(report).toEqual({
+      message: 'Renderer process gone',
+      fields: { reason: 'oom', exitCode: 0, url: APP_URL },
+    });
+  });
+
+  it('reports a crash with its exit code', () => {
+    expect(reportRenderProcessGone({ reason: 'crashed', exitCode: 133 }, APP_URL)?.fields).toEqual({
+      reason: 'crashed',
+      exitCode: 133,
+      url: APP_URL,
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


jschwxrz:
closing as outdated

PR #2986, opened 2026-08-11
fix(automations): reliably deliver prompts to unattended agents

BTM-cmd:
## Summary

- route unattended PTY automations through ACP when the provider can only receive the initial prompt via synthetic keystrokes
- preserve explicit ACP and normal argv-based PTY providers
- verify Kimi receives the complete assignment as the structured first `initialQueue` turn and that `startSession` is called

## Root cause

Automation persisted the full prompt, provisioned the task, and marked the run launched. Kimi PTY, however, starts without a prompt argument and depends on delayed synthetic paste/Enter into its TUI. A background card cannot recover if that injection is missed, leaving a running `kimi-code` process with no work. ACP sends the prompt as a protocol turn and acknowledges session startup.

## Checks

- `pnpm run build`
- `pnpm --filter @emdash/emdash-desktop run typecheck`
- `pnpm --filter @emdash/emdash-desktop run lint`
- `pnpm --filter @emdash/emdash-desktop exec vitest run src/main/core/automations/actions/taskCreate.test.ts --project node` (16 passed)
- `git diff --check`

## Risk

This changes transport selection for background automations configured as PTY only when the provider advertises both keystroke-only prompt delivery and ACP support. Interactive/manual task behavior is unchanged.

### DIFF
--- a/apps/emdash-desktop/src/main/core/automations/actions/taskCreate.test.ts
+++ b/apps/emdash-desktop/src/main/core/automations/actions/taskCreate.test.ts
@@ -369,6 +369,36 @@ describe('executeTaskCreate', () => {
     });
   });
 
+  it('uses ACP for unattended providers whose PTY prompt requires keystroke injection', async () => {
+    await executeTaskCreate(
+      {
+        ...automation,
+        conversationConfig: {
+          prompt: 'Complete the full assignment',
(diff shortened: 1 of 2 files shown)


jschwxrz:
closing as outdated

PR #3031, opened 2026-08-24
fix(chat-composer): use MCP icon for server count

jschwxrz:
- replaces the generic plug icon with the shared McpIcon
- aligns the chat input mcp indicator with the app settings tabs


jschwxrz:
@greptile review

PR #3046, opened 2026-08-28
fix(exclude): support leading glob pattern **/ when checking git exclude entries (#3033)

### LINKED ISSUE #3033: [bug]: Auto-update .gitignore appends a duplicate entry when the path is already ignored via a leading **/ glob
## Bug Description

`ensureGitIgnoreEntries` decides whether a hook-config path is already ignored with a hand-rolled pattern matcher that does not understand a **leading** `**/`. So a repo whose `.gitignore` already contains

```
**/.claude/settings.local.json
```

is not recognised as covering the entry `.claude/settings.local.json`, and Emdash appends a redundant duplicate line to `.gitignore` in **every new worktree**. That dirties a tracked file on every task, so each worktree starts with a spurious `git status` change and the line shows up in diffs and PRs unless it is manually stripped.

The matcher (from `app.asar`, v5.5.2):

```js
const isIgnored = (entry) => {
  const norm = entry.replace(/^\/+/, "");
  return existingLines.some((raw) => {
    const p = raw.replace(/^\/+/, "");
    if (p === norm) return true;
    if (p.endsWith("/"))   return norm.startsWith(p);
    if (p.endsWith("/**")) return norm.startsWith(p.slice(0, -2));
    return false;
  });
};
```

Exact matches and trailing `/` / `/**` are handled; a leading `**/` is not.

## Steps to Reproduce

1. In a git repo, put `**/.claude/settings.local.json` in `.gitignore` and commit it.
2. Add that repo as an Emdash project, with **Auto-update .gitignore** (`writeAgentConfigToGitIgnore`) left at its default of on.
3. Create a task/worktree using the Claude provider.
4. `git diff .gitignore` in the new worktree.

**Actual** — `.gitignore` has a second, redundant line appended:

```diff
  **/.claude/settings.local.json
+ .claude/settings.local.json
```

**Expected** — no change, because the existing `**/` pattern already ignores that path.

## Suggested fix

Strip a leading `**/` from the pattern before comparing, e.g.:

```js
const p = raw.replace(/^\/+/, "").replace(/^\*\*\//, "");
```

That covers the common case. More generally, matching against real gitignore semantics (or just asking git — `git check-ignore -q <path>`) would also catch `.claude/`, `*.local.json`, negations, and other patterns the current three-case check misses.

## emdash Version

5.5.2

## Operating System

Windows 11 Pro (10.0.26200)

## Additional Context

Workaround is to turn off **Auto-update .gitignore** per project, or to add the exact literal `.claude/settings.local.json` alongside the glob so the equality branch hits.

Related: #1944 (feature request for the toggle that now exists).

teddiesloco:
### Summary

Fixes #3033.

When `ensureEmdashGitExcluded` checks whether the runtime state directory is already excluded, it failed to recognize existing entries defined with a leading glob prefix such as `**/`. This caused redundant duplicate entries to be appended.

### Changes
- Normalize leading `/` and `**/` glob prefixes when checking lines in `.git/info/exclude`.

### DIFF
--- a/apps/emdash-desktop/src/core/features/projects/node/ensure-emdash-excluded.test.ts
+++ b/apps/emdash-desktop/src/core/features/projects/node/ensure-emdash-excluded.test.ts
@@ -116,6 +116,15 @@ describe('ensureEmdashGitExcluded', () => {
     expect(writeFile).not.toHaveBeenCalled();
   });
 
+  it('treats glob patterns like **/.emdash/ and leading slashes as already excluded (#3033)', async () => {
+    const { files, writeFile } = makeFs({
+      gitType: 'directory',
+      excludeContent: '**/.emdash/\n',
+    });
+    await ensureEmdashGitExcluded(files, '/repo');
+    expect(writeFile).not.toHaveBeenCalled();
+  });
+
   it('does not rewrite when the exclude read was truncated', async () => {
     // A truncated view could miss an existing entry past the cut; rewriting it would
     // drop the tail of the file, so bail instead.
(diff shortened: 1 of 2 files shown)


jschwxrz:
thank you for opening this pr, but closing this as this is no longer up to date

PR #3049, opened 2026-08-28
fix(chat-ui): render reference-style markdown links and images

eeshsaxena:
## Summary

Reference-style markdown links and images are dropped when rendering a message.

The block/inline walk in `parse.ts` handles inline `link` / `image` mdast nodes, but reference-style syntax parses to `linkReference` / `imageReference` nodes, which hit the `default` case in `phrasingsToRuns` and are ignored. The result is that the link's **visible text disappears too**, leaving a gap:

```md
See [the docs][d] for more.

[d]: https://example.com/docs
```

Before this change that rendered as `See  for more.` — "the docs" and its URL both gone.

## Fix

Add a small remark plugin (`remark-resolve-references.ts`, in the same style as `remark-inline-mentions.ts`) that rewrites `linkReference` → `link` and `imageReference` → `image` using the document's `definition` nodes, before the walk. The existing `link` / `image` renderers then handle them unchanged, so there are no changes to `phrasingsToRuns` / `blockToBlocks`.

`remark-parse` only emits reference nodes when a matching definition exists (an undefined reference like `[text][missing]` is already plain text), so every reference resolves. Identifier matching is case-insensitive per CommonMark.

## Before / after (real parser output)

```ts
parseMarkdownToBlocks("m", "See [the docs][d].\n\n[d]: https://example.com/docs")
// before: runs = [{ text: "See " }, { text: "." }]                                   ❌ "the docs" dropped
// after:  runs = [{ text: "See " }, { text: "the docs", href: "https://example.com/docs" }, { text: "." }]
```

Covers full (`[t][id]`), collapsed (`[id][]`), and shortcut (`[id]`) references, plus image references (`![alt][id]`). Inline links (`[t](url)`) are unchanged; undefined references stay literal.

## Testing

Added a `reference-style links and images` block to `parse.test.ts`. I verified the fix by running the real `parseMarkdownToBlocks` through `tsx` against all reference variants (output shown above); the assertions match. I wasn't able to run the full `pnpm` vitest suite locally (monorepo install was too slow in my environment), so the jsdom Vitest project runs in CI. The change is scoped to `packages/chat-ui/src/core/markdown/`.

### DIFF
--- a/packages/chat-ui/src/core/markdown/parse.test.ts
+++ b/packages/chat-ui/src/core/markdown/parse.test.ts
@@ -258,3 +258,35 @@ describe('blockquote paragraphs → variant: quote', () => {
     expect(prose?.depth).toBe(1);
   });
 });
+
+describe('reference-style links and images', () => {
+  const linkRun = (runs: InlineRun[]): InlineText | undefined =>
+    runs.find((r): r is InlineText => r.kind === 'text' && r.href !== undefined);
+
+  it('resolves a full reference link to its definition URL (text is not dropped)', () => {
+    const runs = firstProseRuns('See [the docs][d] for more.\n\n[d]: https://example.com/docs');
+    const link = linkRun(runs);
+    expect(link?.text).toBe('the docs');
+    expect(link?.href).toBe('https://example.com/docs');
+    expect(textSegments(runs).join('')).toContain('the docs');
+  });
+
+  it('resolves collapsed and shortcut references', () => {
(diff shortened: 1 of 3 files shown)


eeshsaxena:
This is a silent content-loss bug rather than a styling one: the link text disappears entirely, so a message that says "see [the docs][d]" renders as "see  ." with a gap, and the reader has no idea a link was there. Reference-style links show up a lot in AI/agent output and pasted docs, which is why it is worth handling. The fix stays out of the walk itself by resolving references to inline nodes in a pre-pass.

PR #3067, opened 2026-08-31
fix(ssh): disable ssh2 strictVendor so non-OpenSSH servers can forward unix sockets

### LINKED ISSUE #3064: [bug]: Remote Workspace Server never becomes reachable over Tailscale SSH — "strictVendor enabled and server is not OpenSSH or compatible version"
## Bug Description

Since upgrading to emdash 1.2.x, remote projects on a host I reach through **Tailscale SSH** no longer work. The Machines view shows the SSH connection as OK, but the Workspace Runtime stays in the "starting" state and eventually fails. When I use "Restart" on the Workspace Runtime card the app shows:

```
strictVendor enabled and server is not OpenSSH or compatible version
```

On the remote host itself the workspace-server daemon is installed, running and healthy. The problem is on the path the desktop app uses to talk to it.

### Setup

- Laptop: macOS, emdash 1.2.1 (stable). Remote host added in emdash by its Tailscale IP (`100.x.y.z`), user `ubuntu`.
- Remote host: Ubuntu (Linux 7.0, x86_64), Tailscale 1.102.3 with **Tailscale SSH enabled** (`tailscale set --ssh`). Tailscale SSH is the only way the laptop currently authenticates to this host; the OpenSSH `sshd` on the box is not used for this connection.
- Remote host worked fine with emdash 1.1.x (remote projects, worktrees, terminals) for weeks.

## Steps to Reproduce

1. Have a Linux host on a tailnet with Tailscale SSH enabled and reachable from the laptop as `ubuntu@100.x.y.z`.
2. In emdash 1.2.1, add the host (Machines → add SSH host with the Tailscale IP), or open an existing remote project on it.
3. Watch the Workspace Runtime card, or click "Restart" on it.

**Actual**

- Workspace Runtime never reaches "healthy". Restart surfaces `strictVendor enabled and server is not OpenSSH or compatible version`.
- On the remote host, the install and start steps clearly succeed. `journalctl -u tailscaled` shows every command emdash ran, executed by Tailscale SSH (no OpenSSH `sshd` involvement — `auth.log` has no sessions for these connections):

  ```
  ssh-session(...): starting non-pty command: [... --login-shell=/usr/bin/zsh ... --cmd=uname -s]
  ssh-session(...): starting non-pty command: [... --cmd=printf '%s\n' "$HOME"]
  ssh-session(...): starting non-pty command: [... --cmd=readlink -- /home/ubuntu/.emdash/workspace-server/current]
  ssh-session(...): starting non-pty command: [... --cmd=set -eu
    ... curl -fsSL --output "$install_script" -- https://releases.emdash.sh/workspace-server/0.1.2/install.sh ...
    sh "$install_script" --base-url https://releases.emdash.sh/workspace-server --version 0.1.2]
  ssh-session(...): starting non-pty command: [... --cmd=/home/ubuntu/.emdash/workspace-server/current/bin/emdash-workspace-server start --socket /home/ubuntu/.emdash/workspace-server/run/workspace.sock]
  ssh-session(...): Session complete
  ```

  The same sequence (`readlink` → `start`) repeats every few minutes as the app retries; `~/.emdash/workspace-server/install.lock` and `run/workspace.sock.lock` get created/removed on each retry.
- The daemon is up and answers locally on the host:

  ```
  $ ~/.emdash/workspace-server/current/bin/emdash-workspace-server status --socket ~/.emdash/workspace-server/run/workspace.sock
  workspace-server daemon running at /home/ubuntu/.emdash/workspace-server/run/workspace.sock (version 0.1.2, uptime 697245ms)

  $ cat ~/.emdash/workspace-server/versions/0.1.2/manifest.json
  {"name":"@emdash/workspace-server","version":"0.1.2","protocolVersion":"5.1.0","os":"linux","arch":"x64","nodeVersion":"24.14.0","ripgrepVersion":"15.2.0"}

  $ cat ~/.emdash/workspace-server/run/workspace.sock.log
  workspace-server wire socket listening at /home/ubuntu/.emdash/workspace-server/run/workspace.sock
  {"level":"warn",...,"worker":"tui-agents","error":"ExecError: tmux list-sessions -F #{session_name}\t#{session_activity} failed (exit 1)","msg":"TuiAgentsRuntime: reconcile precheck failed"}
  ```

  (The tmux warning is just "no tmux server running"; the main process and all 14 `*-runtime.mjs` workers stay alive.)

**Expected**

The Workspace Runtime connects to the already-running daemon and shows "healthy", as remote hosts did with emdash 1.1.x.

## emdash Version

1.2.1 (also reproduced with the 1.2.0 → 1.2.1 update; 1.1.40 worked on the same host)

## Operating System

Laptop: macOS (Apple Silicon). Remote: Ubuntu, Linux 7.0.0-1011-aws, x86_64, Tailscale 1.102.3.

## Additional Context

Where the error comes from, as far as I can see in the code:

- The desktop app dials the remote daemon by forwarding the unix socket over the SSH connection: `ssh-streamlocal-transport.ts` → `proxy.forwardOutStreamLocal(target.socketPath)` → `client.openssh_forwardOutStreamLocal(...)` (`core/services/ssh/node/operations/streamlocal.ts`).
- In `ssh2` (`lib/client.js`) `openssh_forwardOutStreamLocal` throws exactly this message unless `strictVendor` is `false` or the server's version string matches `/^OpenSSH_(?:(?![0-4])\d)|(?:\d{2,})/`. `strictVendor` defaults to `true`, and emdash does not set it (`grep -rn strictVendor apps packages` finds nothing).
- Tailscale's SSH server identifies itself as `SSH-2.0-Tailscale`, so the check fails before any channel is opened. The `tailscaled` log accordingly shows no forwarding attempt at all — only the `exec` sessions above.
- `WorkspaceServerProvisioner.tryDial` classifies the thrown error as `unreachable`, which triggers install → `start` → `waitUntilReady` → the same dial → `daemon-start-failed`. That matches the observed loop and the misleading "could not start" state even though the daemon is running.

For reference, Tailscale's SSH server (`ssh/tailssh/tailssh.go`) registers a `direct-streamlocal@openssh.com` channel handler, so unix-socket forwarding may work once `ssh2` is allowed to try it (`strictVendor: false`). I have not been able to verify that on 1.102.3 because the client never gets that far.

Happy to test a canary build against this host.

fatih-acar:
### Description

`ssh2` gates its OpenSSH extension methods behind `strictVendor`, which defaults to `true` and
only lets a request through when the server's version banner matches
`/^OpenSSH_(?:(?![0-4])\d)|(?:\d{2,})/` (`ssh2/lib/client.js`). The desktop app reaches a remote
workspace server by forwarding its unix socket over SSH — `ssh-streamlocal-transport.ts` →
`proxy.forwardOutStreamLocal(...)` → `client.openssh_forwardOutStreamLocal(...)` — so on any
non-OpenSSH server that call failed client-side before a channel was ever opened.

Tailscale SSH announces itself as `SSH-2.0-Tailscale`, so remote hosts reached that way failed
every dial with `strictVendor enabled and server is not OpenSSH or compatible version`. Exec
sessions go through a normal `session` channel and were unaffected, which is why the SSH
connection showed as OK in Machines while the Workspace Runtime looped install → start → dial and
settled on a misleading `daemon-start-failed`, even with a healthy daemon on the host.

This sets `strictVendor: false` on the connect config so the server decides. A server that
genuinely does not implement `direct-streamlocal@openssh.com` still fails — just with its own
channel-open failure instead of a client-side guess from a version string.

`strictVendor` also guards `openssh_noMoreSessions`, `openssh_forwardInStreamLocal`,
`openssh_unforwardInStreamLocal`, and `hostKeysProve`; none of those are called from this app
today, so the blast radius is the streamlocal path.

### Related issues

Fixes #3064

### Testing

`pnpm run format`, `pnpm run lint`, `pnpm run typecheck` (all 9 projects), and the app's `node`
Vitest project (2954 passed, 1 skipped). Added a unit test asserting the flag is set for both
persisted and transient connections.

Verified end to end against a real Tailscale SSH host (Tailscale 1.102.3, banner confirmed as
`SSH-2.0-Tailscale`) by running an echo server on a remote unix socket and dialing it through the
production path — `resolveSshConnectConfig` → `SshConnectionManager` → `SshClientProxy.forwardOutStreamLocal`:

- Without the change: `strictVendor enabled and server is not OpenSSH or compatible version`,
  thrown at `streamlocal.ts:32` — the exact error from the issue.
- With the change: channel opens and data round-trips through the forwarded socket.

That also answers the open question in the issue: Tailscale SSH 1.102.3 does implement
`direct-streamlocal@openssh.com`.

<details>
<summary>Checklist</summary>

</details>

### DIFF
--- a/apps/emdash-desktop/src/core/services/ssh/node/connect/resolve-ssh-connect-config.test.ts
+++ b/apps/emdash-desktop/src/core/services/ssh/node/connect/resolve-ssh-connect-config.test.ts
@@ -808,4 +808,15 @@ describe('resolveSshConnectConfig', () => {
     expect(readFiles).toEqual([expect.stringContaining('/.ssh/corp_ed25519')]);
     expect(result.config.privateKey).toBe('ALIAS KEY');
   });
+
+  it('disables strictVendor so OpenSSH extensions work against non-OpenSSH servers', async () => {
+    const transient = await resolveSshConnectConfig(
+      { kind: 'transient', config: { ...baseConfig(), password: 'secret' } },
+      deps()
+    );
+    const persisted = await resolveSshConnectConfig({ kind: 'persisted', row: row() }, deps());
+
+    expect(transient.config.strictVendor).toBe(false);
+    expect(persisted.config.strictVendor).toBe(false);
+  });
 });
--- a/apps/emdash-desktop/src/core/services/ssh/node/connect/resolve-ssh-connect-config.ts
+++ b/apps/emdash-desktop/src/core/services/ssh/node/connect/resolve-ssh-connect-config.ts
@@ -98,6 +98,13 @@ export async function resolveSshConnectConfig(
     host,
     port,
     username,
+    // ssh2 gates the OpenSSH extensions we rely on (notably
+    // direct-streamlocal@openssh.com, used to reach the workspace-server unix socket)
+    // behind a version-string match on the server banner. Non-OpenSSH servers that do
+    // implement those extensions — Tailscale SSH announces itself as SSH-2.0-Tailscale —
+    // would otherwise be rejected client-side before a channel is ever opened. Let the
+    // server decide: an unsupported extension still fails, just with its own error.
+    strictVendor: false,
     readyTimeout:
       resolved?.connectTimeout !== undefined
         ? resolved.connectTimeout * 1000


jschwxrz:
hi, thanks for opening this pr! but will close it as I already merged a fix for this

PR #3094, opened 2026-09-02
fix(tui): recover restored sessions after restart

### LINKED ISSUE #3093: [bug]: Restored TUI tabs can become unresponsive after app restart
### Bug Description

After quitting and reopening Emdash, some restored TUI conversation tabs display their previous terminal contents but no longer accept input. The failure is provider-independent and has been observed with multiple TUI agents.

That’s how some tabs looks like after app restart. It used to be active with some work history before the app was closed and now it’s unresponsive:
Image

### Steps to Reproduce

1. Start several PTY/TUI agent conversations with tmux disabled.
2. Quit and reopen Emdash.
3. Select the restored conversation tabs and type into them.

**Actual:** Some tabs retain stale terminal output while their backend session is missing. Typed input is silently discarded.

**Expected:** Selecting a restored tab should rehydrate/resume its session. If recovery fails, the UI should expose the failure instead of presenting an apparently interactive terminal.

### Technical Findings

- The TUI connector ignores unsuccessful `sendInput` results, unlike the terminal connector.
- Conversation tabs do not follow the terminal-tab pattern of rehydrating a disconnected session on activation.
- The file-backed JSON KV store retains pre-serialization values in memory, including nested `undefined` properties; later session-intent validation can therefore fail even though the on-disk JSON is valid.
- macOS can report an absent tmux server as `error connecting to ... (No such file or directory)`, which is not recognized by the current expected-failure matcher. This vetoes reconciliation and produces repeated sweep warnings.

### Suggested Direction

Follow existing repository patterns:

- normalize file-backed KV values through JSON serialization, as the DB-backed KV stores do;
- rehydrate inactive TUI sessions on tab activation, as terminal tabs do;
- handle and log TUI input failures rather than discarding results;
- recognize the macOS no-server tmux error and cover it with tests.

### Version

v1.2.2

mykola42:
## Summary

- retry failed TUI hydration when an inactive restored tab becomes active
- normalize file-backed KV values after serialization so optional fields cannot poison later validation
- recognize the macOS tmux missing-socket response as an expected empty-server state
- log failed TUI input delivery instead of silently discarding it
- add regression coverage for restore retry, persistence normalization, tab activation, and tmux detection

## Validation

- formatting applied to changed files before the final lifecycle edit
- git diff whitespace checks pass
- full repository checks deferred to GitHub CI because the required Node runtime is not installed locally

Closes #3093

### DIFF
--- a/apps/emdash-desktop/src/core/features/conversations/api/browser/conversation-manager.ts
+++ b/apps/emdash-desktop/src/core/features/conversations/api/browser/conversation-manager.ts
@@ -24,6 +24,7 @@ import {
   type ConversationEvent,
   type CreateConversationParams,
 } from '@core/primitives/conversations/api';
+import { log } from '@core/primitives/logging/browser/logger';
 import { makePtySessionId } from '@core/primitives/pty/api';
 import { getConversationsClient } from './client';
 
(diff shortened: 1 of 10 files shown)


jschwxrz:
lgtm, thanks!

jschwxrz (approved):

PR #3111, opened 2026-09-03
feat(terminals): open links in the Emdash browser on option-click

tkohout:
### Description

In an agent/terminal pane, ⌥-click on a URL now opens it directly in Emdash's in-app browser pane, with no confirmation dialog. Plain left-click keeps the existing dialog.

- `open-external-link.ts`: the in-app path of `confirmOpenExternalLink` is extracted into `openLinkInEmdashBrowser(url): boolean` (false when there is no task view to host a pane, or the link is not http/https). New `openExternalLinkFromMouseEvent(event, url, onError)` opens in-app when `altKey` is held and a task pane is available, and otherwise falls through to the dialog.
- `pty.ts`: both terminal link handlers (the OSC 8 `linkHandler` and `WebLinksAddon`) route through the new helper. The `isPrimaryMouseButton` guard is unchanged; ⌥ + left click still reports as primary.
- `pty.ts`: xterm's `altClickMovesCursor` (on by default) turns any alt-click into cursor-movement keys written to the PTY from the same mouseup, after the link activates. `FrontendPty.suppressAltClickCursorMove` switches the option off for the remainder of that event dispatch so an ⌥-click that opened a link does not also poke the shell or agent prompt. Alt-click cursor movement still works when the click is not on a link.

⌘ is deliberately not mapped: ⌘-click already activates file links in the terminal (`isActivationModifierPressed`). Scope is the terminal handlers only; chat links and main-process link requests keep the dialog.

Independent of the dialog layout fix (generalaction/emdash#3110); both apply cleanly on main.

### Related issues

Follow-up to #3110. Requested as a faster path for opening dev-server links from the agent pane.

### Testing

- `pnpm run format:check`, `pnpm run lint`, `pnpm run typecheck` in `apps/emdash-desktop`: pass
- `open-external-link.test.ts`: 14 tests pass (8 new): in-app open + focus, false outside a task view, false when the task has no composition yet, non-http ignored, ⌥ skips the dialog, plain click shows the dialog, ⌥ falls back to the dialog with no task view, dialog error path forwards `onError`
- `vitest --project node` for `src/core/features/workbench` and `src/core/features/terminals`: 26 files, 183 tests pass
- Manual, dev build (`pnpm run dev`), remote SSH task with Claude Code in the agent pane, driven over the Chrome DevTools port with Alt held via `keyboard.down`:
  - ⌥-click on `http://localhost:8083/admin`: no dialog rendered, a new `localhost:8083` browser tab opened with the URL (screenshot below; connection refused is expected, the port lives on the remote host)
  - bytes written to the PTY during that click, captured via `terminal.onData`: only the SGR mouse reports the TUI had requested (`ESC[<8;16;40M` etc.), no cursor-movement sequences; `altClickMovesCursor` read back as `true` afterwards
  - plain click on the same link: dialog shown
- `FrontendPty` constructs a real xterm `Terminal`, so the suppression helper is covered by the manual run rather than a node unit test.

⌥-click result, no dialog, pane opened directly:

option-click

<details>
<summary>Checklist</summary>

- [x] I updated docs when behavior or setup changed (no docs cover terminal link behavior)
- [x] I used Conventional Commits for commit messages and the PR title

</details>

https://claude.ai/code/session_013jTBcTyHoPSSBk7Y6AGFnw

### DIFF
--- a/apps/emdash-desktop/src/core/features/terminals/api/browser/pty/pty.ts
+++ b/apps/emdash-desktop/src/core/features/terminals/api/browser/pty/pty.ts
@@ -5,7 +5,7 @@ import {
   isPrimaryMouseButton,
 } from '@core/features/terminals/api/browser/pty/file-link-provider';
 import { buildTerminalFontFamily } from '@core/features/terminals/api/browser/pty/terminal-font';
-import { confirmOpenExternalLink } from '@core/features/workbench/api/browser/open-external-link';
+import { openExternalLinkFromMouseEvent } from '@core/features/workbench/api/browser/open-external-link';
 import { copyTextToClipboard } from '@core/primitives/desktop-host/browser/host-client';
 import { log } from '@core/primitives/logging/browser/logger';
 import { cssColorToHex, cssVar } from '@core/primitives/styling/browser/cssVars';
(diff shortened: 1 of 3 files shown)


jschwxrz:
closing this as not planned

PR #3119, opened 2026-09-04
feat(acp): fold contiguous tool calls into "Ran N commands" groups

tkohout:
### Description

A turn that runs several commands, reads a few files, and calls a tool or two produced one row per call. Even with #7 folding each execute row to a line, a busy turn was still a stack of a dozen lines between two sentences of prose. This folds every run of contiguous tool calls into a single "Ran 3 commands, read a file, used 2 tools" line, expandable to the individual rows.

### Before / after

Three commands in one turn. Before (with generalaction/emdash#3118 applied), one line per call:

before: three single-line rows

After, one group line:

after: one group line

Expanded, the children keep their own row renderers:

expanded group with three execute rows

A mixed run with a file read, three commands, and two MCP calls:

expanded mixed group

### Why this is better

- **One line per burst of activity, not per call.** Agents routinely fire five to twenty tool calls between two sentences. The reader wants to know that work happened and roughly what kind; the individual calls are detail on demand. This is the shape Claude desktop, Codex, and Cursor all use ("Ran 2 commands, used a tool").
- **The label carries the summary.** Counts by kind, in a fixed order, with singular forms: "Ran a command, read 2 files, ran a search, used 3 tools". A reader can tell a build-and-test burst from a research burst without expanding.
- **Nothing moves while it runs.** The group is one label line in every status, shimmering while a child is active, and the label grows as calls arrive. The tool-group unit's running preview window is removed, so children no longer pop into view and vanish as each call finishes.
- **Reads were already grouped.** The reducer folded contiguous read calls into a "read-batch" group; this generalises the same mechanism rather than adding a second one. Read-only runs keep the `read-batch` kind; mixed runs get a new `tool-run` kind.
- **Edits and subagents stay visible.** File edits keep their diff cards and subagents already own a group, so both are never folded and terminate a run on either side.

### Behaviour

- Two or more contiguous commands, reads, searches, MCP calls, web fetches, or unknown tools become one group. A lone call stays a single row. An assistant message between calls breaks the run.
- Group id derives from the first child, so it stays stable while the run grows.
- Status aggregates: running if any child runs, else error if any failed, else done.
- The read label changes from "2 file reads" to "Read 2 files" for consistency.

### Implementation notes

The fold, status, and label live in a new DOM-free `tool-runs.ts` in the core reducer and replace the read-only `wrapReadGroups`. It runs at the same two places the read fold did: top-level items and a parent's children. Chat-ui needed no new rendering; the existing tool-group unit renders the label and dispatches children to their native rows. The unit's collapsed preview window and its height var are removed.

The Codex fixture snapshot picks up three new groups ("Read 2 files, ran a search" and similar). The Claude fixture has no snapshot test.

Stacked on #3118. GitHub cannot base a cross-fork PR on another fork branch, so this PR targets `main` and its diff includes the #3117 and #3118 commits until those merge; the grouping commit is the only new change here.

### Related issues

None. Builds on #3117 and #3118.

### Testing

- `pnpm exec vitest run` in `packages/core` (225 files, 1652 tests), including the new `tool-runs.test.ts` (10 tests) and two new parser tests
- `pnpm exec vitest run` in `packages/plugins` with the refreshed Codex snapshot
- `pnpm run test` in `packages/chat-ui` and the desktop app's conversations tests
- `pnpm run format`, `pnpm run lint`, `nx typecheck` for core, plugins, chat-ui, and the desktop app
- New story `ToolRunMixed` in the tool-group stories
- Manual: dev app, Claude Code chat conversation (screenshots above)

<details>
<summary>Checklist</summary>

- [x] I updated docs when behavior or setup changed (module and story doc comments)

</details>


jschwxrz:
thank you for your contributions! great updates for chat ui!

PR #3127, opened 2026-09-05
fix(hosts): recover remote connections after sleep and network interruptions

jschwxrz:
- centralize connection health checks and recovery in one supervisor per host
- separate connection intent, runtime access and workspace server maintenance into composed services
- use scoped leases and explicit pins to maintain connections while honoring disconnect
- bind ssh operations, forwarding and provisioning to their owning lifetimes to prevent stale work
- preserve logical connections across outages and reconcile project, terminal and chat state after recovery


jschwxrz:
@greptile-apps

PR #3134, opened 2026-09-06
expand directory symlinks that point outside the worktree

### LINKED ISSUE #3076: [bug]: symlinks to directories outside the worktree are not browsable in the file tree (1.2.0 regression)
## Bug Description

A symlink whose target resolves outside the workspace root is classified as `outside-root` and is neither expandable nor openable in the file tree. It renders as a leaf node with a link icon, and opening it shows "Could not load file". The same links behaved as directories in 1.1.40.

## Steps to Reproduce

1. Create a task/worktree in any project.
2. From the worktree root: `ln -s ~/some/dir/outside shared` (target outside the worktree).
3. Open the task editor's file tree and click `shared`.

Actual: the entry cannot be expanded or opened; opening it yields "Could not load file".

Expected: a symlink to a directory behaves as a directory.

## emdash Version

1.2.2. Introduced in 1.2.0; 1.1.40 is unaffected.

## Operating System

macOS 26.6.2 (Apple silicon)

## Additional Context

- `packages/core/src/runtimes/files/node/tree/directory-reader.ts`, in `classifySymlink`, returns before the `isDirectory()` / `isFile()` classification:
  ```js
  const canonical = await realpath(absolutePath);
  if (!containsPath(rootPath, canonical)) return { target, kind: 'outside-root' };
  ```
- With `symlinkTargetKind === 'outside-root'`, `isExpandableFileEntry` and `isOpenableFileEntry` (`packages/core/src/runtimes/files/api/tree/state.ts`) both return false.
- The same containment applies in `RootPathPolicy.resolveFollowed` and `resolveExistingEntry` (`packages/core/src/runtimes/files/node/fs/path-policy.ts`), and in `enumerate.ts`, where it also overrides `includeSymlinkFiles`.
- `directory-reader.ts` is absent in v1.1.40 and present in v1.2.0.

thedhruvhegde:
closes #3076

1.2 classified any realpath outside the root as outside-root before checking if it was a directory, so those links were not expandable.

directory targets stay kind directory even when the canonical path is outside. listing falls back to the in-root symlink path when resolveFollowed rejects the follow. file symlinks that escape the root are still outside-root.

tested by: pnpm --dir packages/core exec vitest run src/runtimes/files/node/tree/directory-reader.test.ts

### DIFF
--- a/packages/core/src/runtimes/files/node/tree/directory-reader.test.ts
+++ b/packages/core/src/runtimes/files/node/tree/directory-reader.test.ts
@@ -20,9 +20,12 @@ describe('TreeDirectoryReader', () => {
     await mkdir(path.join(root, 'z-directory'));
     await writeFile(path.join(root, 'a.txt'), 'a');
     await writeFile(path.join(outside, 'outside.txt'), 'outside');
+    await mkdir(path.join(outside, 'outside-dir'));
+    await writeFile(path.join(outside, 'outside-dir', 'nested.txt'), 'nested');
     try {
       await symlink('z-directory', path.join(root, 'linked-directory'), 'dir');
       await symlink(path.join(outside, 'outside.txt'), path.join(root, 'outside-file'), 'file');
+      await symlink(path.join(outside, 'outside-dir'), path.join(root, 'outside-dir'), 'dir');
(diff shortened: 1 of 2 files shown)


jschwxrz:
closing as resolved

PR #3136, opened 2026-09-06
run ssh -G for typed hosts so system ProxyCommand still applies

### LINKED ISSUE #2729: SSH connection fails for work hosts resolved through system SSH config
I'm trying to add a corporate/internal SSH host to Emdash. The same host works from Terminal, but Emdash fails when I enter it manually.

I'm using placeholders here because the real hostname and proxy command are work-internal.

### Setup

- macOS
- Emdash 1.1.35
- The host is something like `work-host.internal.example`
- `ssh work-host.internal.example` works in Terminal
- `ssh -G work-host.internal.example` shows that OpenSSH is using a `ProxyCommand` and agent forwarding
- That config comes from a wildcard `Host *.internal.example` rule in the system SSH config, not from a concrete host entry in `~/.ssh/config`

### What happens in Emdash

When I add the host manually and test the connection, Emdash appears to try connecting directly to the hostname. The test fails with:

`getaddrinfo ENOTFOUND work-host.internal.example`

That looks like Emdash is not using the same resolved SSH config that `ssh` uses for this host.

### What I expected

I expected a way to add this as an SSH-config-backed connection by typing the host/alias and having Emdash resolve it with `ssh -G`.

This would cover corporate setups where hosts are handled by wildcard rules in `/etc/ssh/ssh_config` or included system config files, rather than concrete entries in `~/.ssh/config`.

thedhruvhegde:
## summary
typed ssh hostnames without an sshConfigAlias were skipping `ssh -G`, so wildcard / system `ProxyCommand` never applied and the client tried to resolve the hostname locally (`ENOTFOUND`).

this still uses the stored host/port/user instead of rewriting them from config. `ssh -G` is only used for proxy settings.

closes #2729

## test plan
- [ ] add a connection with a typed hostname that only matches a `Host *` / wildcard `ProxyCommand` in `~/.ssh/config` and connect
- [ ] connect with an sshConfigAlias still uses that alias for `-G`
- [ ] `pnpm --dir apps/emdash-desktop exec vitest run src/core/services/ssh/node/connect/resolve-ssh-connect-config.test.ts`

### DIFF
--- a/apps/emdash-desktop/src/core/services/ssh/node/connect/resolve-ssh-connect-config.test.ts
+++ b/apps/emdash-desktop/src/core/services/ssh/node/connect/resolve-ssh-connect-config.test.ts
@@ -29,18 +29,9 @@ function deps(overrides: Partial<SshConnectDeps> = {}): SshConnectDeps {
     readFile: async () => 'PRIVATE KEY',
     getPassword: async () => secret('stored-password'),
     getPassphrase: async () => null,
-    resolveSshConfig: async () => ({
-      hostname: 'resolved.internal',
-      user: 'resolved-user',
-      port: 2201,
-      identityFile: [],
-      identityAgent: '/tmp/resolved-agent.sock',
-      identityAgentDisabled: false,
-      identitiesOnly: false,
-      proxyCommand: undefined,
-      proxyJump: undefined,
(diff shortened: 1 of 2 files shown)


jschwxrz:
closing as resolved

PR #3137, opened 2026-09-06
resize restored tmux panes to the current tab size

### LINKED ISSUE #2867: [bug]: Restored tmux tab does not use full pane width until manually resized
## Bug Description

When restoring a tab backed by a tmux session on a remote server, the tmux session's width is smaller than the available tab area. The session does not fill the full width of the tab.

## Steps to Reproduce

1. Connect to a remote server over SSH (via Tailscale).
2. Open a tab running inside a tmux session.
3. Close/reopen or restore the tab.
4. Observe the tmux session width.

**Actual:** The restored tmux session uses less than the full width of the tab. Manually resizing the window, or dragging a side-panel separator, causes the session to re-fit and finally use the full width.

**Expected:** On restore, the tmux session fills the full width of the tab without manual intervention.

## Environment

- **OS:** macOS (client) → remote Linux server
- **Connection:** SSH over Tailscale
- **emdash Version:** 1.1.38

## Additional Context

Any manual resize event (window resize or moving a side-panel divider) triggers the correct fit, which suggests the session isn't receiving the tab's true dimensions on restore.

thedhruvhegde:
## summary
restored sessions skipped the first resize until a later calibration, so tmux often stayed at a leftover width (e.g. ~80 cols). new sessions now get `updatePtySize` immediately.

also sets `window-size latest` on new tmux sessions so the pane follows the current client instead of the first attach.

closes #2867

## test plan
- [ ] restore a tmux tab after shrinking/growing the window and confirm the pane fills the tab
- [ ] new session still sizes correctly on first attach
- [ ] `pnpm --dir apps/emdash-desktop exec vitest run src/core/services/pty/api/tmux.test.ts`

### DIFF
--- a/apps/emdash-desktop/src/core/features/terminals/browser/pty/use-pty-pane-resize.ts
+++ b/apps/emdash-desktop/src/core/features/terminals/browser/pty/use-pty-pane-resize.ts
@@ -205,7 +205,7 @@ export function usePtyPaneResize(
     const added = sessionIds.filter((id) => !prev.includes(id));
     sessionsRef.current = sessionIds;
     const dims = controllerDimsBoxRef.current!.get();
-    if (dims && added.length > 0 && hasCalibratedRef.current) {
+    if (dims && added.length > 0) {
       for (const id of added) {
         const frontendPty = getFrontendPty(id);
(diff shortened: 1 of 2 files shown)


jschwxrz:
closing as resolved

PR #3138, opened 2026-09-06
drop late xterm device-attribute replies after attach

### LINKED ISSUE #2720: SSH provider: tmux capability-probe replies (`^[[?1;2c`, `^[[>0;276;0c`) leak into the shell prompt on attach
## Summary

On the **SSH remote provider**, stray bytes like `1;2c0;276;0c` (often repeated several
times) occasionally appear **at the shell prompt** of a tmux-backed terminal pane, "out
of nowhere", while the pane is idle:

```
dev@host:~/Progetti/Sencare/api$ 1;2c0;276;0c1;2c0;276;0c1;2c0;276;0c1;2c0;276;0c
```

These are **terminal Device-Attributes replies leaking into the pane's stdin**, caused by
tmux probing terminal capabilities **on every client attach** while the reply round-trips
through the renderer (xterm.js) and back over SSH — slowly enough to miss tmux's
feature-detection read window, so tmux forwards the late reply to the focused pane.

It is **cosmetic / harmless** (inert bytes at the prompt; only a `command not found` if you
hit Enter), but it looks alarming and happens regularly on higher-latency SSH links.

## Decoding the bytes

The visible text is two CSI replies with the (invisible) `ESC[` prefix already consumed:

| Visible | Real sequence | Meaning |
|---|---|---|
| `1;2c` | `ESC[?1;2c` | **Primary Device Attributes (DA1)** reply — "VT100 w/ Advanced Video" |
| `0;276;0c` | `ESC[>0;276;0c` | **Secondary Device Attributes (DA2)** reply — terminal type 0, firmware **276** |

Firmware `276` is **xterm.js**'s signature. Confirmed in the bundled dep
(`node_modules/@xterm/xterm/lib/xterm.mjs`, `@xterm/xterm@6.0.0`), which emits exactly
`[?1;2c` and `[>0;276;0c`.

You only ever see those two CSI replies and **never** a version string, because the third
probe reply (XTVERSION, `ESC P > | … ESC \`) is a non-printable DCS — fully consistent with
the screenshots.

## Root cause

End-to-end chain (SSH provider):

1. **emdash attaches a tmux client on every PTY open.**
   `buildTmuxShellLine()` runs `… && tmux -u attach-session -t <name>`.
   → `apps/emdash-desktop/src/main/core/pty/tmux-session-name.ts:18`

2. **tmux 3.5a probes the outer terminal on every attach.** On each attach tmux writes
   **DA1 (`ESC[c`) + DA2 (`ESC[>c`) + XTVERSION (`ESC[>q`)** to its parent terminal to
   detect `terminal-features`. Verified with a local pty harness (below) — it always sends
   all three, and **pre-setting `terminal-features` does NOT suppress it**.

3. **The outer terminal is the renderer's xterm.js, which auto-replies.** The probe bytes
   stream remote → `Ssh2PtySession.onData` → IPC → renderer → `terminal.write(...)`. xterm.js
   parses the DA/XTVERSION queries and auto-generates replies (`?1;2c`, `>0;276;0c`, version).

4. **The reply round-trips the full renderer→main→ssh2→remote path.** xterm.js's auto-reply
   surfaces via `terminal.onData` → `handleTerminalInput` → `sendInput()` → IPC → main →
   `Ssh2PtySession.write()` → ssh2 channel → **remote tmux stdin**.
   → renderer: `apps/emdash-desktop/src/main/.../use-pty.ts` (`terminal.onData` at `use-pty.ts:535`, `handleTerminalInput` at `:512`)
   → main: `apps/emdash-desktop/src/main/core/pty/ssh2-pty.ts:37` (`write`)
   This path is **far slower** than a local terminal: network RTT (twice — query out, reply
   back) + IPC hops + xterm.js parse + renderer event loop.

5. **tmux's attach-time read window closes before the late reply arrives.** Over SSH the
   DA reply lands *after* tmux has left its "expecting DA reply" state. tmux then treats the
   incoming `ESC[?1;2c` / `ESC[>0;276;0c` as **ordinary pane input** and forwards it to the
   focused pane → the shell at a prompt echoes it. The XTVERSION DCS reply is non-printable,
   so only the two CSI replies are visible.

**Why intermittent / repeated:** it's a timing race between SSH RTT and tmux's read window.
One leaked set appears **per attach / refresh-client event** — opening a terminal, switching
tabs (reattach), window/pane resize, or an SSH reconnect — which is why you often see the
pair repeated N times.

### What it is NOT
- **Not the host cron / session janitor**: `tmux list-sessions` / `kill-session` / process
  reaping never attach a client, so they never trigger a probe.
- **Not a security issue**: the bytes are inert; they sit unparsed at the prompt.

## Environment

- emdash **1.1.35**
- Provider: **SSH remote** (`ssh2@^1.17.0`, `node-pty@1.1.0`), `term: xterm-256color`
  (`ssh2-pty.ts:107`)
- Renderer terminal: **`@xterm/xterm@6.0.0`**
- Remote: Debian 13, **tmux 3.5a**
- Symptom seen in a tmux-backed terminal pane (`emdash-*` session); pane idle at a bash prompt

### A) Prove tmux probes on every attach (local, deterministic)

This isolates step 2. It does **not** reproduce the leak itself (the leak needs the
renderer+SSH latency of steps 3–5), but it confirms the unconditional probe and that
`terminal-features` does not stop it.

<details><summary><code>tmux_probe.py</code></summary>

```python
import os, pty, select, time

def run(label, conf_lines):
    conf = "/tmp/_tmux_probe.conf"
    open(conf, "w").write("\n".join(conf_lines) + "\n")
    sock = "_probe_%d" % int(time.time()*1000 % 100000)
    os.system(f"tmux -L {sock} kill-server 2>/dev/null")
    os.system(f"tmux -L {sock} -f {conf} new-session -d -s s 'sleep 30' 2>/dev/null")
    pid, fd = pty.fork()
    if pid == 0:
        os.execvp("tmux", ["tmux", "-L", sock, "-f", conf, "attach-session", "-t", "s"])
        os._exit(1)
    cap = bytearray(); start = time.time()
    while time.time() - start < 2.5:
        r,_,_ = select.select([fd], [], [], 0.2)
        if fd in r:
            try: data = os.read(fd, 65536)
            except OSError: break
            if not data: break
            cap += data
            if b"\x1b[c" in data:  os.write(fd, b"\x1b[?1;2c")        # DA1 reply, like xterm.js
            if b"\x1b[>c" in data: os.write(fd, b"\x1b[>0;276;0c")    # DA2 reply, like xterm.js
            if b"\x1b[>q" in data: os.write(fd, b"\x1bP>|emdash\x1b\\")# XTVERSION reply
    os.system(f"tmux -L {sock} kill-server 2>/dev/null")
    q = []
    if b"\x1b[c"  in cap: q.append("DA1 (ESC[c)")
    if b"\x1b[>c" in cap: q.append("DA2 (ESC[>c)")
    if b"\x1b[>q" in cap: q.append("XTVERSION (ESC[>q)")
    print(f"[{label}] tmux sent on attach: {q or 'NONE'}")

run("default", [])
run("terminal-features preset", [
    "set -g default-terminal tmux-256color",
    "set -as terminal-features ',xterm-256color:RGB:256:clipboard:ccolour:cstyle:focus:title'"])
```

</details>

Output (tmux 3.5a):

```
[default] tmux sent on attach: ['DA1 (ESC[c)', 'DA2 (ESC[>c)', 'XTVERSION (ESC[>q)']
[terminal-features preset] tmux sent on attach: ['DA1 (ESC[c)', 'DA2 (ESC[>c)', 'XTVERSION (ESC[>q)']
```

### B) Reproduce the leak (real, timing-dependent)

1. Open an SSH project on a host with non-trivial RTT (or add artificial latency).
2. Enable tmux for terminals so panes run under `tmux attach-session`.
3. Open a terminal tab, leave the shell at an idle prompt.
4. Repeatedly switch away/back between tabs (forces detach/reattach) and/or resize the
   window. Each reattach/refresh fires a fresh DA probe.
5. Intermittently, `1;2c0;276;0c` appears at the prompt — one pair per leaked attach.

## Suggested directions (not prescriptive)

- **Swallow orphan DA replies in the SSH PTY path.** emdash already strips xterm.js
  auto-generated focus-event replies (`ESC[I` / `ESC[O`) before the PTY starts
  (`use-pty.ts:516-518`). A similar, bounded filter could drop **unsolicited** DA/XTVERSION
  replies (`ESC[?…c`, `ESC[>…c`, `ESC P>|…ESC\`) heading to the remote when no app requested
  them. Care: real apps do legitimately query DA, so this must be scoped (e.g. only the
  brief post-attach window, or only when the focused pane is a shell).
- **Reduce attach churn.** Keeping the tmux client attached instead of detach/reattach on
  tab focus would make the probe — and thus the race — far rarer.
- **Answer the probe locally on the remote.** Wrapping attach so DA/XTVERSION replies are
  produced on the remote side (no renderer round-trip) removes the latency that opens the
  race window.

Upstream tmux behavior (probe on every attach) is by design, so the durable fix is
emdash-side, in the SSH provider / tmux attach flow.

## Related
- #1994 (ssh2 channel backpressure in the same PTY path — `ssh2-pty.ts`)
- #2580 (detached `emdash-*` tmux sessions piling up; same attach/detach lifecycle)

thedhruvhegde:
## summary
xterm.js can emit DA replies (`ESC [ ? 0 c` / similar) after attach. those were going into the pty and showing up as `0c` / `0;0c` in the shell.

drop those sequences for 2.5s after attach. later user input is unchanged.

closes #2720

## test plan
- [ ] attach a restored tmux session and confirm no `0c` junk in the prompt
- [ ] type normally after attach
- [ ] `pnpm --dir apps/emdash-desktop exec vitest run src/core/features/terminals/browser/pty/xterm-auto-replies.test.ts`

### DIFF
--- a/apps/emdash-desktop/src/core/features/terminals/browser/pty/use-pty.ts
+++ b/apps/emdash-desktop/src/core/features/terminals/browser/pty/use-pty.ts
@@ -23,11 +23,13 @@ import {
   shouldPasteToTerminal,
 } from './pty-keybindings';
 import { getTerminalContextLink } from './terminal-context-link';
+import { isXtermCapabilityReply } from './xterm-auto-replies';
 
 const IS_MAC_PLATFORM =
   typeof navigator !== 'undefined' && /Mac|iPod|iPhone|iPad/.test(navigator.platform);
 const IS_WINDOWS_PLATFORM = typeof navigator !== 'undefined' && /Win/.test(navigator.platform);
 const LAST_SELECTION_COPY_GRACE_MS = 2_000;
+const ATTACH_CAPABILITY_REPLY_GUARD_MS = 2_500;
 
 function getRecentSelection(selection: { text: string; capturedAt: number } | null): string {
   if (!selection) return '';
@@ -458,10 +460,12 @@ export function usePty(
       });
 
       // ── Handle terminal input ──────────────────────────────────────────────
+      const attachGuardUntil = Date.now() + ATTACH_CAPABILITY_REPLY_GUARD_MS;
       const handleTerminalInput = (data: string) => {
         onActivityRef.current?.();
 
         if (!data) return;
+        if (Date.now() < attachGuardUntil && isXtermCapabilityReply(data)) return;
 
         // First-message capture
         if (!firstMessageSentRef.current && onFirstMessageRef.current) {
--- a/apps/emdash-desktop/src/core/features/terminals/browser/pty/xterm-auto-replies.test.ts
+++ b/apps/emdash-desktop/src/core/features/terminals/browser/pty/xterm-auto-replies.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from 'vitest';
+import { isXtermCapabilityReply } from './xterm-auto-replies';
+
+describe('isXtermCapabilityReply', () => {
+  it('matches focus events and xterm.js device-attribute replies', () => {
+    expect(isXtermCapabilityReply('\x1b[I')).toBe(true);
+    expect(isXtermCapabilityReply('\x1b[O')).toBe(true);
+    expect(isXtermCapabilityReply('\x1b[?1;2c')).toBe(true);
+    expect(isXtermCapabilityReply('\x1b[>0;276;0c')).toBe(true);
+    expect(isXtermCapabilityReply('\x1bP>|xterm.js\x1b\\')).toBe(true);
+  });
+
+  it('does not match ordinary typed input', () => {
+    expect(isXtermCapabilityReply('ls\r')).toBe(false);
+    expect(isXtermCapabilityReply('1;2c')).toBe(false);
+    expect(isXtermCapabilityReply('')).toBe(false);
+  });
+});
--- a/apps/emdash-desktop/src/core/features/terminals/browser/pty/xterm-auto-replies.ts
+++ b/apps/emdash-desktop/src/core/features/terminals/browser/pty/xterm-auto-replies.ts
@@ -1,0 +1,9 @@
+const FOCUS_REPLY = /^\x1b\[[IO]$/;
+const DEVICE_ATTRIBUTES_REPLY = /^\x1b\[(?:\?|>)?[\d;]*c$/;
+const XTVERSION_REPLY = /^\x1bP>\|[\s\S]*?\x1b\\$/;
+
+export function isXtermCapabilityReply(data: string): boolean {
+  return (
+    FOCUS_REPLY.test(data) || DEVICE_ATTRIBUTES_REPLY.test(data) || XTVERSION_REPLY.test(data)
+  );
+}


jschwxrz:
closing as already resolved

PR #3154, opened 2026-09-07
run workspace teardown after a runtime restart

### LINKED ISSUE #2886: [bug]: Archived tasks never run the teardown script
### Bug Description

The `.emdash.json` teardown script only runs when a live mounted workspace is terminated. Two trigger points miss it:

- **Archive** dismantles the live workspace without running the script.
- **Delete** skips it for any task with no live session in the current app run: archived tasks, and tasks never opened since launch.

A task that goes archive then delete never runs teardown at all, while the worktree and the resources it manages are destroyed. My teardown runs `xcrun simctl delete` for a simulator the setup script created; every archived task leaks its simulator.

Regressed in #2690.

### Steps to Reproduce

1. Add `.emdash.json` with `{ "scripts": { "teardown": "echo torn-down >> /tmp/teardown-proof.log" } }`
2. Create a task and open it so the workspace mounts (setup script runs, confirming wiring)
3. Archive the task
4. Delete the archived task

Actual: the teardown log was never written after archive nor after delete; the worktree was removed. Same result deleting an unarchived task after an app restart without opening it.

Expected: the teardown script runs before the task's resources are gone for good, per the original lifecycle design.

### emdash Version

v1.1.39

### Operating System

macOS (Darwin 25.5.0)

thedhruvhegde:
## summary
archive/delete only ran the `.emdash.json` teardown script while the workspace was still in the in-memory activation map. after an app restart that map is empty, so archived then deleted tasks skipped teardown and leaked simulators / other setup resources.

deactivate now runs teardown once whenever `lastActivatedAt` is set, then clears it so a second deactivate stays a no-op.

closes #2886

## test plan
- [ ] activate a workspace with a teardown script, restart emdash, archive/delete the task, confirm the script ran
- [ ] live archive still runs teardown once
- [ ] `pnpm --dir packages/core exec vitest run src/runtimes/workspace-registry/node/api/activation.contract.test.ts`

### DIFF
--- a/packages/core/src/runtimes/workspace-registry/node/activation.ts
+++ b/packages/core/src/runtimes/workspace-registry/node/activation.ts
@@ -47,8 +47,8 @@ export type WorkspaceActivationManagerOptions = {
    * single step-writer (spec: activation-scripts-via-terminals).
    */
   recordScriptStep: (id: string, script: ScriptStepScript, state: WorkspaceScriptStepState) => void;
-  /** Persists lastActivatedAt — the only durable trace of an activation. */
-  recordActivated: (id: string, at: number) => Promise<void>;
+  /** Persists lastActivatedAt — the only durable trace of an activation. Null clears it after teardown. */
+  recordActivated: (id: string, at: number | null) => Promise<void>;
   /**
    * The artifact gate (dependency gating): resolves once the background artifact copy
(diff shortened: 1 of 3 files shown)


jschwxrz:
closing as resolved

PR #3160, opened 2026-09-07
verify ssh host keys against known_hosts

### LINKED ISSUE #1360: [bug]: SSH Host Key Verification is Broken
### Bug Description

src/main/services/ssh/SshHostKeyService.ts:96-101, 137-151

The addHostKey() method stores a fingerprint string (e.g., SHA256:abc...) in the keyBase64 field. But verifyHostKey() later tries to base64-decode that fingerprint and re-hash it:

```
  // addHostKey stores fingerprint as keyBase64
  this.knownHosts.set(hostPort, { algorithm: keyType, keyBase64: fingerprint });

  // verifyHostKey tries to decode keyBase64 as a real key and hash it
  const knownFingerprint = this.getFingerprint(Buffer.from(stored.keyBase64, 'base64'));
  if (knownFingerprint === fingerprint) return 'known';  // Will NEVER match
```

Buffer.from("SHA256:...", 'base64') produces garbage, which is then hashed, producing a
completely different fingerprint. This means:
- Every previously-accepted host will appear as "changed" after restart
- Users get habituated to clicking "accept" on host key warnings
- This degrades TOFU (Trust On First Use) to no protection, enabling MITM attacks on SSH
connections

Impact: Man-in-the-middle attacks on SSH remote development sessions. Attacker could intercept code, credentials, and agent communications.

Recommendation: Fix the type confusion — either store actual key buffers in keyBase64, or compare fingerprints directly without re-hashing.

### Steps to Reproduce

.

.

### emdash Version

all

### Operating System

all

### Additional Context

_No response_

thedhruvhegde:
## summary
ssh2 was connecting with no `hostVerifier`, so any host key was accepted.

connect now checks `~/.ssh/known_hosts` (plain and hashed host entries). a changed key is rejected. an unknown host still connects (tofu) so first-time lans are not blocked.

closes #1360

## test plan
- [ ] connect to a host whose known_hosts key matches
- [ ] point a saved host at a different machine with a different key and confirm connect fails as host-key
- [ ] `pnpm --dir apps/emdash-desktop exec vitest run --project node src/core/services/ssh/node/connect/known-hosts.test.ts src/core/services/ssh/node/connect/resolve-ssh-connect-config.test.ts`

### DIFF
--- a/apps/emdash-desktop/src/core/services/ssh/node/connect/known-hosts.test.ts
+++ b/apps/emdash-desktop/src/core/services/ssh/node/connect/known-hosts.test.ts
@@ -1,0 +1,37 @@
+import { createHmac } from 'node:crypto';
+import { describe, expect, it } from 'vitest';
+import { hostKeyDecision } from './known-hosts';
+
+const key = Buffer.from('ssh-ed25519-host-key-bytes');
+const b64 = key.toString('base64');
+
+describe('hostKeyDecision', () => {
+  it('matches a known host key', () => {
+    expect(
+      hostKeyDecision('work.example', key, `work.example ssh-ed25519 ${b64}`)
+    ).toBe('match');
+  });
+
+  it('rejects a changed host key', () => {
+    expect(
+      hostKeyDecision(
+        'work.example',
+        Buffer.from('other-key'),
+        `work.example ssh-ed25519 ${b64}`
+      )
+    ).toBe('mismatch');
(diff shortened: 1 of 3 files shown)


rabanspiegel:
first 0/5 I've seen from Greptile 😂  @thedhruvhegde please don't blindly spam PRs. Appreciate you wanting to contribute.

PR #3172, opened 2026-09-08
fix(browser): route localhost links through the external-link flow

tkohout:
### Description

Clicking a `http://localhost:PORT` link in a task (chat transcript links, resource links, any `target="_blank"` anchor) opened a bare, unmanaged Electron window instead of the usual "Open in Emdash / Open in Browser" dialog, so the page could never land as an in-app browser tab in the workspace.

**Root cause.** `registerExternalLinkHandlers` decided what counts as "the app itself" with a hard-coded rule: in the packaged build, `file://` plus any `http://localhost:PORT` / `http://127.0.0.1:PORT`. That rule dates from when the packaged renderer was served over http; the renderer now loads from the `app://` origin, so the localhost carve-out only ever matched the user's own dev servers. Those URLs got `{ action: 'allow' }` from `setWindowOpenHandler`, which makes Electron spawn a new `BrowserWindow`. Every other http(s) link got `deny` plus the dialog. The same stale rule sat in the `will-navigate` handler, where a top-level localhost navigation would have replaced the app window instead of being intercepted.

**Fix.** Compare the target against the origin the main window was actually loaded from (`ELECTRON_RENDERER_URL` in dev, `app://emdash` in production) instead of pattern-matching hosts. The check lives in a small pure module, `external-link-policy.ts`, so it is unit-testable without Electron. Localhost dev servers now take the same path as every other link and can open as a browser tab in the current task.

Only `window.ts` calls `registerExternalLinkHandlers`, so its signature change (`isDev` → `rendererUrl`) has no other callers.

### Related issues

None filed.

### Testing

- `pnpm run format`, `pnpm run lint` (pre-existing warnings only), `pnpm run typecheck` from `apps/emdash-desktop`
- `pnpm exec vitest run --project node src/main/host/external-link-policy.test.ts` (6 tests: app origin vs. localhost ports, dev origin vs. other localhost ports, file/https/unparsable input)
- Manual, dev app on the isolated profile with a throwaway `python3 -m http.server 8899`: started a Claude Code chat-UI conversation in a task, asked it to reply with `http://localhost:8899/`, and clicked that link in the transcript with the mouse. Ran once with the old packaged-app rule temporarily forced on (the bug only shows in packaged builds, since the dev check compares against the Vite URL), and once with this fix.

### Screenshot/Recording

**Before** (old packaged-app rule forced on in dev): clicking the link in the chat transcript shows no dialog in the main window; the page lands in a separate 800×600 `BrowserWindow` (default Electron size) with nothing but the page.

| Main window after the click, nothing happens | The stray window (packaged app) |
|---|---|
| before main | before stray window |

**After:** the same click goes through the regular dialog, and "Open in Emdash" opens the page as a browser tab in the task.

| Dialog | Browser tab in the workspace |
|---|---|
| after dialog | after tab |

<details>
<summary>Checklist</summary>

- [x] I updated docs when behavior or setup changed (no docs describe this rule)

</details>

### DIFF
--- a/apps/emdash-desktop/src/main/host/external-link-policy.test.ts
+++ b/apps/emdash-desktop/src/main/host/external-link-policy.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it } from 'vitest';
+import { isInternalAppUrl } from './external-link-policy';
+
+const DEV_RENDERER_URL = 'http://localhost:5173';
+const PACKAGED_RENDERER_URL = 'app://emdash/index.html';
+
+describe('isInternalAppUrl', () => {
+  describe('packaged app (app:// renderer)', () => {
+    it('treats the app origin as internal', () => {
+      expect(isInternalAppUrl('app://emdash/index.html', PACKAGED_RENDERER_URL)).toBe(true);
+      expect(isInternalAppUrl('app://emdash/assets/x.js', PACKAGED_RENDERER_URL)).toBe(true);
+    });
+
+    it('does not treat localhost dev servers as internal', () => {
+      expect(isInternalAppUrl('http://localhost:3000/', PACKAGED_RENDERER_URL)).toBe(false);
+      expect(isInternalAppUrl('http://localhost:3000', PACKAGED_RENDERER_URL)).toBe(false);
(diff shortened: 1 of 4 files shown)


jschwxrz (approved):
lgtm, thank you!

Issue #1117, opened 2026-02-25
[feat]: Allow customizing the command-line for agents


emilecantin:
### Feature Summary

Allow changing the command, options and/or environment variables used to launch an agent. Also allow multiple such instances of an agent (e.g 2 copies of "Claude Code")

### Problem or Use Case

I have 2 aliases for Claude Code: claude-work and claude-personal, which both use different settings directories with separate logins for 2 different plans. I'd like to be able to customize this.

### Proposed Solution

Maybe we should have a list of "enabled agents", and allow enabling multiple instances of a "kind" of agent?

### Alternatives Considered

I noticed there's already some "execution settings" for some agents, but Claude Code shows as "not detected" so I can't edit this (probably related to #1111)

rabanspiegel:
<img width="962" height="280" alt="Image" src="https://github.com/user-attachments/assets/e8c1c738-1887-486c-94a5-720d722a6e41" />

<img width="1212" height="1608" alt="Image" src="https://github.com/user-attachments/assets/f934e996-0ba6-4fe7-b0e7-bc135803d77d" />

Thanks for flagging! You can set the CLI command for an agent under settings. This is currently limited to one command per provider. We should allow for an arbitrary amount of aliases. 

This would lead to multiple claude codes in the provider drop down. To differentiate we should allow them to have customizable tags, such as 'work' / 'personal'

Would this go in the direction you'd imagine @emilecantin ?

emilecantin:
@rabanspiegel Definitely, yes.

However as mentioned, I don't see this menu for Claude specifically, but I do see it for othe providers (sounds like #1111).

rabanspiegel:
Ok will fix detection and will allow adding aliases for the same agent

PR #3202, opened 2026-09-15
feat(agents): support multiple Claude instances via a config file

hatzlj:
### Description

Lets a user declare extra instances of the Claude provider (e.g. a second
account) in `agents.json` under Emdash's per-user data directory
(`app.getPath('userData')`, e.g. `~/.config/emdash/agents.json` on Linux),
read once at startup and registered alongside the built-in `claude` provider.
Each declared instance behaves as a fully independent agent — its own auth
status, host-dependency detection/custom CLI path, and env/extra-args in
Settings — reusing the existing generic per-provider-id machinery with no
further changes needed there.

**NOTE:** This PR is based on #3207 (contains its changes as well). So before reviewing this, it makes sense to review and merge #3207, which contains generally applicable fixes to enable this feature. 

**Implementation notes**

- `packages/plugins`: the Claude plugin is now a `createClaudeProvider({ id, name })`
  factory exposed through an `agentFactories` map (`packages/plugins/src/agents/registry.ts`),
  so a config-declared instance is built under a new id/name while reusing
  the same ACP/auth/hooks/trust/MCP behavior. Other providers aren't
  factory-ized in this PR; declaring `"extends": "codex"` today is
  correctly rejected with a warning until the same refactor is done there.
- `apps/emdash-desktop`: `agents.json` is parsed leniently — a malformed or
  id-colliding entry is dropped with a warning rather than failing the whole
  file. Each Wire runtime (`acp`, `tui-agents`, `agent-config`) spawns as its
  own child process with its own plugin registry, so the main process
  republishes the resolved instances through an env var
  (`EMDASH_CONFIGURED_AGENTS`) that every worker's entry file reads on boot
  to register the same instances into its own registry before serving
  requests.
- Follow-up fix folded into this PR: hook install/status previously resolved
  against the ambient environment only, so every instance of a plugin
  reported the same "Configured" root in Settings regardless of its own
  configured env (e.g. a per-instance `CLAUDE_CONFIG_DIR`), and hook
  installation was wired into the legacy PTY conversation path only — ACP
  conversations never installed or checked hooks at all. Both now resolve
  against each instance's own configured env.
- No live file-watching — editing `agents.json` requires an app restart,
  matching `host-settings.json`'s existing out-of-band-edit story.

### Related issues

Relates to #1117

That issue's proposed solution sketches a "list of enabled agents" UI; this PR
takes a config-file approach instead (`agents.json` in Emdash's per-user data
directory) rather than a Settings UI for creating/removing instances — see the
implementation notes above for the reasoning (mainly: it avoids a Wire CRUD
slice and DB persistence for a need a hand-edited file already covers well).
The core ask — multiple instances of one agent kind with independent
env/settings (the exact `claude-work`/`claude-personal` example in the
issue) — is fully covered: each declared instance gets its own
host-dependency detection, auth status, and env/extra-args/custom CLI path
in Settings.

### Testing

- `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` — clean, no new
  warnings.
- `pnpm run test` — full suite green, including new tests for the config
  loader, cross-process registration, hook status/installation env
  threading, and the ACP connection source.
- Manual: ran two Claude instances side by side in dev (`pnpm run dev`)
  with different `CLAUDE_CONFIG_DIR` values set per instance via Settings;
  confirmed both show up in the agent picker, get independent host-dependency
  rows, sign in independently, and now report independent hook status.

n.A.

<details>
<summary>Checklist</summary>

</details>


hatzlj:
@greptile review

PR #3207, opened 2026-09-15
fix(agents): resolve auth, MCP, and hooks against each provider instance's own env

hatzlj:
### Description

`AgentPluginHost`'s non-spawn methods — `checkAuthStatus`, `buildLoginCommand`,
`readMcpServers`/`writeMcpServers`/`removeMcpServer`, and hook install/status
— only ever resolved their config root and environment from the ambient
allowlisted environment, unlike `buildAcpSpawn`/`buildPromptCommand`, which
already merge a per-call, Settings-configured env on top. In practice this
means: a per-provider `CLAUDE_CONFIG_DIR` (or any other Settings-configured
env override) reaches the actual spawned CLI process correctly, but auth
status, CLI login, and MCP server config all silently fall back to the
ambient/default account instead. This is a pre-existing gap in the shared
plugin host — not specific to any one provider, and not introduced by any
recent change — it just becomes much more visible/important once a provider
can have more than one configured instance (upcoming in a follow-up PR).

**What changed**

- `plugin-host`: `checkAuthStatus`, `buildLoginCommand`, `readMcpServers`,
  `writeMcpServers`, and `removeMcpServer` all accept an optional per-call
  `env`, merged over the allowlisted agent env the same way `buildAcpSpawn`
  already does.
- `mcp` capability: a provider's MCP behavior can now optionally declare
  `resolveConfigRoot` (mirroring hooks' `resolveConfigRoots`), so its config
  file is read/written relative to an env-resolved root instead of the
  host's single fixed plugin fs. Providers that omit it are unchanged.
  `passthroughMcpAdapter` takes an optional resolver and exposes it.
- Claude's plugin declares `resolveConfigRoot` for `.claude.json` via
  `envConfigRoot('CLAUDE_CONFIG_DIR', '')` — home by default (where the file
  actually lives), the override directory directly when `CLAUDE_CONFIG_DIR`
  is set (where Claude Code relocates it).
- `agent-config` runtime: a new `ProviderEnvCache`, shared by the hooks,
  auth, and MCP managers, remembers the last per-provider env seen.
  `hooksStatus`/`refreshAuthStatus`/`startLogin` (which already receive env
  from the wire-controller on every call) warm it; MCP's cross-provider
  operations (global save/remove/list, which have no single caller-supplied
  env of their own) and internally triggered refreshes (post-login, cache
  reload) consult it instead of falling back to the ambient default.
- `wire-controller`: `hooksStatus`, `refreshAuthStatus`, and `startLogin` all
  resolve the calling provider's configured Settings env before forwarding
  to the agent-config worker, via a shared helper.
- ACP connection source/runtime: install hooks for a connection before it
  spawns, using the same per-connection env already passed to
  `buildAcpSpawn`. Previously hook installation was wired into the legacy
  PTY conversation path only, so ACP conversations never installed or
  checked hooks at all.

**Known limitation, not fixed here:** `trust.ts` (`.claude.json`
workspace-trust marking) has the identical gap — not addressed in this PR,
since a workspace trusted once but re-prompted under a different env is a
much smaller correctness issue than shared credentials/MCP servers. MCP's
env-awareness is also best-effort: an instance whose hooks/auth were never
queried has no warmed env yet in `ProviderEnvCache`, since the separate MCP
app-slice doesn't currently pass Settings env of its own.

### Related issues

None — this is groundwork surfaced while building support for multiple
Claude instances (see https://github.com/generalaction/emdash/pull/3202).
None of the fixes here depend on that feature; they're general correctness
fixes to the shared plugin host that benefit any provider with a
Settings-configured env override, including single-instance setups (e.g. a
custom `ANTHROPIC_API_KEY`).

### Testing

- `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` — clean, no new
  warnings, across `packages/core`, `packages/plugins`, `apps/emdash-desktop`,
  and `apps/workspace-server`.
- `pnpm run test` — full suite green: `packages/core` (235 files / 1785
  tests), `packages/plugins` (67 files / 392 tests), `apps/workspace-server`
  (62 tests), plus the affected `apps/emdash-desktop` node tests. New unit
  tests cover: `checkAuthStatus`/`buildLoginCommand` env merging,
  `readMcpServers`/`writeMcpServers` resolving against a real per-instance
  root (temp-dir-backed), `passthroughMcpAdapter`'s resolver passthrough,
  `ProviderEnvCache`, `AgentMcpConfigManager`'s env threading/reuse, the
  wire-controller's env resolution for hooks/auth/login, and the ACP
  connection source installing hooks before spawn.

n.A.

<details>
<summary>Checklist</summary>

</details>


hatzlj:
@greptile review

PR #3215, opened 2026-09-16
fix(antigravity): track working and completion status with native hooks

kdcokenny:
### Description

adds native hooks so Emdash tracks when agy is working and when it finishes. waits for `fullyIdle` before marking completion and reports errors separately.

installs under `~/.gemini/config/plugins/emdash`, preserving existing user hooks. agy 1.2.4 validates plugins under `~/.gemini/antigravity-cli` but doesn't discover them there.

### Related issues

same lifecycle hook support added for Muse in #3208.

### Testing

tested on macOS with agy 1.2.4. confirmed real turns send start and stop events through the installed hooks, then rebuilt Emdash and confirmed the working indicator and completion behavior in the app.


kdcokenny:
Hey @jschwxrz CI/CD build validation couldn’t find dist/adapters/codex-acp.cjs. Seems like a pre-existing race condition causing the failure.. any thoughts?

jschwxrz:
yes, the race is addressed as part of #3216

Issue #3198, opened 2026-09-14
Create Task silently disabled when local origin/HEAD points to a deleted remote branch


sibthinonk-cpu:
**Environment:** Emdash 1.2.4, Windows 11 Home 10.0.26200, local project (no SSH)

## Summary

After adding a local repo as a project, the "Create Task" / "New Task" buttons are permanently disabled (grayed out) with no visible error anywhere — no toast, nothing in `emdash.log`, no notification. The project page otherwise looks normal, and the project row is created in `emdash4.db` with a correct `base_ref`.

## Root cause (verified on my machine)

The repo's remote default branch had been renamed on GitHub (`main` was deleted, default is now `prod`), but the local clone's `refs/remotes/origin/HEAD` still pointed to the now-nonexistent `refs/remotes/origin/main`:

```
$ git symbolic-ref refs/remotes/origin/HEAD
refs/remotes/origin/main        # dangling — origin/main no longer exists

$ git rev-parse origin/HEAD
fatal: ambiguous argument 'origin/main': unknown revision or path not in the working tree.
```

`getDefaultBranch()` in the git runtime returns the branch name from `symbolic-ref refs/remotes/<remote>/HEAD` without checking that the target branch actually exists. Project hydration never reaches a "ready" state, so `taskHostActionAvailability` stays `disabled` and every live action (including task creation) is blocked.

Two other projects on the same machine whose `origin/HEAD` resolves to an existing branch work fine — this was the only structural difference between the broken repo and the working ones.

## Workaround / fix

```
git remote set-head origin --auto
```

then restart Emdash — Create Task works again immediately.

## Expected behavior

- Validate that the branch returned from `symbolic-ref refs/remotes/<remote>/HEAD` exists among the fetched refs; if dangling, fall back to `git remote show <remote>` (the code already has that fallback for the *missing*-symref case — `isMissingSymbolicRef` — but not for the *stale*-symref case).
- Surface the disabled reason in the UI. A repo in this state gives the user no clue what is wrong; the disabled button looks like a plan/quota limit.

## Repro

1. Clone a repo, then on the remote rename/delete the default branch and set a new one.
2. `git fetch --prune` locally (prune removes the stale remote-tracking branch but leaves `origin/HEAD` dangling).
3. Add the repo as an Emdash project → the project adds fine, but task creation is disabled with no error.

PR #3219, opened 2026-09-16
fix(git): validate remote HEAD before returning the default branch

kdcokenny:
### Description

after a remote branch is renamed and the old ref is pruned, `origin/HEAD` can still point to the deleted branch. `getDefaultBranch()` trusted that pointer and returned the old name.

this checks that the target resolves to a commit before using it. if it doesn't, the existing remote lookup takes over. if the remote is unavailable, it returns `null`. discovery leaves the local pointer unchanged.

### Related issues

Related to #3198. fixes the stale branch lookup. the disabled task buttons weren't reproduced on Linux, so that part remains unverified.

### Testing

used real local repos to rename `main` to `prod` and fetch with pruning. confirmed the regression tests failed before the fix and passed after, including a remote named `team/upstream` and an unavailable remote.

<details>
<summary>Checklist</summary>

- [x] kept this PR small and focused
- [x] reviewed the diff
- [x] added regression coverage for the changed behavior
- [x] only added comments where the logic isn't obvious

</details>


jschwxrz:
lgtm!

jschwxrz (approved):

PR #3238, opened 2026-09-19
fix(opencode): support v2 status hooks

### LINKED ISSUE #3220: [bug]: opencode v2 agent status isn't updating
### Bug Description

im using opencode v2.0.4 in terminal mode and emdash doesn't show running/completed status updates. is v2 supported yet, or does the hook integration need updating?

### Steps to Reproduce

1. Start an opencode v2.0.4 terminal task in emdash.
2. Submit a prompt and let it finish.
3. Watch the agent status.

the status doesn't reflect the agent working or finishing. i'd expect both transitions to show up.

### emdash Version

Workspace server: 0.1.5.
Investigated checkout: 1.2.4 (`5089812e`).
Installed desktop version wasn't accessible from the remote host.

### Operating System

Remote workspace host: Ubuntu 26.04 LTS, x86_64.

### Additional Context

emdash's plugin still uses the v1 `event` callback API. opencode says those plugins don't run in v2:
https://opencode.ai/v2/docs/build/plugins/migrate-v1

there's also a v1 mapping question: `session.idle` becomes `awaiting-input` in emdash, not `completed`. confirmed with simulated events, not a full v1 session.

should this support both plugin APIs, and is the existing completion mapping intentional?

danielgaskins:
### Description

OpenCode v2 no longer calls the v1 `event` callback, so Emdash does not receive agent status updates.

This keeps the v1 plugin entry point and adds a v2 entry point that subscribes to OpenCode events. Execution events now report working, completed, interrupted, and failed states through Emdash's existing hook pipeline. Session IDs are also read from the v2 event shape.

### Related issues

Fixes #3220.

### Testing

- `pnpm --dir packages/plugins build`
- `pnpm --dir packages/plugins typecheck`
- `pnpm --dir packages/plugins exec vitest run --maxWorkers=1 --no-file-parallelism`
- Formatting and lint checks on the changed files

Not applicable. This changes provider hook handling.

<details>
<summary>Checklist</summary>

- [x] I ran the relevant local checks
- [x] No documentation changes are needed
- [x] I added tests for both OpenCode plugin interfaces
- [x] I used Conventional Commits for the commit message and PR title

</details>


danielgaskins:
No prob, interrupted executions now send a plain notification, which returns the agent to idle instead of marking the run completed. I added a regression test for that path. The formatting concern is also resolved because the combined condition has been removed.

PR #3246, opened 2026-09-21
fix(agents): show agent installation documentation links

### LINKED ISSUE #3245: [bug]: Agent installation documentation links never reach the UI
### Bug Description

Agent plugins can provide an `installDocs` URL, but that link never appears in either agent installation surface.

On current `main`:

- `buildMetadata()` copies the plugin's installation URL into `AgentPayload.installDocs`
- `AgentDetailSheet` and `AgentInfoCard` pass `installOptions`, but not `agentPayload.installDocs`
- `InstallSection` accepts an `installDocs` prop as `_installDocs` and never renders it

This currently affects plugins with dedicated installation pages, including CodeBuddy, Grok, MiMo Code, Muse, Oh My Pi, Prime Agent, and Qoder. The general provider website is not always the same page as the installation guide.

### Steps to Reproduce

1. Open an uninstalled agent that provides `installDocs`.
2. View its installation panel from the agent selector or Settings.
3. Look for the plugin's installation guide.

The installation URL is present in the agent payload but unavailable in the UI.

When `installDocs` is set, the installation panel should offer a small link to that guide. Plugins without the field should keep their current layout.

### Proposed Fix

Pass `agentPayload.installDocs` from both callers and render it alongside the installation controls. Add focused coverage for the full settings sheet, the compact selector card, and plugins without a documentation URL.

I'm happy to handle this one. If that plan sounds about right, I'll put up the patch.

danielgaskins:
Closes #3245

Agent payloads already carry a dedicated installation guide URL, but neither installation UI passed it into `InstallSection`. This passes the URL from Settings and the compact agent selector, then shows a small `Installation guide` link when one is available.

Plugins without an installation URL keep the current layout.

Tested with:

- Focused browser tests for the regular layout, compact layout, and missing URL case
- `nx run @emdash/emdash-desktop:typecheck`
- `oxlint` on the changed files


jschwxrz:
thanks for the pr. would appreciate if you added screenshots if there are any changes to the UI in the future

PR #3258, opened 2026-09-23
fix(git): harden log parsing against control characters in commit messages

Nitish-1303:
### Description

`git log` output was parsed using `\x1f` / `\x1e` as field and record delimiters on the assumption that those bytes can't appear in commit metadata, but git only excludes NUL from commit messages. Messages authored through the API, imported histories, or `git commit -F` can carry control characters; a 0x1E in a message splits one commit into a phantom extra record, and a 0x1F shifts every later field (wrong subject/author, date 0, lost tags).

This switches to NUL-delimited fields (`%x00`) with `-z` record termination and parses fixed 7-field records, the same approach as the existing `%00` refs listing format in `refs.ts`.

### Testing

- New regression test commits a message containing 0x1F (subject) and 0x1E (body) in a real repo and reads it back through `getLog` / `getCommit`. It fails before the fix (2 commits parse as 3, with shifted fields) and passes after.
- Full git runtime suite passes locally (121/121); `oxfmt` clean; typecheck clean on the touched files.

### Checklist

- [x] Small, focused change
- [x] Self-reviewed the diff
- [x] Added regression coverage
- [x] Comments only where the logic isn't obvious


Nitish-1303:
Some context for whoever picks this up. The parser split `git log` output on 0x1F/0x1E assuming commit messages never contain them, but git only reserves NUL. A message carrying those bytes (imports, API-authored commits, `git commit -F`) quietly corrupts the parse: phantom commits appear, or fields land on the wrong commit.

Since NUL is the only byte git actually forbids, the fix moves the delimiters onto it and reads fixed-size records. For normal repos nothing changes: same fields in, same `Commit` objects out, and the existing log tests pass untouched. The new test only covers the edge case that used to break. Happy to adjust anything.

Nitish Yeluru / linkedin.com/in/yeluru-nitish

PR #3259, opened 2026-09-24
fix(notifications): open the right chat tab

kdcokenny:
### Description

clicking a notification in codex chat ui mode opened a new blank tab instead of the actual chat. the handler always used the terminal tab type, which is why cli mode worked.

this uses the conversation's type to open the right tab, including when it's already open in another pane. it also waits for conversation data to load and handles unavailable targets without throwing.

### Testing

tested manually. added regression tests using the real tab layout. confirmed the chat ui cases fail before the fix and pass after it. covers cli mode, split panes, closed tabs, delayed loading and unavailable conversations.

<details>
<summary>Checklist</summary>

- [x] kept this focused on notification navigation
- [x] reviewed the diff
- [x] updated the docs
- [x] added regression coverage
- [x] kept comments for logic that needs explaining

</details>


jschwxrz (approved):

jschwxrz:
lgtm, thank you!

